PR #20, opened 2024-02-26
Map system

KKhanhH:
my apologies for whoever is reading this cursed system


MrRevo3D:
@KKhanhH Hey, looks good to me but I think the argument names of some functions like arg1/param_1 would benefit from some more understandable names. Oh and I'd appreciate if you confirm that the renderer changes don't break the other games.

MrRevo3D:
@KKhanhH Thanks! I'll go ahead and merge it

PR #24, opened 2024-04-01
Various more-or-less minor code cleanups

casept:
* Remove some unused includes to speed up compilation and de-clutter
* Add toggle for easily enabling additional compiler warnings
* Fix most warnings reported by clang using Wall and Wextra
* Also fix some additional warnings reported by GCC using Wall and Wextra
* Add some missing error handling pointed out by compiler warnings
* Probably fix some bugs while at it, hopefully without introducing new ones

I haven't tested the warnings toggle for MSVC, as I don't have a Windows system set up for development. It may be worthwhile to enable some additional warnings for it.

Warnings in vendored libraries have not been resolved. In future, it'd probably be best to define them as separate cmake targets and then only enable warnings for the main impacto target.

I've tried to keep formatting-related changes to a minimum. clang-format should be run over the entire codebase at some point in future, and formatting ideally enforced via CI. Most source files have at least some issues with indentation or missing trailing newlines.

Changes haven't been tested for all games, as I don't have compatible versions of most of them.


src/vm/vm.cpp
```diff
@@ -390,16 +387,6 @@ void DestroyThread(Sc3VmThread* thread) {
   thread->NextContext = next;
 }
 
-static void DestroyScriptThreads(uint32_t scriptBufferId) {
```
MrRevo3D, line 393: While this is not used right now it might be used in some game we decide to implement later so I don't think it's very reasonable to outright remove this.

MrRevo3D:
And regarding `std::tm`, I'm seeing two more values in initializer than needed, am I missing something? It doesn't compile on Windows like that @casept

src/renderer/renderer.cpp
```diff
@@ -187,11 +187,14 @@ void BaseRenderer::DrawProcessedText_BasicFont(ProcessedTextGlyph* text,
           DrawSprite(glyph, dest, color);
           dest.X++;
           dest.Y++;
+          break;
```
MrRevo3D, line 198: This fall through is intentional, I guess we can make this part more explicit and just do
```
          dest.X += 2;
          dest.Y += 2;
          DrawSprite(glyph, dest, color);
          break;
```
or something

src/vm/vm.cpp
casept: Thank you, I've addressed this in a follow-up commit.

casept:
> And regarding `std::tm`, I'm seeing two more values in initializer than needed, am I missing something? It doesn't compile on Windows like that @casept

Aha, according to https://en.cppreference.com/w/cpp/chrono/c/tm different platforms may add additional structure fields, so this is probably down to a libc++ difference. I'll refactor the code to construct an empty instance and then explicitly set the fields we care about.

src/renderer/renderer.cpp
casept: Apparently there's no standard way to annotate explicit fallthrough pre-C++17. I've added an __attribute__ that should handle this for GCC and clang, silencing on MSVC will have to wait until we've upgraded the language version.

MrRevo3D: Unfortunately it doesn't seem to work at all on MSVC (the __attribute__), so probably best to just get rid of the fallthrough altogether

casept:
The std::tm issue has been resolved.

src/renderer/renderer.cpp
casept: Done.

MrRevo3D:
> The std::tm issue has been resolved.

I think that should be it then unless you'd like to add something else

casept:
Should be everything for now.

PR #25, opened 2024-04-12
Basically complete cclcc delusion trigger

KKhanhH:
CCLCC Delusion Triggers+Renderer changes so the drawspriteoffset will rotate around the offset.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -433,7 +433,7 @@ set(Impacto_Header
         src/games/chlcc/systemmenu.h
         src/games/chlcc/trophymenu.h
         src/games/chlcc/delusiontrigger.h
-        
+
         src/games/cc/dialoguebox.h
         src/games/cc/titlemenu.h
         src/games/cc/sysmesbox.h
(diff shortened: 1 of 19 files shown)


CMakeLists.txt
```diff
@@ -430,17 +433,18 @@ set(Impacto_Header
         src/games/chlcc/systemmenu.h
         src/games/chlcc/trophymenu.h
         src/games/chlcc/delusiontrigger.h
-
+
```
MrRevo3D: Stray whitespace changes looks like

src/hud/delusiontrigger.h
```diff
@@ -15,8 +15,11 @@ class DelusionTriggerBase {
   DelusionTriggerBase(int& delusionState, UiState showState);
   virtual void Update(float dt) = 0;
   virtual void Render() = 0;
-  virtual void Show() = 0;
   virtual void Hide() = 0;
+  virtual void Show(){};
+  virtual bool Show(int param1, int param2, int param3) { return true; };
```
MrRevo3D: Missed renaming vars here and below

MrRevo3D (commented):
Just these 2 things and we're good to go for now

CMakeLists.txt
KKhanhH: isn't this one fine? it separates chlcc and cc

KKhanhH: oh never mind added a tab

PR #26, opened 2024-04-13
Cosmic ray correction

KKhanhH:


src/io/mpkarchive.cpp
```diff
@@ -80,12 +80,17 @@ IoError MpkArchive::Create(InputStream* stream, VfsArchive** outArchive) {
...
     uint32_t Id = ReadLE<uint32_t>(stream);
+
     if (Compression != 1 && Compression != 0) {
       ImpLog(LL_Error, LC_IO, "Unknown MPK compression type %d on file %d\n",
              Id, Compression);
       stream->Seek(0x100 - 8, RW_SEEK_SET);
       continue;
     }
+    if (result->IdsToFiles.find(Id) != result->IdsToFiles.end()) {
+      ImpLog(LL_Error, LC_IO, "Duplicate MPK file ID %d\n", Id);
```
MrRevo3D: I'd make this a warning and rather than break, just continue. What if there are more files after a duplicate?

src/io/mpkarchive.cpp
```diff
@@ -80,12 +80,17 @@ IoError MpkArchive::Create(InputStream* stream, VfsArchive** outArchive) {
   for (uint32_t i = 0; i < FileCount; i++) {
     uint32_t Compression = ReadLE<uint32_t>(stream);
     uint32_t Id = ReadLE<uint32_t>(stream);
+
```
MrRevo3D, line 83: If we have an empty line here, let's add one before the duplicate if check as well, seems appropriate to me imo

MrRevo3D (changes requested):

src/io/mpkarchive.cpp
KKhanhH: My concern is if the file count doesn't actually match up with how many files there are in a file, like by a significant amount and then stream starts to enter the data chunk or something

MrRevo3D: Fair, though I'd assume if the file count field is 200 for example, there *should* be at least 200 TOC entries, if there aren't - that's a completely broken file

KKhanhH: Maybe do both a duplicate check and just continue, but also do a check for if offset or size is 0 and break, since it looks like there's some zero padding between TOC and data

MrRevo3D: I think the data is aligned, so technically padding might not be there if it happens to match up perfectly. I guess yes, we can also check offset and size but frankly, I don't think you can truly handle all the cases in which it *might* break with a broken archive from their side

KKhanhH: It looks like it adds an extra chunk if it's on the 2048 byte chunk? CCLCC's system.mpk has 40 files but its data section doesn't start until 0x3000 rather than 0x2800. Should be safe to do that then, I know I can't cover every case but just handling a few possible ones should be fine considering one of their files came broken.

PR #32, opened 2024-05-09
Feature/cclcc yesno

KKhanhH:
Should be complete yes/no trigger


profiles/cclcc/sprites.lua
```diff
@@ -78,7 +78,33 @@ root.SpriteSheets = {
         Path = { Mount = "system", Id = 18 },
         DesignWidth = 2048,
         DesignHeight = 3072
-    }
+    },
+    ["YNTRG_BG0"] = {
```
MrRevo3D: I'd prefer if we followed the naming style of all the other sprites to keep consistency.

profiles/cclcc/yesnotrigger.lua
```diff
@@ -0,0 +1,135 @@
...
+      yesChipY = value6,
+      noChipX = value7,
+      noChipY = value8,
+      bubbleX = value9,
+      bubbleY = value10,
+  }
+end
+
+local data1 = {
+  createYesNoData(698.0, -674., 4, 1, 2372.0 , 278.0  , 2626.0 , 440.0  , 2646.0 , 330.0),
```
MrRevo3D: Some weirdness with commas and spaces here in terms of formatting

src/game.cpp
```diff
@@ -133,6 +134,9 @@ static void Init() {
     TipsNotification::Init();
     DelusionTrigger::Init();
     UI::MapSystem::Init();
+    if (Profile::Vm::GameInstructionSet == +Vm::InstructionSet::CC) {
```
MrRevo3D: I guess it's not a problem right now, but regular C;C does not have this, does it?

Enorovan: It doesn't

src/games/cclcc/yesnotrigger.h
```diff
@@ -0,0 +1,51 @@
...
+enum class BGType { BG0, BG1, BG2, BG3 };
+
+void updateYesNoPos(float startX, float startY, float startScale, float targetX,
+                    float targetY, float targetScale, float transition);
+
+template <class Enum>
+constexpr std::underlying_type_t<Enum> to_underlying(Enum e) noexcept {
+  return static_cast<std::underlying_type_t<Enum>>(e);
+};
+};  // namespace YesNoTrigger
```
MrRevo3D, line 43: Probably does not need semicolon here and move the empty line before this line

MrRevo3D:
@KKhanhH Just some formatting stuff and might be a nitpick but not sure I like how we're starting to mix 3 enum styles at this point... I guess all of everything needs to be rewritten anyway, but something about it irks me a bit.

profiles/cclcc/sprites.lua
```diff
@@ -78,7 +78,33 @@ root.SpriteSheets = {
...
+        DesignWidth = 3840,
+        DesignHeight = 2160
+    },
+    ["YNTRG_BG2"] = {
+        Path = { Mount = "system", Id = 37 },
+        DesignWidth = 3840,
+        DesignHeight = 2160
+    },
+    ["YNTRG_BG3"] = {
+        Path = { Mount = "system", Id = 38 },
```
KKhanhH, line 98: Yeah I think I just used the file names for those

profiles/cclcc/sprites.lua
KKhanhH: > I'd prefer if we followed the naming style of all the other sprites to keep consistency.

Yeah I just used the file names for those

src/game.cpp
KKhanhH: Init does already check if LUA file contains the YesNoTrigger table push, so it's fine to remove

KKhanhH:
Formatting fixed, will need to convert to delta timing delays and add in mouse support for yes/no buttons in the future

PR #36, opened 2024-05-21
Add easy-to-use ASAN toggle to CMakeLists

casept:
Tested on GCC and clang (Linux). MSVC has not been tested, as it's too much of a pain in the ass to get the build to work in my Windows VM.

Already found some unaligned reads, might send a PR later fixing those.


MrRevo3D:
Thank you! Though I'm pretty sure that MSVC won't work since I once tried to enable ASAN there and it was not working at all like that.

PR #41, opened 2024-06-01
feature/lua-config-for-realboot

Enorovan:
expose resolution and fullscreen to config, and load which profile to use from lua


src/main.cpp
MrRevo3D, line 1: @Enorovan forgot to remove changes from this file

PR #42, opened 2024-06-06
Feature/cclcc savesystem

KKhanhH:
Full save and load seems to be working. No quicksave yet. System menu very very scuffed, menu might need overhaul for cc.


src/vm/inst_system.cpp
```diff
@@ -213,11 +213,24 @@ VmInstruction(InstSave) {
   switch (type) {  // TODO: Types 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13,
                    // 16, 20, 21, 72, 30, 31, 32, 33, 34, 35, 41, 50, 51, 66,
                    // 67, 70, 71, 74, 76
-    case 32:
```
MrRevo3D, line 216: What about mo6tw? It will not work now

MrRevo3D:
@KKhanhH This will have to wait a bit until I have a chance to test it I think, but so far MO6TW saving will be broken because of this for sure. And sorry, but could you please rebase your branch against latest master?

PR #46, opened 2024-07-01
Achievements

KKhanhH:
Just the display of the achievement menu, currently displays all achievements, no hidden check or achievement count and stuff yet. 
We'll need to set up a system to actually store achievement information somewhere, since there's no need to actually hook it into ps3/ps4 achievements and deal with their encrypted files.


MrRevo3D:
@KKhanhH Hey, could you please fix the pipeline file? Looks like a formatting issue

src/log.cpp
```diff
@@ -77,7 +77,7 @@ const char* ChannelToString(LogChannel channel) {
   }
 }
 
-void ConsoleWrite(const char* str) { printf("%s", str); }
+void ConsoleWrite(const char* str) { fprintf(stderr, "%s", str); }
```
MrRevo3D: Hm, not sure using stderr for general logs makes sense. Is there any particular issue with buffering?

src/games/chlcc/achievementsystem.cpp
```diff
@@ -0,0 +1,145 @@
+#include "achievementsystem.h"
```
MrRevo3D: I assume this is not specific to chlcc is it? Maybe `games/chlcc` is not quite the right place for this, I'm leaning towards moving this into `src/data/achievementsystemps3.cpp` since this looks very much PS3 specific and not game specific.

profiles/chlcc/achievementsystem.lua
```diff
@@ -0,0 +1,4 @@
+root.AchievementData = {
+  Type = AchievementDataType.CHLCC,
```
MrRevo3D: Same here, the type I think should be "PS3" rather than CHLCC, TROPHY.TRP is a PS3 file format rather than a game format

src/log.cpp
KKhanhH: If it crashes on like a lua load for example, logs won't get flushed out and you'll miss the log for the error. We could flush after every print as well.

CMakeLists.txt
```diff
@@ -818,6 +828,7 @@ include_directories(${SDL2_INCLUDE_DIRS})
 include_directories(${WebP_INCLUDE_DIRS})
 include_directories(vendor/include)
 include_directories(vendor/pcg/include)
+include_directories(vendor/pugixml/include)
```
MrRevo3D, line 829: Can we also please add a license for pugixml to LICENSE.md?

src/log.cpp
MrRevo3D: Let's revert that change and send me an example of the issue in Discord, I'll try to take a look at it
I assume this only is true for non-Windows systems since I don't think Windows has stdout buffering

KKhanhH:
> @KKhanhH Hey, could you please fix the pipeline file? Looks like a formatting issue

I'll just drop this commit, it's a holdover from the rebase

PR #51, opened 2024-09-05
Generic CMake Unix Platform Improvement

bedwardly-down:
This simple change allows coverage of Linux, Mac, the BSDs and any other platform that's not Windows. Cygwin counts as "Unix" but can use Direct3D9, so make an exception there. See here: https://cmake.org/cmake/help/latest/variable/UNIX.html


MrRevo3D:
Hello! Thank you for your contribution!

PR #55, opened 2024-09-08
Support for Steins;Gate (Steam)

casept:
Currently mostly just copy-paste from other games. No graphics yet, but the main menu is reached (the theme music plays).


MrRevo3D:
Hey @casept !
Since this is the Steam version and S;G has a bunch of other versions for other consoles (like the PS3 one we already have support for and the original Xbox 360 one which is probably also different in some way) can we please rename all "sg"/"SG" references in this PR to "sghd"/"SGHD"? Just to better differentiate it from the other ones. And I know this is marked as draft but I'll leave some preliminary comments if you don't mind.

profiles/sg/game.lua
```diff
@@ -0,0 +1,51 @@
+root.ActiveRenderer = RendererType.OpenGL;
+
+root.LayerCount = 100;
+root.GameFeatures = GameFeature.Sc3VirtualMachine | GameFeature.Renderer2D | GameFeature.Input | GameFeature.Audio |
+GameFeature.Video | GameFeature.DebugMenu;
+-- NOTE: 1080p messes up rendering of "loading system data..." message,
```
MrRevo3D, line 6: 1080p is correct for S;G HD. But you are reusing the system message box from C;H LCC which is a base 720p game. You can try to reuse the system message box for example from Memories Off 8 (which looks/acts pretty similar) and it should look correct.

src/profile/ui/titlemenu.cpp
```diff
@@ -54,6 +55,10 @@ void Configure() {
       CC::TitleMenu::Configure();
     } else if (Type == +TitleMenuType::CCLCC) {
       CCLCC::TitleMenu::Configure();
+    } else if (Type == +TitleMenuType::SG) {
+      SG::TitleMenu::Configure();
+    } else {
+      ImpLog(LL_Fatal, LC_Profile, "Unknown title menu type in profile!\n");
```
MrRevo3D: I don't think this should be fatal necessarily, maybe just a warning

src/vm/vm.cpp
```diff
@@ -79,24 +80,28 @@ void Init() {
       OpcodeTableSystem = OpcodeTableSystem_Darling;
       OpcodeTableGraph = OpcodeTableGraph_Darling;
       OpcodeTableUser1 = OpcodeTableUser1_Darling;
+      OpcodeTableGraph3D = nullptr;
```
MrRevo3D: Let's move this initialization up to where these are declared (line 50 in this file) and don't explicitly set it for each set that doesn't have 3D.

src/vm/vm.cpp
```diff
@@ -488,12 +507,18 @@ void RunThread(Sc3VmThread* thread) {
       if (opcodeGrp1 == 0x10) {
         OpcodeTableUser1[opcode](thread);
       } else if (opcodeGrp1 == 0x02) {
-        OpcodeTableGraph3D[opcode](thread);
```
MrRevo3D: You've removed 3D instructions thus breaking all R;Ns.
Let's not do it like this, let's do something like
```
else if (opcodeGrp1 == 0x02 && OpcodeTableGraph3D != nullptr) {
  OpcodeTableGraph3D[opcode](thread);
}
.....
```
Games that don't have 3D will never hit this if every instruction is defined properly on our side. I don't think a special message is needed for this as the reason is the same as it is in "Unknown opcode" case. This would also get rid of "goto" which I don't think is necessary here.

MrRevo3D (changes requested):
Left some comments for the things I've noticed just so I don't forget about it later.

src/vm/vm.cpp
casept: Whoops, must've been left over from when I was debugging a related issue. Fixed.

src/vm/vm.cpp
casept: Done.

src/profile/ui/titlemenu.cpp
casept: OK, downgraded to warning.

MrRevo3D:
Thank you for the changes! 👍

PR #58, opened 2024-09-14
Cclcc savemenu thumbnail

KKhanhH:
thumbnails are a 16 bit sensation


src/game.cpp
```diff
@@ -379,6 +379,10 @@ void Render() {
               CCLCC::YesNoTrigger::YesNoTriggerPtr->Render();
             }
           }
+          
+          if(SaveSystem::Implementation) {
```
MrRevo3D: nitpick, missing space between if and (

src/renderer/dx9/renderer.h
```diff
@@ -32,6 +32,16 @@ class Renderer : public BaseRenderer {
 
   uint32_t SubmitTexture(TexFmt format, uint8_t* buffer, int width,
                          int height) override;
+  std::vector<uint8_t> GetImageFromTexture( uint32_t texture,
```
MrRevo3D, line 35: Some formatting weirdness here and below. Please reformat with clang-format we usually use

PR #60, opened 2024-09-17
Linting

KKhanhH:


KKhanhH:
@MrRevo3D automatic linting should work now. I added an pipeline triggers for PR (from forks too), although that pipeline needs to be merged to master to be active

PR #62, opened 2024-09-18
Bugfix/cclcc saveload

KKhanhH:
Should fix the loaded from saves being broken. Not sure why there's so many threads created when you repeatedly load though, maybe MAINTHDP ScrWork needs to get reset or something. Potential source of crash if you repeatedly load


KKhanhH:
/todo killthread and reset thread not implemented so thats probably why

PR #65, opened 2024-09-21
CC(LCC) Backlog

PringlesGang:
This PR implements the backlog menu for both C;C and C;CLCC.

It further converts the existing backlog UI implementation to use inheritance, to be more extendable for implementing additional games.

Also implements `InstSetRevMes` for adding a message directly to the backlog.

This should not break the existing MO6TW backlog menu, though I recommend you test this yourself as I can't myself.


profiles/common/scriptvars.lua
```diff
@@ -52,6 +52,7 @@ root.ScriptVars = {
     SW_MASK3ALPHA_OFS=2788,
     SW_MAINTHDP=3304,
     SW_RESTARTMASK=3318,
+    SW_SETREVMES_VALUE=3319,
```
KKhanhH: worklist has this var as SW_PLATFORM, would it make sense to rename it to match? @MrRevo3D

MrRevo3D: Hm, well, I do trust the worklist, but the way it's used here doesn't seem like it would be SW_PLATFORM. @PringlesGang could you please explain what SETREVMES_VALUE is actually used for? I don't think I'm seeing it

src/games/cc/backlogmenu.cpp
```diff
@@ -0,0 +1,68 @@
...
+using namespace Impacto::Profile::BacklogMenu;
+using namespace Impacto::Profile::CC::BacklogMenu;
+
+BacklogMenu::BacklogMenu() : UI::BacklogMenu::BacklogMenu() {
+  switch (Impacto::Profile::SystemMenu::Type) {
+    default:
+      MaskSprite = nullptr;
+      break;
+    case UI::SystemMenuType::CCLCC:
+      MaskSprite = &Profile::CCLCC::SystemMenu::MenuMask;
```
KKhanhH: yeah since we have a separate backlogmenu.lua, it probably would be a good idea to have a separate profile and class for it too, so there's no need to switch on the types

src/text.cpp
```diff
@@ -160,6 +160,23 @@ uint8_t ProcessedTextGlyph::Flags() const {
   return Profile::Charset::Flags[CharId];
 }
 
+void ProcessedTextGlyph::Move(std::vector<ProcessedTextGlyph>& text,
```
KKhanhH: I did add a move function to dialoguepage that moves both the text glyphs and the name, would that not work?

profiles/common/scriptvars.lua
PringlesGang: Yeah I kinda didn't know what to do with this at first.
It gets set at initialization of the game, but doesn't seem to change from there. It's used the script in `macrosys2.scx` in all games, where its value determines whether it adds a long dash separator in the backlog (for example at the start of each chapter), or other stuff, like the date and time in CC, using `InstSetRevMes`.
Its starting value differs between different games (`0x4000` in CCLCC PSV, `0x20000` in CC Steam and CCLCC PS4, `0x80000` in CC Android), so I figured I'd best make it a profile configuration
Frankly I think the name is terrible too, so I'm more than open to suggestions

src/games/cc/backlogmenu.cpp
PringlesGang: You mean add a separate `MenuMask` sprite copy in each component that uses it, and pass it along in the constructor?

KKhanhH: Actually I was thinking it might just have been better to keep the mask sprite inside the backlog profile namespace, rather than getting it from system menu, since lua load handles the switching anyway. Backlog class can just use sprite directly from backlog profile. I don't think having a couple of duplicate sprites is that big of a deal?

profiles/common/scriptvars.lua
KKhanhH: Yeah, it does sound like it's just to indicate the platform the game is running on, and not just specific to the backlog stuff, so maybe have it as SW_PLATFORM. Using profile configuration sounds good instead since SW_PLATFORM isn't actually platform for us.

src/ui/backlogmenu.cpp
```diff
@@ -43,109 +50,228 @@ BacklogMenu::BacklogMenu() {
...
       LastFocusedMenu->IsFocused = true;
     } else {
-      UI::FocusedMenu = 0;
+      UI::FocusedMenu = nullptr;
     }
-    IsFocused = false;
+  }
+}
+
+inline bool inVerticalHoverBounds(const Widget* entry) {
```
KKhanhH: Remove the inline and capitalize function name

KKhanhH (changes requested):

src/text.cpp
PringlesGang: Pretty sure I added this before you added that, but yeah it's changed to use yours now

src/ui/backlogmenu.cpp
PringlesGang: Done

src/games/cc/backlogmenu.cpp
PringlesGang: You're right; should be good now

profiles/common/scriptvars.lua
PringlesGang: Renamed it and moved it to `game.lua`

PR #68, opened 2024-09-27
Load spritesheets concurrently

newo-2001:
This PR loads Spritesheets in parallel on different threads using `std::async`.

I was annoyed by the slow startup times when testing things, especially in debug mode.
After some profiling I found that ~40% of the time was spent in stb_image.
By splitting this work over multiple threads the time-to-menu is significantly faster with a relatively simple change.
Loading sprite sheets for CC;LCC in Debug mode went from ~22s to ~13s for me (Ryzen 7 5800X)

There are no doubt more optimal ways to squeeze out more performance but my main goal was to speed up iteration time during development by picking up some low hanging fruit, and this seems to be a good start.

CPU utilization before:
image

CPU utilization after:
image

### DIFF
--- a/src/profile/sprites.cpp
+++ b/src/profile/sprites.cpp
@@ -46,7 +46,8 @@ void LoadSpritesheets() {
       Window->Shutdown();
     }
 
-    futures.emplace_back(std::tuple(name, std::async(std::launch::async, LoadTexture, stream, name)));
+    futures.emplace_back(std::tuple(
+        name, std::async(std::launch::async, LoadTexture, stream, name)));
 
     Pop();
   }
--- a/src/profile/sprites.cpp
+++ b/src/profile/sprites.cpp
@@ -4,12 +4,29 @@
 #include "../log.h"
 #include "../renderer/renderer.h"
 #include "../texture/texture.h"
+#include <future>
 
 namespace Impacto {
 namespace Profile {
 
+static Texture LoadTexture(Io::Stream* stream, std::string name) {
+  Texture texture{};
+  if (!texture.Load(stream)) {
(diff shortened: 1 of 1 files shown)


KKhanhH:
Hey, thanks for the PR.

I think we'd have to test if this works with the switch version or not. It seems to be crashing when I try to emulate the release from this branch at least. if std::async doesn't work then it might be necessary to use SDL_Threads kinda like how WorkerQueue does it.

newo-2001:
Are there any resources on how to setup the dev environment for switch so I can hack away at it?

KKhanhH:
I don't think we have any, you'll probably need devkitpro, try taking a look at the pipeline for the switch build to see how that works. As for running it on the switch itself, you'll just need to put the runtime files (profiles, shaders, profile.txt, games) on the root of the SD card.

KKhanhH:
I guess you could also just use the pipeline as a build environment

newo-2001:
Thanks, I'll have look tomorrow!

KKhanhH:
It might also not be limited to just switch, I need to setup a Linux environment to test things as well, but the crash could just also be an implementation thing between gcc and msvc, so if Linux also crashes, then fixing that crash will probably fix it on switch as well

newo-2001:
I'm gonna go off the assumption that getting it working through SDL threads will solve the problem. Will probably try to implement some generic threadpool.

KKhanhH:
Sorry, ignore everything i just said... I just didn't copy the latest profile. Switch build runs fine

KKhanhH:
Once i merge fstream changes as well, you should see a significant improvement in loading times

PR #70, opened 2024-10-02
C;CLCC wait icon

PringlesGang:
Implements the C;CLCC wait icon (as well as improving the logic for the wait icon in general)

- Implements wait icon logic
- Moves rendering functions out to static methods of the specific wait icon rendering types for readability and extensibility
- Properly implements `SpriteAnim` and `SpriteAnimFixed` logic for wait icons

The wait icon rendering function _technically_ also checks for `SF_SHOWWAITICON + 1` and `+ 2`, but only if the window priority `ScrWork[4359]` is changed in `MESsetWinPri`, which doesn't seem to happen yet in Impacto, so I left it to simply `SF_SHOWWAITICON` for now.


src/hud/waiticondisplay.cpp
```diff
@@ -15,87 +18,147 @@ using namespace Impacto::Profile::Dialogue;
...
-
-      } else {
-        Renderer->DrawSprite(
-            SpriteAnim.CurrentSprite(),
-            glm::vec2(pos.x + WaitIconOffset.x, pos.y + WaitIconOffset.y), col);
-      }
+
+static void RenderSpriteAnim(glm::vec2 pos, glm::vec4 opacityTint,
+                             DialoguePageMode mode) {
+  if (GetFlag(Profile::ScriptVars::SF_SHOWWAITICON))
```
KKhanhH: I think animation changes make more sense in an update loop than in the render functions

src/hud/waiticondisplay.cpp
```diff
@@ -15,87 +18,147 @@ using namespace Impacto::Profile::Dialogue;
...
+    SpriteAnim.StartIn();
+  else
+    SpriteAnim.StartOut();
+
+  if (FixedSpriteAnim.IsOut()) return;
+
+  if (DialogueBoxCurrentType == +DialogueBoxType::CHLCC) {
```
KKhanhH, line 82: You could probably cut this down to one drawsprite call and have a single level of if/else checks that sets tint & offset depending on branch, should help with readability

src/hud/waiticondisplay.cpp
KKhanhH: Also, do the startin and startout functions need to be called continuously? Can they not be checked for completion first, or is that not how fixedspriteanimation works?

src/spriteanimation.h
KKhanhH, line 1: Class FixedSpriteAnimationDef; doesn't actually exist does it? It's not in this PR but I noticed it in this file.

KKhanhH (changes requested):

src/hud/waiticondisplay.cpp
PringlesGang: Yup, it's much more readable now

src/spriteanimation.h
PringlesGang: Yeah, that's still a remnant from when I first implemented `FixedSpriteAnimation` for the Skip/Auto icons; forgot to remove it. It's gone now

src/hud/waiticondisplay.cpp
PringlesGang: Moved 'em to `Update`, and also had them only fire `StartIn/Out` when `IsOut/In`.
Don't really see what difference that last one makes, since now it's continuously calling `IsIn/Out`, but I suppose semantically it makes a little more sense

KKhanhH: Mostly semantics yeah, and I think IsIn/IsOut is basically guaranteed to be inlined with how simple it is, versus StartIn which then calls on a virtual function defined in a cpp (not sure if we have link time optimizations enabled yet)

KKhanhH (approved):

PR #72, opened 2024-10-12
Slow down C;C(LCC) backlog fade when opened from menu

PringlesGang:
The C;C(LCC) backlog fade animation is twice as slow when opened from the menu as opposed to opening it directly. This PR implements such functionality.

### DIFF
--- a/src/vm/inst_dialogue.cpp
+++ b/src/vm/inst_dialogue.cpp
@@ -226,6 +226,7 @@ VmInstruction(InstMesRev) {
   PopUint8(type);
   switch (type) {
     case 0:  // DispInit
+      UI::BacklogMenuPtr->Show();
       ImpLogSlow(LL_Warning, LC_VMStub,
                  "STUB instruction SetMesModeFormat(type: DispInit)\n");
       break;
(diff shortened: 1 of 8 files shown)


src/vm/inst_dialogue.cpp
```diff
@@ -226,7 +226,6 @@ VmInstruction(InstMesRev) {
   PopUint8(type);
   switch (type) {
     case 0:  // DispInit
-      UI::BacklogMenuPtr->Show();
```
KKhanhH: Was this already here for MO6 btw?

KKhanhH: if it was, maybe just add a //TODO or something to change it to update loop based show

PringlesGang: Yeah, it was already there for MO6, but we also already have update loop-based showing, and that should open the backlog with every `InstMesRev_DispInit` call already (at least in the C;C scripts). I could still ask @MrRevo3D to verify it still works for MO6 though

PringlesGang: Yeah it, along with *many* other things (unrelated to this PR) have broken. I've reverted this deletion and changed the logic to still work for C;C(LCC) with it, and will get to undoing the rest of the MO6 regression in another PR

KKhanhH (approved):

PR #73, opened 2024-10-13
Very early C;CLCC LibraryMenu stuff and some bugfixing to merge before the progress report

Enorovan:


profiles/cclcc/hud/extramenus.lua
```diff
@@ -2,19 +2,39 @@ root.ExtraMenus = {
...
 
+-- Common
+root.Sprites["LibraryMask"] = {
+    Sheet = "MenuChip",
+    Bounds = { X = 154, Y = 140, Width = 1900, Height = 1061 },
+};
+
 -- ClearList
-root.Sprites["ClearListBackground"] = {
+root.Sprites["MenuBackground"] = {
```
KKhanhH: I don't think clearlist should actually have background sprite from the titlemenu, titlemenu is supposed to be rendered while clearlist is

src/vm/inst_misc.cpp
```diff
@@ -96,7 +96,7 @@ VmInstruction(InstSystemMenu) {
       break;
     case 1: {
       if (!UI::SystemMenuPtr->ChoiceMade) {
-        if (!(Interface::PADinputButtonWentDown & Interface::PAD1B) &&
```
KKhanhH: revert this

KKhanhH (changes requested):

KKhanhH (approved):

PR #78, opened 2024-10-14
CHLCC intro sequence

newo-2001:
This is a draft PR for visiblity.

This PR adds some of the animations for the quite intricate CHLCC intro sequence which involves bouncing, exploding and falling stars.
To satisfy these needs, I added a new `PathAnimation` class which represents something moving along a path in space.

One thing of note is that the sound that plays during the intro sequence is started from the script with a `type=1` parameter. This was previously unimplemented but it seems to pause the playback of the audio immediately. This will then later be resumed by the drawing code in order to synchronize the sound with the star bouncing animation.


src/vm/inst_sound.cpp
```diff
@@ -44,18 +44,24 @@ VmInstruction(InstSEplay) {
...
   PopUint8(type);
+
   if (type != 2) {
     PopExpression(effect);
     PopExpression(loop);
     ScrWork[SW_SEREQNO + channel] = effect;
     Audio::Channels[Audio::AC_SE0 + channel]->Volume =
         (ScrWork[SW_SEVOL + channel] / 100.0f) * 0.3f;
+
+    // Hack to prevent playing the first sound byte before pausing
```
newo-2001: When pausing playback immediately after starting it, like this code does. A few milliseconds of sound is still played before it can be paused. This workaround is not a good solution, but I think we might have to extend the `AudioChannel` interface in order to support this use case.

newo-2001: #161 resolves the underlying issue, when it is merged the hack can be removed

src/renderer/opengl/renderer.cpp
```diff
@@ -866,6 +879,46 @@ void Renderer::DrawCHLCCMenuBackground(const Sprite& sprite, const Sprite& mask,
...
+
+  if (CurrentMode != R2D_ColorShift) {
+    CurrentMode = R2D_ColorShift;
+    Flush();
+  }
+
+  glBindVertexArray(VAOSprites);
+  glUseProgram(ShaderProgramColorShift);
+  glActiveTexture(GL_TEXTURE0);
+  glBindTexture(GL_TEXTURE_2D, sprite.Sheet.Texture);
```
PringlesGang: These two calls could just be `glBindTextureUnit(0, sprite.Sheet.Texture)` (though this whole function will need to be rewritten anyway once #165 is merged in)

src/renderer/renderer.h
```diff
@@ -17,6 +18,8 @@ enum class RendererOutlineMode { None, BottomRight, Full };
 
 constexpr inline int MaxFramebuffers = 10;
 
+BETTER_ENUM(RendererBlendMode, int, Normal, Additive)
```
PringlesGang: This could just be an enum class. I don't think you'll have to do anything fancy with it

src/shaders/opengl/SpriteColorShift_frag.glsl
```diff
@@ -0,0 +1,14 @@
...
+out vec4 color;
+
+uniform sampler2D ColorMap;
+uniform vec4 ColorShift;
+
+void main() {
+    color = texture(ColorMap, uv) * tint;
+    float alpha = color.w;
+    color = clamp(color + ColorShift, 0.0, 1.0);
+    color.w = alpha;
```
PringlesGang: `.a` is a lot clearer here than `.w`

src/games/chlcc/introsequence.cpp
```diff
@@ -0,0 +1,383 @@
...
+
+void IntroSequence::DrawDelusionADVText() const {
+  if (SeiraAnimation.IsOut()) {
+    float animationStep;
+    float progress =
+        std::modf(DelusionADVAnimation.Progress * IntroDelusionADVSpriteCount,
+                  &animationStep);
+
+    for (size_t index = 0; index < IntroDelusionADVSpriteCount; index++) {
+      if (index > animationStep) continue;
```
PringlesGang: Can't you just loop until `std::min(IntroDelusionADVSpriteCount, animationStep + 1)`?

PringlesGang (changes requested):

src/games/chlcc/introsequence.cpp
newo-2001: I don't see why not honestly..

src/renderer/opengl/renderer.cpp
newo-2001: Yeah, I just copied and adjusted an existing function. I'd suggest redoing it when merging with the other PR.

src/games/chlcc/introsequence.cpp
```diff
@@ -0,0 +1,383 @@
...
+    auto& [origin, angle] = FallingStarSeeds[i];
+
+    int random = CALCrnd(100);
+    origin.x = -120 + i * 110 + (random + 10) * 10;
+    origin.y = (random + 10) * -10;
+
+    angle = CALCrnd(8192) / 8192.0f * M_PI * 2;
+  }
+
+  PanningAnimation.SetDuration(IntroPanningAnimationDuration);
```
KKhanhH, line 34: If the animations are ordered, can we put them into an array, durations too, and then loop over those to do addanimation or setduration? It should make it easier to read or modify maybe.

KKhanhH: For naming animation elements in the array you could use an enum as the index.

src/games/chlcc/introsequence.cpp
```diff
@@ -0,0 +1,383 @@
...
+  Renderer->DrawSprite(IntroBackgroundSprite, glm::vec2(0.0f),
+                       {1.0f, 1.0f, 1.0f, PanningAnimation.Progress});
+
+  glm::vec2 zoomFactor =
+      designDimensions / 16.0f + designDimensions / 16.0f * 7.0f * progress;
+
+  Renderer->SetBlendMode(RendererBlendMode::Additive);
+
+  for (size_t i = 0; i < IntroHighlightCount; i++) {
+    constexpr float SCALE = 1.5f;
```
KKhanhH: constexpr static and lowercase name is fine

src/sequencedanimation.h
```diff
@@ -0,0 +1,34 @@
...
+
+struct ChildAnimation {
+  Animation* ChildAnimation;
+  float StartTime;
+  float EndTime;
+};
+
+class SequencedAnimation : public Animation {
+ public:
+  void AddAnimation(Animation* animation, float startTime, float duration);
```
KKhanhH: If Animation* should never be nullable, then the API should only accept Animation&, and assign the address of that to ChildAnimation or have ChildAnimation store a reference_wrapper.

src/renderer/vulkan/renderer.h
```diff
@@ -106,6 +106,10 @@ class Renderer : public BaseRenderer {
   void DrawCHLCCMenuBackground(const Sprite& sprite, const Sprite& mask,
                                const RectF& dest, float alpha) override;
 
+  void DrawSpriteColorShift(const Sprite& sprite, const RectF& dest,
+                            glm::vec4 tint, float angle,
+                            glm::vec4 colorShift) override {}
```
KKhanhH: Add a todo for these empty functions

src/games/chlcc/introsequence.cpp
KKhanhH: Just in case, store the result of the std::min first, and use that variable as the loop condition, just in case it calls std::min repeatedly or something.

src/games/chlcc/titlemenu.cpp
```diff
@@ -222,6 +224,15 @@ TitleMenu::TitleMenu() {
...
   CurrentExtraItems = LockedExtraItems;
+
+  SpinningCircleAnimation.LoopMode = AnimationLoopMode::Loop;
+  SpinningCircleAnimation.SetDuration(SpinningCircleAnimationDuration);
+  SpinningCircleFlashingAnimation.LoopMode =
+      AnimationLoopMode::ReverseDirection;
+  SpinningCircleFlashingAnimation.SetDuration(
+      SpinningCircleFlashingAnimationDuration);
+
+  IntroSequence = std::make_unique<Impacto::CHLCC::IntroSequence>();
```
KKhanhH: Can this not just be a value?

KKhanhH (changes requested):

src/games/chlcc/introsequence.cpp
newo-2001: They are not strictly ordered, some animations start before others are finished so there might be any number of these active at any given time. Some of the animations are simply appended and others are inserted before the current end point. Personally I think it would end up being a bigger mess.

src/games/chlcc/titlemenu.cpp
newo-2001: This proves to be rather troublesome, I'll walk you through what I tried, maybe you have a better idea.

The problem arises because of the code that restarts the animation after the video finishes playing:
```
if (!IntroSequence->IntroAnimation.IsOut()) {
    IntroSequence = std::make_unique<Impacto::CHLCC::IntroSequence>();
}
```

I wanted to avoid having methods like `Reset()` or `Restart()` that manage state on the IntroSequence class because it is more error-prone and opted for just creating a fresh object.

I tried changing it to `IntroSequence = Impacto::CHLCC:IntroSequence();`
But this is problematic because the `SequencedAnimation` stores pointers to the members of the temporary object, those are then copied by value by the default assignment operator resulting in de-referencing garbage.

Then I tried writing a custom move assignment operator but that would also require recreating the `SequencedAnimation` object in the move assignment operator.

I frankly don't know what the idiomatic way of dealing with this is whilst avoiding some form of 2-phase initialization or heap allocation.

newo-2001:
> Not sure where we also want to put this PR, before or after the renderer refactor

I don't think it really matters to be honest, I'm probably gonna be involved with resolving the conflicts regardless of which goes first.

src/games/chlcc/titlemenu.cpp
KKhanhH: Std::optional would probably work in this case then

KKhanhH: That being said, I don't really like that you have to reconstruct the whole object to restart the animation, it's kind of inconsistent with the other animation classes we have, and it reinitializes and reloads textures and durations that stay the same.

newo-2001: After taking a look at what such a `Reset()` method would like it dawned on me that resetting all the animation progresses can be handled by the SequencedAnimation class simply by calling `IntroAnimation.StartIn(true)`

I haven't found a way to avoid re-creating the mask texture because I believe the renderer does not expose a way to write into the texture directly, and making a screencap after clearing the framebuffer also seems like a bad idea.

KKhanhH: Yeah, this looks a lot better, thanks

PringlesGang (approved):

KKhanhH (approved):

PR #81, opened 2024-10-16
C;CLCC config menu UI

PringlesGang:
Will implement the UI for the C;CLCC config menu
(Will do the actual implementation in a separate PR once this is done)


src/games/cclcc/optionsmenu.cpp
```diff
@@ -18,61 +16,398 @@ namespace CCLCC {
...
+  BasicPage = new Group(this);
+  for (int i = 0; i < 4; i++) {
+    BasicPage->Add(new OptionsBinaryButton(BinaryBoxSprite, OnSprite, OffSprite,
+                                           LabelSprites[i], pos, highlightTint,
+                                           select, highlight),
+                   FDIR_DOWN);
+
+    pos.y += EntriesVerticalOffset;
+  }
+  Pages.push_back(BasicPage);
```
KKhanhH: Do you delete pages elements? probably should make Pages a vector of unique_ptrs so it deletes itself

KKhanhH: or just store the pages by value in an array tbh. it's fixed anyway

src/games/cclcc/optionsmenu.cpp
```diff
@@ -18,61 +16,398 @@ namespace CCLCC {
...
+        VoiceSliderTrackSprite, NametagSprites[i], PortraitSprites[2 * i],
+        PortraitSprites[2 * i + 1], pos, highlightTint, SliderSpeed, select,
+        highlight);
+    VoicePage->Add(widget, FDIR_RIGHT);
+  }
+
+  // Loop separately to overwrite the direction set at initial adding
+  // First entry won't set anything; skip
+  for (int i = 1; i < entries; i++) {
+    Widget* const widget = VoicePage->Children.at(i);
```
KKhanhH: Minor nitpick but you probably don't need .at if you're already checking bounds with for loop anyway + debug mode has the bounds assertions

src/games/cclcc/optionsmenu.cpp
```diff
@@ -18,61 +16,398 @@ namespace CCLCC {
...
       ScrWork[SW_SYSSUBMENUNO] == 5) {
     Hide();
   } else if (ScrWork[SW_SYSSUBMENUCT] >= 32 && State == Hidden &&
              ScrWork[SW_SYSSUBMENUNO] == 5) {
     Show();
   }
+
+  if (State != Hidden) {
+    UpdateInput(dt);
+    Pages.at(CurrentPage)->Update(dt);
```
KKhanhH: Probably also shouldn't be using at in update loops

src/games/cclcc/optionsmenu.h
```diff
@@ -10,13 +16,43 @@ class OptionsMenu : public Menu {
...
+  void UpdateInput(float dt);
+  void Render() override;
 
  private:
+  void GoToPage(int pageNumber);
+  void PageButtonOnHover(int pageNumber);
+
+  void Select(OptionsEntry* entry);
+  void Highlight(Widget* entry);
+  inline bool AnyEntrySelected() {
```
KKhanhH: Inline is implicit for class definitions so you don't need it

src/ui/widget.h
```diff
@@ -24,7 +24,7 @@ class Widget {
 
   virtual void Update(float dt);
   virtual void UpdateInput() = 0;
-  virtual void Render() = 0;
```
KKhanhH: I'm not sure how often you would have renderless widgets so I think it would be better to leave the base as abstract so then you get compile errors if you don't define it. For clickbutton you could just add an empty override.

src/ui/widgets/cclcc/optionsentry.cpp
```diff
@@ -0,0 +1,106 @@
...
+void OptionsEntry::MoveTo(glm::vec2 pos) {
+  const glm::vec2 relativePosition = pos - Bounds.GetPos();
+  Widget::MoveTo(pos);
+  EntryButton.Move(relativePosition);
+}
+
+void OptionsEntry::EntryButtonOnClick(ClickButton* target) {
+  if (Selected) return;
+
+  Audio::Channels[Audio::AC_REV]->Play("sysse", 2, false, 0.0f);
```
KKhanhH: Wrong audio channel

src/util.h
```diff
@@ -72,6 +72,30 @@ struct RectF {
             rect.Y + rect.Height <= Y + Height);
   }
 
+  constexpr RectF operator+(RectF const& other) const {
```
KKhanhH: I dont know if I like overloading operator + and operator- for this. Geometrically adding every member of a rectangle doesn't really mean anything and the new rect doesn't really relate much to the originals.

src/vm/inst_misc.cpp
```diff
@@ -219,14 +219,6 @@ VmInstruction(InstOption) {
                  "STUB instruction Option(type: Init)\n");
       break;
     case 1:
-      if (!((Interface::PADinputButtonWentDown & Interface::PAD1B) ||
```
KKhanhH, line 215: Is this not used in any of the other games as well?

KKhanhH (changes requested):

src/ui/widgets/cclcc/optionsvoiceslider.cpp
```diff
@@ -0,0 +1,110 @@
...
+          RectF(pos.x + VoiceSliderOffset.x, pos.y + VoiceSliderOffset.y,
+                box.ScaledWidth(), box.ScaledHeight()),
+          sliderSpeed, select, highlight),
+      Portrait(portrait),
+      MutedPortrait(mutedPortrait) {
+  Bounds =
+      RectF(Bounds.X, Bounds.Y, VoiceEntryDimensions.x, VoiceEntryDimensions.y);
+  EntryButton.Bounds = Bounds;
+
+  std::function<void(ClickButton*)> onClick = std::bind(
```
KKhanhH: apparently std::bind is slower than the lambda equivalent so we might want to consider changing these and all the other bind assignments

src/games/cclcc/optionsmenu.cpp
PringlesGang: They store unique pointers now

src/games/cclcc/optionsmenu.cpp
PringlesGang: Yeah, I guess debug mode should indeed pick up on that. Changed them all to direct indexing

src/games/cclcc/optionsmenu.cpp
PringlesGang: Changed this too

src/games/cclcc/optionsmenu.h
PringlesGang: Oh really? Cool

src/ui/widget.h
PringlesGang: Yeah, it's probably better to leave it an opt-out thing, you're right

src/ui/widgets/cclcc/optionsentry.cpp
PringlesGang: Haha whoops

src/ui/widgets/cclcc/optionsvoiceslider.cpp
PringlesGang: Is this a thing I should change here right now, or do we refactor this in bulk later?

src/vm/inst_misc.cpp
PringlesGang: From what I could tell from the decomp, case 1 should call the main update loop of the options menu, which we do somewhere else already regardless of the VM.

All this switch case used to do is set `SF_SUBMENUEXIT` based on user input, which every implemented case of the options menu already does in _its_ update loop. It should not be used by any game for which the options menu is implemented.

The only thing I see that _might_ go wrong here, is the instruction passing despite `SF_SUBMENUEXIT` not being triggered yet (which I have fixed now).

src/util.h
PringlesGang: I mostly used this overload to add a movement vector to the `RectF`; would it be better I overload `constexpr RectF operator+(const glm::vec2 movement_vector) const` instead, add a `void RectF::Translate(float x, float y)` method, or simply deal with painfully constructing new `RectF`-s like we used to?

src/ui/widgets/cclcc/optionsvoiceslider.cpp
KKhanhH: we can just refactor in bulk, i guess i'll do it in my cpp modernizations branch

src/util.h
KKhanhH: I think translate using operator+ makes sense with a vec 2

src/games/cclcc/optionsmenu.cpp
```diff
@@ -18,60 +16,261 @@ namespace CCLCC {
...
+
+    pos.y += SoundEntriesVerticalOffset;
+  }
+  Pages.push_back(std::move(soundPage));
+
+  std::unique_ptr<Group> voicePage = std::make_unique<Group>(this);
+  constexpr int columns = 3;
+  constexpr int entries = 12;
+  for (int i = 0; i < entries; i++) {
+    glm::vec2 pos = VoicePosition;
```
KKhanhH: You're shadowing pos here

src/games/cclcc/optionsmenu.cpp
```diff
@@ -18,60 +16,261 @@ namespace CCLCC {
...
+
+  UpdateEntryMovementInput(dt);
 }
 
 void OptionsMenu::Render() {
   if (State != Hidden && ScrWork[SW_SYSSUBMENUCT] >= 32 &&
       ScrWork[SW_SYSSUBMENUNO] == 5) {
-    // glm::vec4 col(1.0f, 1.0f, 1.0f, FadeAnimation.Progress);
-    Renderer->DrawSprite(BackgroundSprite, glm::vec2(0.0f));
+    glm::vec4 col(1.0f, 1.0f, 1.0f, FadeAnimation.Progress);
```
KKhanhH: Since everything is already being made const, it might not be a bad idea to make col const and maskTint const with const glm::vec4 maskTint = col * glm::vec4{glm::vec3{1}, (float)0xa0 / 0x100}};

src/games/cclcc/optionsmenu.cpp
PringlesGang: Oh yeah that can just reuse the old pos. Thanks for the heads-up

src/games/cclcc/optionsmenu.cpp
PringlesGang: Good one

KKhanhH (approved):

PR #83, opened 2024-10-26
Animation refactor

newo-2001:
Performed some cleanup in the animation class:
- Changed the existing enums to better enums
- Changed direction into an enum
- Added some helper methods: `Stop()` and `SetDuration()`

Previously the behavior of animations was not well defined when updated with high `dt` values (i.e. the game losing focus for an extended period of time or tripping a breakpoint)
Now the animation progress is properly updated to account for possibly missed loops.

### DIFF
--- a/src/games/cclcc/mapsystem.cpp
+++ b/src/games/cclcc/mapsystem.cpp
@@ -208,7 +208,8 @@ int MapSystemCCLCC::MapLoad(uint8_t* data) {
         : (MapPartsDisp[i].state == Showing || MapPartsDisp[i].state == Shown)
             ? 1 - progress / conversionFactor
             : 0;
-    MapPartsDisp[i].fadeAnim.Direction = inOrOut;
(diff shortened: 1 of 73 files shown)


src/games/cclcc/mapsystem.cpp
```diff
@@ -208,7 +208,8 @@ int MapSystemCCLCC::MapLoad(uint8_t* data) {
         : (MapPartsDisp[i].state == Showing || MapPartsDisp[i].state == Shown)
             ? 1 - progress / conversionFactor
             : 0;
-    MapPartsDisp[i].fadeAnim.Direction = inOrOut;
+    MapPartsDisp[i].fadeAnim.Direction =
+        inOrOut ? AnimationDirection::In : AnimationDirection::Out;
```
newo-2001, line 212: This was previously setting direction to 1 or 0, I'm assuming this was meant to be 1 or -1

KKhanhH (approved):

PR #104, opened 2025-01-19
C;C LCC config menu implementation

PringlesGang:
Connects the backend options variables to the config menu UI for C;C LCC, implementing these options in the backend as necessary.


src/vm/inst_misc.cpp
```diff
@@ -212,10 +212,6 @@ VmInstruction(InstOption) {
                  "STUB instruction Option(type: Init)\n");
       break;
     case 1:
-      if (!GetFlag(SF_SUBMENUEXIT)) {
-        ResetInstruction;
-        BlockThread;
-      }
```
PringlesGang: This instruction is not supposed to block; in the scripts it stays in a loop where it also checks for the "reset to default" keypress, among other things.

This change does <ins>not</ins> break options menus for other games.

src/inputsystem.h
```diff
@@ -46,5 +46,14 @@ inline bool KeyboardButtonIsDown[SDL_NUM_SCANCODES] = {false};
...
 
+// Controller settings
+
+// Advance text on L/R stick, arrow keys, etc.
+inline bool AdvanceTextOnDirectionalInput = false;
+
+// Interact with trigger using left and right input in addition
+// to their regular counterparts
+inline bool DirectionalInputForTrigger = false;
+
```
PringlesGang: Not entirely happy with the location of these variables. Lmk if you have a better place to put them.

src/audio/audiosystem.h
```diff
@@ -18,5 +18,8 @@ inline float MasterVolume = 1.0f;
 inline float GroupVolumes[ACG_Count];
 inline AudioChannel* Channels[AC_Count];
 
+inline bool VoiceMuted[VoiceCount];
+inline float VoiceVolume[VoiceCount];
```
PringlesGang: Not entirely happy with the location of these variables. Lmk if you have a better place to put them.

src/games/cclcc/delusiontrigger.cpp
```diff
@@ -91,8 +92,15 @@ void DelusionTrigger::Update(float dt) {
           ScrWork[6336] = 0xffff;
         }
         if (LastDelusionState == DelusionState) {
-          if (!(PADinputButtonWentDown & PAD1L2)) {
-            if (PADinputButtonWentDown & PAD1R2) {
+          int padLeft = PAD1L2;
```
KKhanhH: It might be better to use the GetControlState function here and add something to delusion triggers for it. Also there's a padcustom element for right trigger + directional key, so we should use that instead

profiles/common/scriptvars.lua
```diff
@@ -9,6 +9,8 @@ root.ScriptVars = {
     SW_PLAYDATA_ALPHA=1131,
     SW_MUSICMODE_ALPHA=1132,
     SW_TOTALPLAYTIME=2001,
+    SW_CHARACTERIDMAPPING=2068,
+    SW_CHARACTERID=2100,
```
KKhanhH: 2100 should be SW_ANIME0CHANO, 2101 is SW_ANIME1CHANO, 2102 is SW_ANIME2CHANO

VERSION
```diff
@@ -1 +1 @@
-0.2.1.
\ No newline at end of file
+0.3.1.
```
KKhanhH: reset the patch number

src/games/cclcc/delusiontrigger.cpp
KKhanhH: I'm reworking the input system a little bit, so maybe hold off on this one

src/games/cclcc/optionsmenu.h
```diff
@@ -20,8 +23,22 @@ class OptionsMenu : public UI::OptionsMenu {
   void Update(float dt) override;
   void UpdateInput(float dt) override;
   void Render() override;
+  void ResetToDefault() override;
 
  private:
+  inline std::unique_ptr<Widgets::Group> CreateBasicPage(
```
KKhanhH: No need for inline

src/inputsystem.h
KKhanhH: Maybe it would be better to have all the user configurable options in one place like ConfigSystem or OptionsSystem class?

KKhanhH (changes requested):

src/inputsystem.h
PringlesGang: Oh yeah that's a great idea! Move 'em all (minus the group volumes) into an `Impacto::ConfigSystem` namespace in `src/configsystem.h`

profiles/common/scriptvars.lua
PringlesGang: Gotcha

VERSION
PringlesGang: Whoops

src/games/cclcc/optionsmenu.h
PringlesGang: I figured I might direct the compiler to inline, as they're all called only once in the constructor, but I wanted some cleaner code so moved them into separate functions

src/games/cclcc/delusiontrigger.cpp
PringlesGang: Oh yeah good idea. The `MousohTrigger` codes are 40 and 41, but in `GetControlState` 40 is already taken up by `CT_ResetOptions` for MO8, so maybe abstract some game-dependent stuff away in your reworking?

PringlesGang: Okay, rebased and it's now in `GetControlState` using the right `PadCustom` indices

KKhanhH (approved):
Tested with config saves

PR #107, opened 2025-02-01
Turbo on hold component

PringlesGang:
Adds a `TurboOnHoldHandler` that handles rapidfire events upon a (sequence of) buttons being held.


src/ui/widgets/cclcc/tipstabgroup.cpp
```diff
@@ -79,72 +84,84 @@ void TipsTabGroup::UpdatePageInput(float dt) {
...
-        if (checkScrollBounds()) {
-          ScrollPosY = TipsEntriesScrollbar->EndValue;
+    }
+
+    PageUpDownButtonHoldHandler.Update(dt);
+    const int pageUpDownShouldFire = PageUpDownButtonHoldHandler.ShouldFire();
+    const bool pageMovement = (bool)(pageUpDownShouldFire & PAD1RIGHT) ^
+                              (bool)(pageUpDownShouldFire & PAD1LEFT);
+
+    if (pageMovement) {
```
KKhanhH, line 124: Might be good to make vertical and horizontal changes exclusive, otherwise things might get messy if both get held down

KKhanhH (approved):

src/ui/widgets/cclcc/tipstabgroup.cpp
PringlesGang: Should be good now

PR #109, opened 2025-02-07
C;C LCC character fading

PringlesGang:
Implements character sprite fading support for C;C LCC.


profiles/cclcc/game.lua
```diff
@@ -18,6 +18,8 @@ root.Vm = {
     UseReturnIds = true,
     ScrWorkChaStructSize = 40,
     ScrWorkBgStructSize = 40,
+    ScrWorkCaptureStructSize = 20,
```
KKhanhH, line 21: Cclcc has differently sized capture struct to CC?

src/background2d.cpp
```diff
@@ -76,21 +76,86 @@ void Background2D::MainThreadOnLoad() {
   BgSprite.Bounds =
       RectF(0.0f, 0.0f, BgSpriteSheet.DesignWidth, BgSpriteSheet.DesignHeight);
   Show = false;
-  Layer = -1;
+  std::fill(Layers.begin(), Layers.end(), -1);
+}
+
+bool Background2D::OnLayer(int layer) {
```
KKhanhH: You can probably just define this in the header file, it's small enough

src/background2d.h
```diff
@@ -5,8 +5,7 @@
 #include "spritesheet.h"
 #include "loadable.h"
 
-#define BackgroundRenderer(name) \
-  void name(Background2D* bg, int bgId, glm::vec4 col)
+#define BackgroundRenderer(name) void name(Background2D* bg, glm::vec4 col)
```
KKhanhH: I wonder if we really need this macro now, especially since we can make these functions member functions and they only have one arg anyway

src/game.cpp
```diff
@@ -355,6 +396,17 @@ void Render() {
               }
             }
 
+            for (size_t capId = 0; capId <= 1; capId++) {
+              if (!GetFlag(SF_CAP1DISP + capId)) continue;
```
KKhanhH, line 403: Should it be 1 and not MaxScreenCaptures here

PringlesGang: Yup, missed that one

profiles/cclcc/game.lua
PringlesGang: Actually C;C's the one that needs to be 20 as well whoops

src/background2d.cpp
PringlesGang: Okay

src/background2d.h
PringlesGang: I think the issue is mostly with the function table at the bottom of the file. If we make them member functions, each instance would have to haul its own render table as well

We could also replace the function table with a delegation method

KKhanhH: The render table just needs to be a static variable inside the class, it actually shouldn't be static in the header in the first place, since that recreates the table in every translation unit, I'll push a change for it.

VERSION
```diff
@@ -1 +1 @@
-0.3.0.
+0.4.0.
```
KKhanhH: this is probably more of a .3.1 or something

src/background2d.cpp
```diff
@@ -76,82 +87,132 @@ void Background2D::MainThreadOnLoad() {
...
-    }
-    int renderType = ScrWork[SW_BG1FADETYPE + ScrWorkBgStructSize * bgId];
-    BackgroundRenderTable[renderType](this, bgId, col);
+      break;
+
+    case Vm::InstructionSet::CC:
+      if (ScrWork[SW_BGEFF1_MODE + ScrWorkBgEffStructSize * bgId] == 1)
+        col.a =
+            ScrWork[SW_BGEFF1_ALPHA + ScrWorkBgEffStructSize * bgId] / 256.0f;
+
```
KKhanhH, line 117: Probably want a break here

src/background2d.cpp
```diff
@@ -28,6 +28,17 @@ void Background2D::Init() {
     Screencaptures[i].IsScreencap = true;
   }
 
+  for (size_t i = 0; i < Framebuffers.max_size(); i++) {
+    Framebuffers[i].BgSprite =
+        Sprite(SpriteSheet(Profile::ResolutionWidth, Profile::ResolutionHeight),
```
KKhanhH: I think you want window height since profile resolutions are currently fixed

KKhanhH (changes requested):

PR #120, opened 2025-03-23
BgEff positioning

PringlesGang:
- Partially implements masked BgEff rendering, in case of rectangles
- Correctly positions BgEff based on rotation and scaling
- Implements Bg(Eff) tinting through the script
- Fixes C;H LCC wait icon not rendering
- Fixes early rounding on script trig functions
- Fixes C;C LCC backlog entry masked rendering

Does NOT implement proper polygonal masking


src/background2d.cpp
```diff
@@ -149,13 +149,16 @@ void Background2D::RenderBgEff(int bgId, int layer) {
...
-  DisplayCoords = glm::vec2(-x, -y);
+  DisplayCoords.x = ScrWork[SW_BGEFF1_POSX + structSize * bgId] +
+                    ScrWork[SW_BGEFF1_OFSX + 20 * bgId] -
+                    ScrWork[SW_BGEFF1_ORIGINX + structSize * bgId] -
+                    ScrWork[SW_BGEFF1_ORIGIN_OFSX + 20 * bgId];
+  DisplayCoords.y = ScrWork[SW_BGEFF1_POSY + structSize * bgId] +
+                    ScrWork[SW_BGEFF1_OFSY + 20 * bgId] -
+                    ScrWork[SW_BGEFF1_ORIGINY + structSize * bgId] -
+                    ScrWork[SW_BGEFF1_ORIGIN_OFSY + 20 * bgId];
+  DisplayCoords.x *= Profile::DesignWidth / 1280.0f;
```
KKhanhH: Don't think you want this

PringlesGang: With the fix you had before it was literally printing off-screen
This both reflects the decomp and matches up pixel-perfect with PS4 screenshots

KKhanhH: That's strange, on my end it would show only on a smaller top left section of the screen. Which decompile are you referencing, android's? Android does the division thing since they have a smaller resolution than the design res.

PringlesGang: I'm referencing the LCC PS4 binary directly, as C;C LCC seems to be the only game that has such a complicated BgEff class. From what I understood ScrWork just kinda always works in 720p, regardless of the actual target game's resolution

PringlesGang: For example, this is what I'm getting right now with impacto, whereas before it would render the framebuffer off-screen
impacto

For reference, this is what it should look like according to the PS4
ps4

KKhanhH: Ah I see, that's definitely not very intuitive then

PringlesGang: At least that's what this wisdom of the ancients leads me to believe. MAGES certainly made an engine alright
https://github.com/CommitteeOfZero/impacto/blob/cdea02891a69d0711568446537afb5eac5e5f3db/src/vm/interface/scene2d.cpp#L37-L38

KKhanhH: too lazy to change position in scripts across platforms i guess

PringlesGang:
Fyi this still doesn't look great at all through Vulkan and DX9, mostly because frame buffers aren't implemented for those yet

src/background2d.cpp
```diff
@@ -130,92 +134,134 @@ void Background2D::RenderCapture(int capId, int layer) {
...
-  std::invoke(BackgroundRenderTable[renderType], this, col);
+  std::invoke(BackgroundRenderTable[renderType], this);
 }
 
 void Background2D::RenderBgEff(int bgId, int layer) {
   if (Status != LS_Loaded) return;
 
-  const int structSize = ScrWorkBgEffStructSize;
+  const int structOffset = ScrWorkBgEffStructSize * bgId;
+  const int structOfsOffset = 20 * bgId;
```
KKhanhH: Move OfsSize into a lua variable

src/background2d.cpp
```diff
@@ -130,92 +134,134 @@ void Background2D::RenderCapture(int capId, int layer) {
...
-  col.a = (ScrWork[SW_CAP1ALPHA + ScrWorkCaptureStructSize * capId] +
-           ScrWork[SW_CAP1ALPHA_OFS + 10 * capId]) /
-          256.0f;
+  Tint = glm::vec4(255.0f);
+  Tint.a = ScrWork[SW_CAP1ALPHA + ScrWorkCaptureStructSize * capId] +
+           ScrWork[SW_CAP1ALPHA_OFS + 10 * capId];
+  if (GameInstructionSet == +Vm::InstructionSet::CC) {
+    const uint32_t rgb =
+        ScrWork[SW_CAP1FILTER + ScrWorkCaptureStructSize * capId];
+    Tint = {rgb & 0xff, (rgb >> 8) & 0xff, (rgb >> 16) & 0xff, Tint.a};
```
KKhanhH: Does ScrWorkGetColor work here? or is it out of order

src/background2d.cpp
```diff
@@ -130,92 +134,134 @@ void Background2D::RenderCapture(int capId, int layer) {
...
 
-  glm::vec4 col = glm::vec4(1.0f);
-  col.a = (ScrWork[SW_CAP1ALPHA + ScrWorkCaptureStructSize * capId] +
-           ScrWork[SW_CAP1ALPHA_OFS + 10 * capId]) /
-          256.0f;
+  Tint = glm::vec4(255.0f);
+  Tint.a = ScrWork[SW_CAP1ALPHA + ScrWorkCaptureStructSize * capId] +
+           ScrWork[SW_CAP1ALPHA_OFS + 10 * capId];
+  if (GameInstructionSet == +Vm::InstructionSet::CC) {
+    const uint32_t rgb =
```
KKhanhH: There is also probably another condition or something here, the tint is what's killing the startup bgs

KKhanhH (changes requested):

src/background2d.cpp
PringlesGang: Didn't even know that existed. Definitely should use that instead

src/background2d.cpp
PringlesGang: I see...

PringlesGang: Fixed it; the default ScrWork value wasn't set

src/background2d.cpp
PringlesGang: Done

KKhanhH (approved):

PR #122, opened 2025-04-01
Help Menu Implementation

Quessian-Phaedrus-Sternabite:
Very rudimentary help menu implementation created after some time spent looking through the existing code. I was unsure about how to add images, but at the very least, a menu now exists for things to be added to.


src/games/cclcc/titlemenu.cpp
```diff
@@ -46,8 +46,12 @@ void TitleMenu::ContinueButtonOnClick(Widgets::Button* target) {
 }
 
 void TitleMenu::ExtraButtonOnClick(Widgets::Button* target) {
+<<<<<<< Updated upstream
```
KKhanhH, line 49: committed some merge conflict stuff here, probably should drop the merge commit and rebase from master instead

PR #123, opened 2025-04-05
C;CLCC Clearlist Implementation

Enorovan:
Full Implementation of the C;CLCC Clearlist


src/profile/games/cclcc/clearlistmenu.h
```diff
@@ -7,14 +7,20 @@ namespace Profile {
...
 namespace ClearListMenu {
 
+int constexpr Endings = 10;
+
 inline Sprite ClearListBookLayerSprite;
+inline Sprite ClearListCompleteBookLayerSprite;
 inline Sprite ClearListGuideSprite;
-inline Sprite LibraryMaskSprite;
+inline Sprite ClearListMaskSprite;
+
```
KKhanhH, line 16: Can you use std::array<Sprite, Endings> instead?

Enorovan: I don't know the syntax for the GetMember and it's done like that in all other instances

KKhanhH: Ah ok, I guess i'll just do an update of every profile array later then

VERSION
```diff
@@ -1 +1 @@
-0.3.4.
+0.4.4.
```
KKhanhH: Reset minor to 0

src/games/cclcc/clearlistmenu.cpp
```diff
@@ -49,43 +47,93 @@ void ClearListMenu::Hide() {
...
+    2,  // 804
+    5,  // 805
+    3,  // 806
+    7,  // 807
+    8,  // 808
+    0,  // 809
+    9   // 810
+};
+
+void ClearListMenu::DrawEndingSprites(glm::vec4 transition) {
```
KKhanhH: this can probably be a const ref

PR #155, opened 2025-04-17
Cclcc/save checksum

KKhanhH:
Adds checksums to saves to get backwards hardware save compatibility

Handles #149


KKhanhH:
Possible issues with loading a save and entering black screen if loading save on real hardware? Might depend on save file.

PR #161, opened 2025-04-22
Redo OpenAL implementation

newo-2001:
This PR re-implements the OpenAL AudioChannel.

I ran into some audio related bugs whilst chipping away at the CHLCC intro sequence, but the original implementation was holding a lot of state and was hard to reason about.
I re-implemented it from scratch, keeping most of the logic roughly the same and fixing the problem I was facing in the process.
I also took the opportunity to modernize it a bit and do some cleanup.


KKhanhH:
Coulr you undo the rename or have it in a separate commit? Git treats them as completely separate files whichakes the review harder

src/audio/openal/openalaudiochannel.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+void OpenALAudioChannel::Play(AudioStream* stream, bool loop,
+                              float fadeInDuration) {
+  EndPlayback();
+
+  Stream = stream;
+  FadeDuration = fadeInDuration;
+  FadeProgress = FadeDuration == 0 ? 1 : 0;
+  State = FadeDuration == 0 ? ACS_Playing : ACS_FadingIn;
+}
```
PringlesGang: You're not updating the loop state here

src/audio/openal/openalaudiochannel.cpp
```diff
@@ -0,0 +1,237 @@
...
+void OpenALAudioChannel::Stop(float fadeOutDuration) {
+  if (fadeOutDuration == 0) {
+    EndPlayback();
+    return;
+  }
+
+  State = ACS_FadingOut;
+  FadeDuration = fadeOutDuration;
+  FadeProgress = 0;
+}
```
PringlesGang: What happens here when the stream is already stopped? Will it fade out again from full volume?

PringlesGang (changes requested):

PringlesGang:
Also, perhaps you could change the enums in `audiocommon.h` to enum classes while you're at it?

src/audio/openal/openalaudiochannel.cpp
newo-2001: Nice catch, I noticed a few minutes ago that the menu music wasn't looping, guess I was stupid.

src/audio/openal/openalaudiochannel.cpp
newo-2001: You're right better guard it, but I guess if it was still fading out that seems like the reasonable thing to do.

newo-2001:
It doesn't seem to work properly for streams starting at an offset, I'll address that tomorrow.

src/audio/openal/openalaudiochannel.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+  alSourcef(Source, AL_PITCH, 1);
+  alSource3f(Source, AL_POSITION, 0, 0, 0);
+  alSource3f(Source, AL_VELOCITY, 0, 0, 0);
+  alSourcei(Source, AL_LOOPING, AL_FALSE);
+  UpdateGain();
+}
+
+OpenALAudioChannel::~OpenALAudioChannel() {
+  if (Stream) delete Stream;
```
KKhanhH: Probably good to just make stream a unique_ptr as well

newo-2001: Tried that, but it gave me some issues converting raw pointers to the base class at the call site due to incomplete types, I believe the change would be more intricate than initially anticipated.

newo-2001:
I believe that is the issue I mentioned earlier because for me CHLCC seemed to work fine. But when I checked CCLCC it didn't work properly. I quickly noticed that CLCC's streams sometimes start with an offset. The code assumed streams would always start at 0 but that is not the case, I think fixing that will make it behave properly.

KKhanhH:
> > Also, perhaps you could change the enums in `audiocommon.h` to enum classes while you're at it?
> 
> Just attempted this, but it made the code way less readable in its current state due to line wrapping literally everywhere. I'll leave that for later.

It might help to remove ACS_ prefixes and whatnot if you didn't already do that. You can also use scoped "using enum" statements where you need  to since we're on c++20 now

src/audio/openal/openalaudiochannel.cpp
KKhanhH: Can you not just include audiostream in audiochannel header?

newo-2001:
That sounds like it might make it more manageable, but I'll do that separately to keep the PR's focused.

Also, I believe I did manage to overcome the incomplete type issue that was preventing me to use `unique_ptr`.

src/audio/openal/openalaudiochannel.cpp
newo-2001: Yeah, I was really confused by the error message, I thought the type itself had some fields that were forward declared causing it to be incomplete, but `AudioStream` itself was just forward declared in `audiocommon.h`...

KKhanhH: yeah, if you use unique_ptr with a forward declared class, then you have to make sure the destructor for the class is defined in the cpp file since it can't see the pointed-to-class's destructor. Easiest to just include it if there's nothing circular.

PringlesGang (changes requested):
When loading another save while a voice line is still playing, there's a small blip of the line that is still played during the loading. Probably something wrong with `EndPlayback()`?

KKhanhH:
> When loading another save while a voice line is still playing, there's a small blip of the line that is still played during the loading. Probably something wrong with `EndPlayback()`?

It looks like it's in the paused state, then resumes and so plays a bit before it does a stop.

newo-2001:
The problem was that the script called Pause() upon opening the menu. Following with Resume() and Play() in the same frame when loading. This caused the paused stream to play until the script reached the Play() call.

I updated the code to not immediately perform the state transitions when the functions are called and instead only queue them for when the audio channel is updated.

PringlesGang (approved):

KKhanhH (approved):
LGTM, just make sure it's rebased on top of master and we can merge it

PR #162, opened 2025-04-22
Polygonal masking

### LINKED ISSUE #133: C;CLCC: Fix fade to white blinking to black for a few frames

PringlesGang:
Implements the polygonal masking as is often used in C;C LCC to render characters in a trianglular or quadrangular frame.

Also:
- Fixes #133


src/background2d.cpp
```diff
@@ -202,16 +202,21 @@ void Background2D::RenderBgEff(int bgId, int layer) {
 
   const size_t vertexCount = maskType == 0 ? 4 : 3;
   std::vector<glm::vec2> vertices;
```
KKhanhH: this could probably just be an array if it's fixed to 3/4 elements

src/renderer/opengl/window.cpp
```diff
@@ -316,6 +317,13 @@ void GLWindow::Update() {
     glFramebufferTexture2D(GL_READ_FRAMEBUFFER, GL_COLOR_ATTACHMENT0,
                            GL_TEXTURE_2D, ReadRenderTexture, 0);
 
+    glGenRenderbuffers(1, &StencilBuffer);
+    glBindRenderbuffer(GL_RENDERBUFFER, StencilBuffer);
+    glRenderbufferStorage(GL_RENDERBUFFER, GL_STENCIL_INDEX8, WindowWidth,
+                          WindowHeight);
+    glNamedFramebufferRenderbuffer(ReadRT, GL_STENCIL_ATTACHMENT,
```
KKhanhH: I don't think this function exists in openglES

src/renderer/renderer.cpp
```diff
@@ -79,6 +81,57 @@ void BaseRenderer::DrawCCMessageBox(Sprite const& sprite, Sprite const& mask,
...
+  if (vertices.size() == 3) {
+    DrawVertices(RectSprite.Sheet, vertexAttributes,
+                 std::array<uint16_t, 3>{0, 1, 2});
+    return;
+  }
+
+  std::vector<uint16_t> indices;
+  indices.reserve(vertices.size() * 3);
+
+  glm::vec2 centerOfMass = {0.0f, 0.0f};
```
KKhanhH: could also just std::acumulate / size

src/util.cpp
```diff
@@ -206,7 +206,7 @@ glm::mat4 TransformationMatrix(const glm::vec3 scalingOrigin,
   }
 
   if (rotation != glm::quat()) {
-    matrix *= glm::mat4_cast(normalize(rotation));
+    matrix = matrix * glm::mat4_cast(glm::normalize(rotation));
```
KKhanhH, line 209: what was wrong here

src/util.h
```diff
@@ -382,8 +382,8 @@ inline float NormalizeRad(float rad) {
   if (rad < 0) rad += 2.0f * (float)M_PI;
   return rad - (float)M_PI;
 }
-constexpr float ScrWorkAngleToRad(float angle) {
-  return angle * (float)(2.0f * M_PI / (float)(1 << 16));
+constexpr float ScrWorkAngleToRad(int angle) {
+  return ((float)angle * 2.0f * (float)M_PI) / (float)(1 << 16);
```
KKhanhH: maybe we should use std::number instead of macro + cast

src/background2d.cpp
PringlesGang: It could yeah, but then I'd have to do a manual subspan construction when passing it to `DrawConvexShape`, so I think this is a little cleaner overall, as the size of the array does matter.

src/renderer/opengl/window.cpp
PringlesGang: What OpenGL versions do we need to support?

src/util.cpp
PringlesGang: Nothing really, just wanted to be a little more explicit

src/background2d.cpp
KKhanhH: Subspan is fine, id rather not do a heap alloc for something like this

src/renderer/opengl/window.cpp
KKhanhH: GL3 and ES3.2 is what we request right now

src/renderer/renderer.cpp
PringlesGang: Used reduce instead to avoid having to add the redundant `glm::vec2(0.0f)`

PR #165, opened 2025-04-27
Renderer transformation refactor

PringlesGang:
A refactor of the renderer to make more use of generic transformation matrices instead of a myriad purpose-built overloads.
Also adds useful functions for generating these transformation matrices, and reduces code duplication within and across the renderer implementations.


src/renderer/renderer.h
```diff
@@ -44,22 +44,77 @@ class BaseRenderer {
   virtual void FreeTexture(uint32_t id) = 0;
   virtual YUVFrame* CreateYUVFrame(float width, float height) = 0;
 
+  virtual void DrawSprite(const Sprite& sprite, const CornersQuad& dest,
+                          std::span<const glm::vec4, 4> tints,
+                          bool inverted = false) = 0;
+  void DrawSprite(const Sprite& sprite, glm::mat4 transformation,
```
KKhanhH, line 72: Maybe default this to the identity matrix

PringlesGang: No, I want transformation to be a decision. With a default identity it would still do the redundant matrix multiplication

src/util.cpp
```diff
@@ -125,4 +165,9 @@ RectF RectF::Coalesce(const RectF& first, const RectF& second) {
   return rect;
 }
 
+glm::quat ScrWorkAnglesToQuaternion(const int x, const int y, const int z) {
```
KKhanhH: This can probably be inlined in header since its so small

profiles/common/scriptvars.lua
```diff
@@ -78,6 +78,8 @@ root.ScriptVars = {
     SW_BGEFF1_OFSY=2801,
     SW_BGEFF1_ALPHA_OFS=2806,
```
Enorovan: Maybe fix ordering there (and in src/scriptvars.h)

PringlesGang:
Best of luck Dext; you're gonna need it o7

src/background2d.cpp
```diff
@@ -222,6 +226,15 @@ void Background2D::RenderBgEff(int bgId, int layer) {
   Origin = {0.0f, 0.0f};
   for (const glm::vec2 vertex : vertices) Origin += vertex / (float)vertexCount;
 
+  // Transform vertices
+  const glm::mat4 transformation =
+      TransformationMatrix({Origin, 0.0f}, {Scale, 1.0f}, {Origin, 0.0f},
```
KKhanhH: I think having a Transformation matrix overload for vec2 versions with quart might not be a bad idea, repeatedly typing in orign, 0 and scale, 1 is probably mistake prone.

src/character2d.cpp
```diff
@@ -178,9 +188,9 @@ void Character2D::Update(float dt) {
         Character2DState const& state = stateItr->second;
         auto& stateIndices =
             std::get_if<Character2DState::MVLData>(&state.Data)->Indices;
-        std::copy(stateIndices.get(), stateIndices.get() + state.Count,
-                  MvlIndices.begin() + MvlIndicesCount);
-        MvlIndicesCount += state.Count;
+        MvlIndices.reserve(MvlIndices.size() + state.Count);
```
KKhanhH: Reserve isn't needed here since insert should do that

src/character2d.h
```diff
@@ -86,10 +84,8 @@ class Character2D : public Loadable<Character2D> {
   ankerl::unordered_dense::map<int, Character2DState> States;
   std::vector<int> StatesToDraw;
 
-  int MvlVerticesCount;
-  std::vector<float> MvlVertices;
-  std::array<uint16_t, MaxMvlIndices> MvlIndices;
-  int MvlIndicesCount;
+  std::vector<VertexBufferSprites> MvlVertices;
+  std::vector<uint16_t> MvlIndices;
```
KKhanhH, line 88: Have you tested if there's any perf hit to using vector instead of array here?

src/games/cclcc/delusiontrigger.cpp
```diff
@@ -184,166 +184,189 @@ void DelusionTrigger::Update(float dt) {
...
+
   if ((ScrWork[SW_DELUSION_LIMIT] == Delusion_Both) ||
       (ScrWork[SW_DELUSION_LIMIT] == Delusion_NegOnly)) {
-    float topLeftX =
-        2944.0f - (float)((ScrWork[SW_DELUSION_SPIN_COUNTER] * 1248) >> 5);
-    Renderer->DrawSpriteOffset(
-        NegativeDelusionSprite, glm::vec2{topLeftX, -109.0f},
-        glm::vec2{424.0f, 557.0f},
-        {1.0f, 1.0f, 1.0f, (float)spinAlpha / 256.0f}, {1, 1}, spinAngle);
+    const glm::vec2 topLeft = {
```
KKhanhH: These probably shouldn't be named topLeft or center and more like offset or something, I can't remember if these names were actually accurate or not or something i just put in.

src/games/chlcc/delusiontrigger.cpp
```diff
@@ -334,17 +334,18 @@ void DelusionTrigger::Render() {
...
   Renderer->DrawSprite(
       ScreenMask, RectF(0.0f, 0.0f, 1280.0f, 720.0f),
       glm::vec4(1.0f, 1.0f, 1.0f, (backgroundAlpha * 160) / 65536.0));
 
-  Renderer->DrawMaskedSpriteOverlay(
-      BackgroundSprite, ScaledMask, RectF(0.0f, 0.0f, 1280.0f, 720.0f),
-      glm::vec4{1.0f}, (backgroundAlpha * 160) >> 8, 20, true,
-      spinAngle * 2 * (float)M_PI / 65536.0f);
+  CornersQuad dest = RectF(0.0f, 0.0f, 1280.0f, 720.0f)
+                         .Rotate(ScrWorkAngleToRad(spinAngle), {640, 360});
```
KKhanhH: It doesn't look like the right thing is spinning here. The mask should be spinning, not the bg.

src/renderer/dx9/renderer.cpp
```diff
@@ -402,83 +292,45 @@ void Renderer::DrawSprite(Sprite const& sprite, CornersQuad const& dest,
...
+  alpha = std::clamp(alpha, 0, fadeRange + 256);
+  const float alphaRange = 256.0f / fadeRange;
+  const float constAlpha = ((255.0f - alpha) * alphaRange) / 255.0f;
 
   // Do we have space for one more sprite quad?
   EnsureSpaceAvailable(4, sizeof(VertexBufferSprites), 6);
 
-  Flush();
-  EnsureShader(ShaderMaskedSprite);
+  std::array<float, 4> alphaRes = {alphaRange, constAlpha};
```
KKhanhH: Should be 2? Or just remove the template arguments and let deduction handle it.

src/renderer/dx9/renderer.cpp
```diff
@@ -402,83 +292,45 @@ void Renderer::DrawSprite(Sprite const& sprite, CornersQuad const& dest,
...
-  if (!Drawing) {
-    ImpLog(LogLevel::Error, LogChannel::Render,
-           "Renderer->DrawMaskedSprite() called before BeginFrame()\n");
+           "Renderer->DrawMaskedSpriteOverlay() called before BeginFrame()\n");
     return;
   }
 
-  if (alpha < 0) alpha = 0;
-  if (alpha > fadeRange + 256) alpha = fadeRange + 256;
+  if (sprite.Sheet.IsScreenCap) Flush();
```
KKhanhH, line 317: Previous version always did flush for mask, is change intended?

src/renderer/dx9/renderer.cpp
```diff
@@ -490,31 +342,28 @@ void Renderer::DrawMaskedSprite(Sprite const& sprite, Sprite const& mask,
 
   IndexBufferFill += 6;
 
-  QuadSetUV(sprite.Bounds, sprite.Sheet.DesignWidth, sprite.Sheet.DesignHeight,
-            (uintptr_t)&vertices[0].UV, sizeof(VertexBufferSprites));
-  QuadSetUV(sprite.Bounds, sprite.Bounds.Width, sprite.Bounds.Height,
-            (uintptr_t)&vertices[0].MaskUV, sizeof(VertexBufferSprites));
+  QuadSetUV(sprite.Bounds, sprite.Sheet.GetDimensions(), &vertices[0].UV,
+            sizeof(VertexBufferSprites));
+  QuadSetUV(mask.Bounds, mask.Sheet.GetDimensions(), &vertices[0].MaskUV,
```
KKhanhH: Behavior here seems to be different

src/renderer/dx9/renderer.cpp
```diff
@@ -524,20 +373,25 @@ void Renderer::DrawMaskedSpriteOffset(const Sprite& sprite, const Sprite& mask,
   // Do we have space for one more sprite quad?
   EnsureSpaceAvailable(4, sizeof(VertexBufferSprites), 6);
 
-  Flush();
-  // TODO: handle inverted sprites
-  EnsureShader(ShaderMaskedSprite);
+  std::array<float, 4> alphaRes = {alphaRange, constAlpha};
```
KKhanhH: Same as the other comment

src/video/ffmpegplayer.cpp
```diff
@@ -476,13 +476,10 @@ void FFmpegPlayer::Update(float dt) {
...
-    float heightScale = Profile::DesignHeight / VideoTexture->Height;
-    glm::vec4 tint = glm::vec4(1.0f);
-    tint.a = videoAlpha;
-    Renderer->DrawVideoTexture(VideoTexture, glm::vec2(0.0f, 0.0f), tint,
-                               glm::vec2(widthScale, heightScale), 0.0f,
-                               IsAlpha);
+    const RectF dest = {0.0f, 0.0f, Profile::DesignWidth,
+                        Profile::DesignHeight};
+    const glm::vec4 tint = {1.0f, 1.0f, 1.0f, videoAlpha};
+    Renderer->DrawVideoTexture(*VideoTexture, dest, tint, IsAlpha);
```
KKhanhH, line 482: Does this not scale by the videotexture width and height?

src/util.cpp
```diff
@@ -5,6 +5,149 @@
...
+  for (glm::vec4& vertex : vertices) {
+    vertex -= glm::vec4(origin, 0.0f);
+    vertex = rotationMatrix * vertex;
+    vertex += glm::vec4(origin, 0.0f);
+  }
+
+  // Move into screen
+  if (stayInScreen) {
+    float maxZ = 0.0f;
+    for (const glm::vec4 vertex : vertices) maxZ = std::max(maxZ, vertex.z);
```
KKhanhH: std::max_element if you want the std algorithm for it

KKhanhH (changes requested):

src/renderer/dx9/renderer.cpp
PringlesGang: `SetPixelShaderConstantF` expects an array of a multiple for four floats, so we should make room for that to avoid undefined behaviour. I'll explicitly set them to 0.0f now to avoid confusion

src/background2d.cpp
PringlesGang: Especially since scaling is done before rotation, you're right

src/renderer/dx9/renderer.cpp
PringlesGang: Nice catch

src/video/ffmpegplayer.cpp
PringlesGang: The old version scaled the texture by a factor of `Profile::DesignDimensions / VideoTexture->Dimensions`, meaning it would render to a quad with dimensions `VideoTexture->Dimensions * (Profile::DesignDimensions / VideoTexture->Dimensions = Profile::DesignDimensions`.
In the new version I've just taken out the computation, and render to a quad the size of `Profile::DesignDimensions` directly

src/util.cpp
PringlesGang: Ooh that's a useful function

src/games/chlcc/delusiontrigger.cpp
PringlesGang: Should be fixed

src/renderer/dx9/renderer.cpp
PringlesGang: Not indended, thanks

src/character2d.h
PringlesGang: There is, going from ~200fps to ~80fps in debug mode with three characters on screen, but frankly we should be doing these transformations in the vertex shader anyway...
Lemme know if I should look into that in this PR already or if this can wait until the full-scale refactor

KKhanhH: We should leave it as an array for now then and move the transformation into the vertex shader later

PringlesGang: Oh wait I think I misunderstood your comment. I think most of the performance hit came from the matrix multiplication; I doubt the switch to vector really did anything noticable. Could profile a little more closely later

src/background2d.cpp
```diff
@@ -222,6 +226,14 @@ void Background2D::RenderBgEff(int bgId, int layer) {
   Origin = {0.0f, 0.0f};
   for (const glm::vec2 vertex : vertices) Origin += vertex / (float)vertexCount;
 
+  // Transform vertices
+  const glm::mat4 transformation = TransformationMatrix(
```
KKhanhH, line 230: You can use the vec2 overload here

KKhanhH: oh is rotation origin and translation origin still only vec3? i feel like it might as well be completely vec2 if the others are vec2

src/character2d.h
KKhanhH: Ah okay, you should also probably do profiling in optimized build, matrix multiplication will probably be a lot faster there. There's also probably some SIMD that gets done by default too

src/background2d.cpp
PringlesGang: The 3d rotation should still definitely be a vec3, because with the rotation you're already working in 3d, and might have to supply a depth in the rotational origin, but the translation could indeed just kinda be reduced to a vec2

KKhanhH:
Vulkan seems to be artifacting a lot, it's especially noticable on chlcc, but I'm not sure if it was always like that

PringlesGang:
> Vulkan seems to be artifacting a lot, it's especially noticable on chlcc, but I'm not sure if it was always like that

Yeah, Vulkan is very jittery on master too, so I tried not to make things any worse, but also didn't really make an effort to make it much better either. Would need to do a deep dive into learning Vulkan for that at some point

src/character2d.h
PringlesGang: Yeah okay, I did some more profiling and it looks to me like the vectorization doesn't really seem to be an impact at all (as in, the loading of the vertices is so fast visual studio can't even process anything meaningful).

I've also checked the matrix multiplications in in release mode, and the fps in the cclcc system menu (which thus far is the worst case scenario I could find) goes from ~800fps on master down to ~340fps, which is a perf hit, _but it's still 340fps_, and again should be moved to the vertex shader anyway

KKhanhH: Yeah we should just move it to the vertex shader. I'd like to see what the numbers are on a worse device like a steam deck though, I assume your computer is probably pretty beefy, so if the perf is a 60% hit on a crappy device it might be a lot worse

src/games/cclcc/systemmenu.cpp
```diff
@@ -234,66 +164,65 @@ void SystemMenu::Render() {
...
+      return glm::vec2(transformedVertex);
     };
 
-    TransformImage(bgSpriteBounds, bgDisp, angleX, angleY, angleZ, scale,
-                   BGTranslationOffset, SpriteGridVertices,
-                   DisplayGridVertices);
-    Renderer->DrawVertices(SystemMenuBG.Sheet, SpriteGridVertices,
-                           DisplayGridVertices, 21, 11);
+    const CornersQuad bgDisp =
+        CornersQuad(BGDispOffsetTopLeft, BGDispOffsetBottomLeft,
```
KKhanhH: brace initialize instead

src/renderer/dx9/renderer.h
```diff
@@ -34,66 +34,42 @@ class Renderer : public BaseRenderer {
...
+  void DrawMaskedSpriteOverlay(const Sprite& sprite, const Sprite& mask,
+                               const CornersQuad& spriteDest,
+                               const CornersQuad& maskDest, int alpha,
+                               int fadeRange, glm::mat4 spriteTransformation,
+                               glm::mat4 maskTransformation,
+                               std::span<const glm::vec4, 4> tints,
+                               bool isInverted, bool useMaskAlpha) override;
+
+  void DrawVertices(const SpriteSheet& sheet,
+                    std::optional<const SpriteSheet> mask,
```
KKhanhH: This will always be a copy, for a const object it would make more sense to get a reference somehow, pointer or maybe a typedefed optional<reference_wrapper>?

KKhanhH: You could also just do default constructible spritesheet and validate it using texture id or something

src/renderer/opengl/renderer.h
```diff
@@ -125,68 +93,82 @@ class Renderer : public BaseRenderer {
...
 
-  Sprite RectSprite;
+  uint16_t NextFreeIndex = 0;
+
+  const glm::mat4 Projection =
+      glm::ortho(0.0f, Profile::DesignWidth, Profile::DesignHeight, 0.0f);
 
   // ShaderCompiler compiler
-  ShaderCompiler* Shaders;
+  std::shared_ptr<ShaderCompiler> Shaders;
```
KKhanhH: The lifetime of Shaders isn't unknown or ambiguous, so a shared_ptr isn't necessary here. Scene3D is the only user and it's a member of renderer, so its lifetime will be tied to renderer. Try to avoid shared_ptr unless necessary cuz it does have a perf hit with reference counting + atomic operations.

KKhanhH: I'd say just stick to unique_ptr (owner) + ref/raw ptr(users) as a general pattern, but here i think you might be able to just store the value itself and pass references to children

src/renderer/opengl/renderer.h
```diff
@@ -125,68 +93,82 @@ class Renderer : public BaseRenderer {
...
-  void QuadSetUVFlipped(RectF const& spriteBounds, float designWidth,
-                        float designHeight, uintptr_t uvs, int stride);
-  void QuadSetPosition(RectF const& transformedQuad, float angle,
-                       uintptr_t positions, int stride);
-  void QuadSetPosition(CornersQuad destQuad, float angle, uintptr_t positions,
-                       int stride);
-  void QuadSetPosition3DRotated(RectF const& transformedQuad, float depth,
-                                glm::vec2 vanishingPoint, bool stayInScreen,
-                                glm::quat rot, uintptr_t positions, int stride);
+  std::unique_ptr<SpriteShader> SpriteShaderProgram;
```
KKhanhH: These are all different types and can just be stored as values

KKhanhH: Use optional if late initialization is needed

src/renderer/opengl/shader.h
```diff
@@ -74,5 +74,222 @@ class ShaderCompiler {
...
+  void UploadUniforms(SpriteUniforms uniforms) override;
+
+ private:
+  const GLint ProjectionLocation;
+  const GLint TransformationLocation;
+
+  const GLint TexturesLocation;
+};
+
+struct SpriteInvertedUniforms {
```
KKhanhH, line 141: Is this struct necessary

src/renderer/opengl/shader.h
```diff
@@ -74,5 +74,222 @@ class ShaderCompiler {
...
+  const GLint LumaLocation;
+  const GLint CbLocation;
+  const GLint CrLocation;
+  const GLint IsAlphaLocation;
+};
+
+struct MaskedSpriteNoAlphaUniforms {
+  bool operator==(const MaskedSpriteNoAlphaUniforms& other) const = default;
+
+  glm::mat4 Projection = glm::mat4();
```
KKhanhH: just do glm::mat4 Projection{}

src/renderer/opengl/renderer.cpp
```diff
@@ -264,489 +223,390 @@ YUVFrame* Renderer::CreateYUVFrame(float width, float height) {
...
+
+  if (IndexBuffer.size() + indices.size() > MaxIndexCount) Flush();
+
+  const uint16_t maxIndex = *std::max_element(indices.begin(), indices.end());
+
+  VertexBuffer.insert(VertexBuffer.end(), vertices.begin(), vertices.end());
+
+  const size_t indexOffset = IndexBuffer.size();
+  IndexBuffer.resize(IndexBuffer.size() + indices.size());
+  std::transform(
```
KKhanhH, line 235: you can avoid the resize with back_inserter on IndexBuffer.end()

src/renderer/opengl/renderer.h
```diff
@@ -125,68 +93,82 @@ class Renderer : public BaseRenderer {
...
-
   bool Drawing = false;
 
-  static int constexpr VertexBufferSize = 1024 * 1024;
-  static int constexpr IndexBufferCount =
-      VertexBufferSize / (4 * sizeof(VertexBufferSprites)) * 6;
+  static constexpr size_t TextureUnitCount = 15;
+  static constexpr std::array<GLint, TextureUnitCount> Textures = {
+      0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14};
+  std::array<std::unique_ptr<TextureUnit>, TextureUnitCount> TextureUnits;
```
KKhanhH: Does this need to be unique_ptrs? You might be able to store TextureUnits or optional TextureUnits instead

src/renderer/opengl/textureunit.h
```diff
@@ -0,0 +1,72 @@
...
+
+  GLuint GetSampler() const { return Sampler; }
+
+  void Flush() {
+    if (Info.use_count() == 1) Info->Flushed = true;
+  }
+  void Reserve() { Info->Flushed = false; }
+
+ private:
+  const GLuint Unit;
```
KKhanhH: If it's meant to be or copyable, maybe don't have member variables as const since those kill the assignment operators

src/renderer/renderer.cpp
```diff
@@ -114,162 +120,250 @@ void BaseRenderer::DrawProcessedText(std::span<const ProcessedTextGlyph> text,
...
-      glm::vec4 color = RgbIntToFloat(text[i].Colors.OutlineColor);
-      color.a = outlineOpacity;
-      if (smoothstepGlyphOpacity) {
-        color.a *= glm::smoothstep(0.0f, 1.0f, text[i].Opacity);
-      } else {
-        color.a *= text[i].Opacity;
+    // Add outline to the front of the buffers
+    vertices.insert(vertices.end(), vertices.begin(), vertices.end());
+
+    indices.resize(indexCount * 2);
```
KKhanhH: back_inserter

KKhanhH (changes requested):

src/renderer/opengl/shader.h
PringlesGang: Not strictly, but I thought it useful to give it its own struct anyway in case `SpriteInverted` ever has to divert from `Sprite`

src/renderer/opengl/renderer.h
PringlesGang: `TextureUnit`s don't have a default constructor, so I needed late initialization. Optional is possible tho

src/renderer/opengl/renderer.cpp
PringlesGang: Won't that continuously push back all the elements, possibly expanding multiple times, instead of resizing just once?

KKhanhH: Yeah you can reserve in that case for the allocation, using back_inserter lets you avoid the default construction

src/renderer/dx9/renderer.h
KKhanhH: I'm thinking it might be better to just make it a pointer instead for consistency with everything else (also the draw text functions are already nullable pointers)

KKhanhH (approved):

PR #178, opened 2025-06-13
Warning fixes

PringlesGang:
Aims to fix most warnings in the repo. (At least enough to the point that we can actually enable warnings again)


KKhanhH:
Might as well have warnings as errors so we don't run into the same warnings

src/io/lzxstream.cpp
```diff
@@ -72,7 +72,7 @@ int64_t LzxStream::Read(void *buffer, int64_t sz) {
 }
 
 int64_t LzxStream::Seek(int64_t offset, int origin) {
-  int64_t absPos;
+  int64_t absPos = offset;
```
KKhanhH: use immediately invoked lambda idiom instead, i.e. 
int64_t absPos = [&]{
  switch (origin) {
    case RW_SEEK_SET: return offset;
    ...
  }
}();

PringlesGang: Wouldn't it then just start whining about a possible non-value-returning code path instead? We have to pick a default

src/ui/widgets/group.cpp
```diff
@@ -36,6 +36,7 @@ void Group::Add(Widget* widget, FocusDirection dir) {
     case FDIR_UP:
       oppositeDir = FDIR_DOWN;
       break;
+    default:
```
KKhanhH: let's avoid default cases whenever possible, better warnings/errors when you miss an enum case

PringlesGang: Fair point

src/io/lzxstream.cpp
KKhanhH: in that case we can just default it to an error or 0 then, it's probably better to make that an origin an enum instead of an int though

KKhanhH: this way you can also make it const too

KKhanhH: Normally in cases like this I just throw an error, haven't tested how well that works on switch though.

src/profile/charset.cpp
```diff
@@ -17,7 +17,7 @@ void Load() {
     PushInitialIndex();
     while (PushNextTableElement() != 0) {
       int i = EnsureGetKey<int32_t>();
-      Flags[i] = EnsureGetArrayElement<int>();
+      Flags[i] = (uint8_t)EnsureGetArrayElement<int>();
```
KKhanhH: uint8_t should work directly as the ensureget type

PringlesGang: Will it also work properly with the alignment and stuff of the Lua table? Like, will it actually pop off enough?

KKhanhH: the implementation is just calling the lua_getnumber function and then static casted to T

PringlesGang: Cool

src/debugmenu.cpp
```diff
@@ -625,7 +626,7 @@ void ShowScriptDebugger() {
             Io::FileMeta scriptMeta;
             Io::VfsGetMeta("script", it->second.first, &scriptMeta);
             bool isBreakpoint = true;
-            snprintf(buf, 256, "%08lX - %d - %s - %s", it->second.second,
+            snprintf(buf, 256, "%08X - %d - %s - %s", it->second.second,
```
KKhanhH: switch this to fmt format to n

src/debugmenu.cpp
```diff
@@ -792,10 +794,10 @@ void ShowObjects() {
         // Only OpenGL for now
         if (Profile::ActiveRenderer == +RendererType::OpenGL) {
           ImVec2 pos = ImGui::GetCursorScreenPos();
-          ImGui::Image((ImTextureID)spriteSheet.second.Texture,
+          ImGui::Image((ImTextureID)(intptr_t)spriteSheet.second.Texture,
```
KKhanhH: should this cast be intptr_t? Texture isn't really a pointer

src/games/cclcc/librarymenu.cpp
```diff
@@ -35,6 +35,10 @@ LibrarySubmenu& LibraryMenu::GetMenuFromType(
     case LibraryMenuPageType::Movie:
       return static_cast<LibrarySubmenu&>(*UI::MovieMenuPtr);
   }
+
+  ImpLog(LogLevel::Fatal, LogChannel::General,
+         "Unknown library submenu type {}", (int)menuType);
+  return static_cast<LibrarySubmenu&>(*UI::AlbumMenuPtr);
```
KKhanhH: I'm starting to think it would just make more sense to throw an exception for these kinda errors, they should terminate on fnoexceptions, though maybe we should rethink that as well on switch.

src/games/chlcc/introsequence.cpp
```diff
@@ -221,24 +224,25 @@ void IntroSequence::DrawBackground() const {
...
   } else if (StarBounceAnimation.Progress < 0.536f) {
     float progress =
         (StarBounceAnimation.Progress - 0.357f) / (0.536f - 0.357f);
-    y -= std::sin(progress * M_PI) * 0.094f * DesignHeight;
+    y -= std::sin(progress * (float)std::numbers::pi) * 0.094f * DesignHeight;
   } else if (StarBounceAnimation.Progress < 0.714f) {
     float progress =
         (StarBounceAnimation.Progress - 0.536f) / (0.714f - 0.536f);
-    y -= std::sin(progress * M_PI) * 0.094f * DesignHeight;
+    y -= std::sin(progress * (float)std::numbers::pi) * 0.094f * DesignHeight;
```
KKhanhH: std::numbers::pi_v\<float\>

src/vm/interface/input.cpp
```diff
@@ -36,9 +36,11 @@ static void UpdateFromPADCode(int PADcode, PADInputType type) {
   const auto& axisDownLightArr = type == PADInputType::WentDown
                                      ? Input::ControllerAxisWentDownLight
                                      : Input::ControllerAxisIsDownLight;
+  /*
   const auto& axisDownHeavyArr = type == PADInputType::WentDown
```
KKhanhH: use [[maybe_unused]] instead

src/vm/interface/scene3d.cpp
```diff
@@ -19,10 +19,10 @@ static void UpdateScrWorkAnimations(float dt) {
   for (size_t i = 0; i < CurrentScrWorkAnimations.size(); i++) {
     int id = CurrentScrWorkAnimations[i];
     ScrWorkAnimations[id].MainAnimation.Update(dt);
-    for (size_t i = 0; i < ScrWorkAnimations[id].AnimationData.size(); i++) {
-      ScrWork[ScrWorkAnimations[id].AnimationData[i].Target] = glm::mix(
-          ScrWorkAnimations[id].AnimationData[i].From,
-          ScrWorkAnimations[id].AnimationData[i].To,
+    for (size_t j = 0; j < ScrWorkAnimations[id].AnimationData.size(); j++) {
```
KKhanhH: since j isn't being used directly, it's better to just make this for range

src/vm/inst_system.cpp
```diff
@@ -523,15 +523,14 @@ VmInstruction(InstSystemMes) {
...
-  std::tm* now = std::localtime(&t);
-  ScrWork[SW_TIMEYEAR] = now->tm_year + 1900;
-  ScrWork[SW_TIMEMONTH] = now->tm_mon + 1;
-  ScrWork[SW_TIMEDAY] = now->tm_mday;
-  ScrWork[SW_TIMEHOUR] = now->tm_hour;
-  ScrWork[SW_TIMEMINUTE] = now->tm_min;
-  ScrWork[SW_TIMESECOND] = now->tm_sec;
-  ScrWork[SW_TIMEWEEK] = now->tm_wday;
+  const tm dateTime = CurrentDateTime();
+  ScrWork[SW_TIMEYEAR] = dateTime.tm_year;
```
KKhanhH: Tm year is from 1900

src/util.cpp
```diff
@@ -350,4 +350,28 @@ RectF RectF::BoundingBox(const CornersQuad& first, const RectF& second) {
...
+                       now - std::chrono::floor<std::chrono::years>(now))
+                       .count();
+
+  return tm{
+      .tm_sec = (int)time.seconds().count(),
+      .tm_min = (int)time.minutes().count(),
+      .tm_hour = (int)time.hours().count(),
+      .tm_mday = (int)(unsigned)date.day(),
+      .tm_mon = (int)(unsigned)date.month(),
+      .tm_year = (int)date.year(),
```
KKhanhH: year since 1900

src/util.cpp
```diff
@@ -350,4 +350,28 @@ RectF RectF::BoundingBox(const CornersQuad& first, const RectF& second) {
...
+  const int yday = std::chrono::duration_cast<std::chrono::days>(
+                       now - std::chrono::floor<std::chrono::years>(now))
+                       .count();
+
+  return tm{
+      .tm_sec = (int)time.seconds().count(),
+      .tm_min = (int)time.minutes().count(),
+      .tm_hour = (int)time.hours().count(),
+      .tm_mday = (int)(unsigned)date.day(),
+      .tm_mon = (int)(unsigned)date.month(),
```
KKhanhH: month is 0 based

src/vm/inst_system.cpp
```diff
@@ -523,15 +523,14 @@ VmInstruction(InstSystemMes) {
...
+  ScrWork[SW_TIMEMONTH] = dateTime.tm_mon;
```
KKhanhH: +1

CMakeLists.txt
```diff
@@ -1167,6 +1159,8 @@ if (NOT DEFINED IMPACTO_DISABLE_DX9)
     )
 endif ()
 
+set(CMAKE_COMPILE_WARNING_AS_ERROR TRUE)
```
KKhanhH: should set this just for the impacto target

src/renderer/3d/model.cpp
```diff
@@ -62,21 +62,19 @@ void Model::EnumerateModels() {
   uint32_t currentBackgroundModel = 0;
 
   g_ModelIds = (uint32_t*)malloc(g_ModelCount * sizeof(uint32_t));
```
KKhanhH: should kill the malloc as well or add a todo for it

src/renderer/3d/model.cpp
```diff
@@ -62,21 +62,19 @@ void Model::EnumerateModels() {
...
       (uint32_t*)malloc(g_BackgroundModelCount * sizeof(uint32_t));
-  g_BackgroundModelNames =
-      (char**)malloc(g_BackgroundModelCount * sizeof(char*));
+  g_BackgroundModelNames.reserve(g_BackgroundModelCount);
 
   for (auto const& file : listing) {
     if (Profile::Scene3D::ModelsToCharacters.count(file.first) != 0) {
       g_ModelIds[currentModel] = file.first;
-      g_ModelNames[currentModel] = strdup(file.second.c_str());
+      g_ModelNames[currentModel] = file.second;
```
KKhanhH: this change isn't the same, are you sure the lifetime of file outlasts g_ModelNames?

src/texture/ddsloader.cpp
```diff
@@ -183,11 +183,15 @@ bool internal_readimg(Stream* stream, unsigned char* dst, int w, int h, int d) {
         for (int x = 0; x < w; x++) {
           if (!stream->Read(&pixel, m_Bpp)) return false;
           k = (z * h * w + y * w + x) * m_nchans;
-          dst[k + 0] = ((pixel & m_dds.fmt.rmask) >> m_redR) << m_redL;
-          dst[k + 1] = ((pixel & m_dds.fmt.gmask) >> m_greenR) << m_greenL;
-          dst[k + 2] = ((pixel & m_dds.fmt.bmask) >> m_blueR) << m_blueL;
+          dst[k + 0] =
+              (unsigned char)(((pixel & m_dds.fmt.rmask) >> m_redR) << m_redL);
```
KKhanhH: use uint8_t instead

src/text.h
```diff
@@ -123,19 +123,19 @@ struct DialoguePage {
...
+  size_t LastLineStart;
   DialoguePageMode PrevMode = DPM_ADV;
 };
 
 inline DialoguePage* DialoguePages;
 inline int DialoguePageCount = 0;
 
 int TextGetStringLength(Vm::Sc3VmThread* ctx);
-int TextGetMainCharacterCount(Vm::Sc3VmThread* ctx);
+// int TextGetMainCharacterCount(Vm::Sc3VmThread* ctx); TODO: Implement
```
KKhanhH: don't comment out the whole function, use [[maybe_unused]]

src/debugmenu.cpp
PringlesGang: I don't quite understand what you want here? Do you just want me to replace the buffer with a std::string using fmt?

KKhanhH: change the snprintf to fmt's format_to_n call

PringlesGang: Ahh gotcha, I thought you wanted me to replace the format specifiers with %n and got confused

src/debugmenu.cpp
PringlesGang: I know, but `ImTextureID` is a typedeffed `void*`, and casting a `uint32_t` to a `void*` directly throws warnings

src/games/cclcc/librarymenu.cpp
PringlesGang: Yeah, at this point the whole application's in UB anyway, so exceptions are probably better

src/games/chlcc/introsequence.cpp
PringlesGang: omg that's so nice

CMakeLists.txt
PringlesGang: From what I understand it will compile with/without warnings depending on the current state of `COMPILE_WARNING_AS_ERROR` (defaulted to `CMAKE_COMPILE_WARNING_AS_ERROR`) at the time the target is created. Because only the impacto target is defined after this point, it should only target impacto
I could not find any better way to do this, but if you know of any please lmk

src/renderer/3d/model.cpp
PringlesGang: That one's still just a `uint32_t*`, but I'll change it to a vector anyway
Tho I have a sneaking suspicion this whole file's gonna have to be rewritten anyway once we get back to working on R;N again...

src/renderer/3d/model.cpp
PringlesGang: Whoops you're right

CMakeLists.txt
KKhanhH: You should be able to just do set_property(TARGET impacto PROPERTY COMPILE_WARNING_AS_ERROR ON) after add_executable/add_library

src/debugmenu.cpp
KKhanhH: Size_t might be better if it doesn't warn

KKhanhH (approved):

src/util.cpp
```diff
@@ -350,4 +350,28 @@ RectF RectF::BoundingBox(const CornersQuad& first, const RectF& second) {
   return BoundingBox(second, first);
 }
 
+tm CurrentDateTime() {
+  const auto now = std::chrono::system_clock::now();
```
KKhanhH: doing some more testing with this, it looks like this implementation differs in that it provides the UTC time and not system clock time, which is what we had before

KKhanhH (changes requested):

PR #180, opened 2025-06-19
Updated CHLCC SystemMenu and docs [Renamed Branch]

Nergon123:


src/util.h
```diff
@@ -320,15 +320,25 @@ inline void MakeLowerCase(std::string& str) {
...
   static_assert(std::is_pod<T>::value, "!std::is_pod<T>");
+#endif
   T value;
   memcpy(&value, ptr, sizeof value);
   return value;
 }
 
 template <typename T>
 void UnalignedWrite(void* ptr, T value) {
+#if __cplusplus >= 202002L  // Suppress deprecated warning in C++20+
```
KKhanhH: revert this, warnings will be handled in another PR

src/games/chlcc/systemmenu.h
```diff
@@ -20,6 +20,10 @@ class SystemMenu : public Menu {
   void MenuButtonOnClick(Widgets::Button* target);
 
  private:
+  void DrawRunningSelectedLabel(float offsetY);
+  void UpdateRunningSelectedLabel(float dt);
+  float currentRunningPosition = 0.0f;
```
KKhanhH: Class member variables should be PascalCase

src/games/chlcc/systemmenu.cpp
```diff
@@ -112,18 +121,23 @@ void SystemMenu::Update(float dt) {
...
 
   if (State == Shown && IsFocused) {
+    UpdateRunningSelectedLabel(dt);
     MainItems->Update(dt);
   }
 }
+
 void SystemMenu::Render() {
   if (State != Hidden) {
+    int indexOfActiveButton = GetIndexOfActiveButton();
```
KKhanhH: You can use CurrentlyFocusedElement for this to get the pointer to the currently focused button, and then get the Id member of it, so no need for the function

.gitignore
```diff
@@ -287,3 +287,14 @@ vendor/avcpp
 games/*/gamedata
 games/*/savedata
 games/*/trophydata
+
+# Compiled Files
```
KKhanhH: This probably shouldn't need to be in the .gitignore. You can generate the build files in a contained folder that's already gitignored and it won't be tracked. build i believe is already gitignored

Nergon123: oh, okay I was just running "make" command and its generating build in root for some reason 
CMake generating makefile with 

```
# The top-level build directory on which CMake was run.
CMAKE_BINARY_DIR = /home/nergon/repo/impacto

```

PR #182, opened 2025-06-28
Chlcc system message stars

newo-2001:
This PR adds the rotating stars in the CHLCC SysMesBox when it is loading.
In order for this to show properly in Impacto, we have to make these loading tasks asynchronous and move them to a background thread so that they don't block the main thread.


src/games/chlcc/sysmesbox.h
```diff
@@ -19,6 +26,10 @@ class SysMesBox : public UI::SysMesBox {
   virtual void AddChoice(uint8_t* str) override;
 
  private:
+  static constexpr size_t LoadingStarCount = 14;
+  std::array<SysMesBoxStar, LoadingStarCount> LoadingStars;
```
PringlesGang: Just put the max value in the array and use `LoadingStars.size()`

PringlesGang (approved):

src/data/achievementsystem.cpp
```diff
@@ -18,14 +18,21 @@ class AchievementFileLoader
...
+    if (MainThreadCallback) MainThreadCallback();
+
     // Let's not report errors until we finalize the implementation
     result = AchievementError::OK;
 
     ScrWork[SW_SAVEERRORCODE] = (int)result;
   }
+
+ private:
+  std::function<void(void)> MainThreadCallback;
```
KKhanhH, line 38: Maybe instead of this, just override a mainthreadcallback function in the achievement base and call that

KKhanhH: actually it's probably fine like this, since you want queued trophies to only be ephemeral anyway. Reset the std::function though once you call it so that it doesn't use up memory.

newo-2001: The texture data needs to be stored somewhere until the main thread can submit it because the achievements only hold on to the sprite, where would you propose to store that?

PR #188, opened 2025-07-05
Save on worker thread

newo-2001:
This PR changes the way `InstSave` works to more closesly resemble the actual VM, this involves executing some of the work on worker threads so that loading animations can still be rendered during this process.

Some issues that have been addressed in order to perform this work on a background thread are:
- Texture submission have to be delayed and performed on the main thread.
- VM state such as ScrWork and FlagWork have to be copied to a buffer in a blocking manner before flushing it asynchronously, as running other VM threads whilst this is happening could lead to corruption. This part of saving and loading has been moved to `SaveSystemData()` and `LoadSystemData()`. This seems to be how all the games are doing it internally as well.

~~I descoped doing save file writing asynchronously because it interacts with the renderer again which is likely done in the scripts through a separate call to `Save()`~~

I was unable to find any regressions in CCLCC. As for CHLCC and MO6TW, your mileage may vary due to the implementations being incomplete, but most of the issues I found were seemingly already present on master.

This is an enabler for #182 because we can't render loading animations whilst blocking the main thread.


src/renderer/opengl/3d/scene.cpp
```diff
@@ -89,7 +89,7 @@ void Scene3D::Render() {
   SetupFramebufferState();
 
   for (int i = 0; i < Profile::Scene3D::MaxRenderables; i++) {
-    if (Renderables[i]->Status == LS_Loaded &&
```
PringlesGang, line 92: `LoadStatus::Loaded` on lines 81 and 92

src/games/chlcc/savesystem.cpp
```diff
@@ -19,15 +20,87 @@ using namespace Impacto::Profile::ScriptVars;
...
+  stream.Seek(0xbc2, SEEK_SET);
+  Io::ReadArrayLE<uint8_t>(BGMFlags, &stream, 100);
+
+  stream.Seek(0xc26, SEEK_SET);
+  Io::ReadArrayLE<uint8_t>(MessageFlags, &stream, 10000);
+
+  stream.Seek(0x358E, SEEK_SET);
+  Io::ReadArrayLE<uint8_t>(GameExtraData, &stream, 1024);
+
+  stream.Seek(0x3b06, SEEK_SET);  // TODO: Actually load system data
```
PringlesGang, line 90: Shouldn't this still be a part of `WriteSaveFile`?

src/games/mo6tw/savesystem.cpp
```diff
@@ -17,16 +18,69 @@ using namespace Impacto::Profile::ScriptVars;
...
+  stream.Seek(0xbc2, SEEK_SET);
+  Io::ReadArrayLE<uint8_t>(BGMFlags, &stream, 100);
+
+  stream.Seek(0xc26, SEEK_SET);
+  Io::ReadArrayLE<uint8_t>(MessageFlags, &stream, 10000);
+
+  stream.Seek(0x3336, SEEK_SET);
+  Io::ReadArrayLE<uint8_t>(GameExtraData, &stream, 1024);
+
+  stream.Seek(0x3b06, SEEK_SET);  // TODO: Actually load system data
```
PringlesGang, line 53: Same here

src/games/chlcc/savesystem.cpp
newo-2001: `WriteSaveFile()` calls `Io::WriteArrayLE<uint8_t>(SystemData.data(), stream, SystemData.size());` where `SystemData.size() = 0x3b06` so it should already move the write head.

src/games/cclcc/savesystem.cpp
```diff
@@ -79,22 +79,24 @@ SaveError SaveSystem::CheckSaveFile() {
...
-SaveError SaveSystem::CreateSaveFile() {
+void SaveSystem::CreateSaveFile() {
   using CF = Io::PhysicalFileStream::CreateFlagsMode;
   Io::Stream* stream;
   IoError err = Io::PhysicalFileStream::Create(
       SaveFilePath, &stream,
       CF::CREATE_IF_NOT_EXISTS | CF::CREATE_DIRS | CF::WRITE);
   if (err != IoError_OK) {
-    return SaveFailed;
+    ImpLog(LogLevel::Error, LogChannel::IO,
```
KKhanhH, line 97: Why remove the return type? this is used by game scripts to display an error

src/data/savesystem.cpp
```diff
@@ -9,29 +12,92 @@ namespace Impacto {
...
-  if (Implementation)
-    return Implementation->MountSaveFile();
-  else
-    return SaveOK;  // Just so we don't get stuck at loading save data in
-                    // script.
+void MountSaveFile() { ExecuteLoader<SaveFileLoader>(); }
+void CheckSaveFile() { ExecuteLoader<SaveFileChecker>(); }
+
+void WriteSaveFile() {
+  if (Implementation) Implementation->WriteSaveFile();
```
KKhanhH: This should be threaded as well

src/games/cclcc/savesystem.cpp
newo-2001: My best guess is that at some point I thought impacto's CreateSaveFile function matched a different one in the binary, which did not have a return value. I checked again, and it indeed does seem to have it.

src/data/savesystem.cpp
newo-2001: This was a bit tricky because it needs to retrieve thumbnail data, which must happen on the main thread. It looks like the mages engine synchronously flushes the thumbnails to a buffer on the main thread first (when `InstSave(16)` is called), and then simply copies the contents when writing, so I copied this approach.

KKhanhH: Yeah that should probably be fine, thumbnail data isn't that big and main bottleneck is disk i/o

src/games/cclcc/savesystem.cpp
KKhanhH: We don't always have to match what they do if it's an improvement or provides more detail either

PringlesGang: I don't entirely agree with this. The error code is usually written directly into ScrWork, after which the script branches based on the error received. In this case we'd like to follow the binary as closely as possible, so we don't end up in bad state on the script side.
(Although apparently in this case it *should* indeed return something)

KKhanhH: We should at least keep the return codes, even if we don't write into the scripts. More infornation is better to have, even if we only use it internally.

PringlesGang: Yeah, a conversion when writing into ScrWork would be best

profiles/common/scriptvars.lua
```diff
@@ -369,6 +369,7 @@ root.ScriptVars = {
     SF_ACTORSVOICE_UNLOCK1=840,
     SF_EXTRA_ENA=860,
     SF_MOVIE_UNLOCK1=910,
+    SF_SYSMESLOADING=1226,
```
KKhanhH: this is officially named SF_SAVEICON

PringlesGang (approved):

PR #189, opened 2025-07-05
Background2D/Character2D refactor

PringlesGang:
Refactors the `Character2D` and `Background2D` classes to be more extendable and make better use of ScrWork

Also fixes the character sprites briefly blinking on the screen when loaded in


profiles/chlcc/scriptvars.lua
```diff
@@ -78,19 +108,27 @@ sv.SW_BG1PRI2 = 2408; -- Effectively disable
 sv.SW_BG1DISPMODE = 2409;
 sv.SW_BG1FADECT = 2410;
 sv.SW_BG1FADETYPE = 2411;
+sv.SW_BG1ROTZ = 2412;
 sv.SW_BG1ALPHA = 2413;
 sv.SW_BG1MASKNO = 2414;
 sv.SW_BG1MASKFADERANGE = 2415;
+sv.SW_BG1FILTER = 0; -- CHLCC has no bg filter
```
KKhanhH, line 115: if not used then they don't need to be set

KKhanhH: should probably namespace these at some point though

profiles/common/scriptvars.lua
```diff
@@ -237,13 +226,13 @@ root.ScriptVars = {
     -- R;NE/DaSH End
     SW_CHA1POSX=5100,
     SW_CHA1POSY=5101,
-    SW_CHA1POSZ=5102,
+    SW_CHA1ROTZ=5102,
```
KKhanhH: i think it's 2 is X, 3 Y, 4 Z

src/vm/interface/scene2d.cpp
```diff
@@ -140,7 +140,7 @@ void LinkBuffers(int linkCode, int currentBufferId,
         currentBuffer->Links[i].LinkedBuffer = childBuf;
         currentBuffer->Links[i].Direction = (LinkDirection)dir;
         switch (dir) {
```
KKhanhH: maybe just cast dir instead? or use currentBuffer->Links[i].Direction

src/background2d.cpp
```diff
@@ -258,68 +291,114 @@ void Background2D::Render(const int bgId, const int layer) {
...
+            ScrWork[SW_CAP1PRI2 + structOffset]};
+  Show = GetFlag(SF_CAP1DISP + capId);
+  RenderType = ScrWork[SW_CAP1FADETYPE + structOffset];
+
+  MaskNumber = ScrWork[SW_CAP1MASKNO + structOffset];
+  FadeRange = ScrWork[SW_CAP1MASKFADERANGE + structOffset];
+  FadeCount = ScrWork[SW_CAP1FADECT + structOffset];
+
+  const int dispMode = ScrWork[SW_CAP1DISPMODE + structOffset];
+  switch (dispMode) {
```
KKhanhH, line 312: Add a comment or enum for each mode

src/character2d.cpp
```diff
@@ -325,16 +327,21 @@ void Character2D::UpdateState(const int chaId) {
             ScrWork[SW_CHA1ROTZ_OFS + structOfsOffset]);
   }
 
-  Face = ScrWork[SW_CHA1FACE + structOffset] << 16;
+  if (Profile::Vm::GameInstructionSet == +Vm::InstructionSet::CHLCC) {
```
KKhanhH: should MO6 be here as well?

KKhanhH (approved):
Just a few minor changes, if no regressions and looks fine then you can merge

profiles/chlcc/scriptvars.lua
PringlesGang: It's set to 0 so the game doesn't read garbage data and set that as the tint (and to be explicit about its absence)

profiles/common/scriptvars.lua
PringlesGang: Bruh how did I fuck that up

src/background2d.cpp
PringlesGang: Frankly I don't know what they mean :/

src/character2d.cpp
PringlesGang: Dunno, but I should probably make _not_ tinting the default anyway

PR #195, opened 2025-07-13
BgEff rendering and Character masking

PringlesGang:
Renders the bgeff masks as well as color-masking the character sprites depending on the background in C;C LCC.


src/background2d.cpp
```diff
@@ -74,7 +74,51 @@ bool Background2D::LoadSync(uint32_t bgId) {
     if (err != IoError_OK) return false;
     BgTexture.Load(stream);
     delete stream;
+
+    if (Profile::UseBgEffects) {
+      Io::FileMeta meta;
+      err = Io::VfsGetMeta("bg", bgId, &meta);
+      if (err != IoError_OK) return false;
+
+      const std::string baseName =
```
KKhanhH: could use a string_view instead here, if you use the string view constructor with two iterators

src/background2d.h
```diff
@@ -139,4 +139,6 @@ inline Background2D ShaderScreencapture;
 
 inline ankerl::unordered_dense::map<int, Background2D*> Backgrounds2D;
 
+inline Background2D* LastRenderedBackground = nullptr;
```
KKhanhH: maybe these kind of variables would make more sense as static class members, they're just kind of freely unscoped right now

src/background2d.h
```diff
@@ -141,4 +142,14 @@ inline ankerl::unordered_dense::map<int, Background2D*> Backgrounds2D;
 
 inline Background2D* LastRenderedBackground = nullptr;
 
+inline std::map<int, std::array<ShaderProgramType, 4>> BgEffShaders;
```
KKhanhH: unless you need iterator stability, use ankerl maps, and if you don't need ordered sorting then use unordered

src/profile/data/bgeff.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+#include <map>
+#include "../profile_internal.h"
+#include "../../background2d.h"
+
+namespace Impacto {
+namespace Profile {
+namespace BgEff {
+
+const static std::map<uint8_t, ShaderProgramType> ShaderIdToShaderType{
```
KKhanhH: This could probably just be a better enum and better enum methods for safe conversion

KKhanhH: If it's coming from the lua you can also use DefineEnumInt and put the enum directly in the lua

PringlesGang: Making this a better enum seems to be difficult, as maps need the values to be default constructable, which better enums aren't (can enable that though, through some whack construction)

KKhanhH: I was wondering if its a 1:1 mapping then do you need the map? I meant you could replace it entirely with better enum

src/background2d.cpp
PringlesGang: I took all this out before I even opened the PR

KKhanhH: Im doing comments by commit so you might see old stuff, itll say outdated though

src/profile/data/bgeff.cpp
KKhanhH: Also map values shouldn't need to be default constructible if you use emplace functions

PringlesGang: It's also whining at array construction 💀 
As in, `BgEffShaderMap[bgId]` returns a `std::array<ShaderProgramType, 4>`

KKhanhH: Use .at or *(.find)

KKhanhH: I shouldve said prefer find in this case so you can also avoid the contains check (which is a double lookup)

PringlesGang: Yeah I know, I just figured the performance hit would be negligible and I think `.find` is uglier, but whatevs

KKhanhH: Unfortunately this is c++, so a proper optional won't exist for another 3 years

PR #198, opened 2025-07-18
Fix missing save thumbnails

newo-2001:
Fixes a regression from #188 that causes save thumbnails to become corrupted after leaving the game without saving.

The issue was that during the loading process the image data was not written to the intermediate buffer in memory.

### DIFF
--- a/src/games/cclcc/savesystem.cpp
+++ b/src/games/cclcc/savesystem.cpp
@@ -217,16 +217,14 @@ void SaveSystem::LoadEntryBuffer(Io::MemoryStream& stream, SaveFileEntry& entry,
 
   tex.Init(TexFmt_RGB, SaveThumbnailWidth, SaveThumbnailHeight);
 
-  // CCLCC PS4 Save thumbnails are 240x135 RGB16
-  std::vector<uint16_t> thumbnailData(SaveThumbnailWidth * SaveThumbnailHeight);
-
   int thumbnailPadding = 0xA14;
   stream.Seek(thumbnailPadding, SEEK_CUR);
-  Io::ReadArrayLE<uint16_t>(thumbnailData.data(), &stream,
-                            thumbnailData.size());
+  Io::ReadArrayLE<uint8_t>(entry.ThumbnailData.data(), &stream,
+                           entry.ThumbnailData.size());
 
-  for (int i = 0; i < thumbnailData.size(); i++) {
-    uint16_t pixel = thumbnailData[i];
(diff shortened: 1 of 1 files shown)


src/games/cclcc/savesystem.cpp
```diff
@@ -217,16 +217,14 @@ void SaveSystem::LoadEntryBuffer(Io::MemoryStream& stream, SaveFileEntry& entry,
...
-  Io::ReadArrayLE<uint16_t>(thumbnailData.data(), &stream,
-                            thumbnailData.size());
+  Io::ReadArrayLE<uint8_t>(entry.ThumbnailData.data(), &stream,
+                           entry.ThumbnailData.size());
 
-  for (int i = 0; i < thumbnailData.size(); i++) {
-    uint16_t pixel = thumbnailData[i];
+  for (int i = 0; i < entry.ThumbnailData.size() / 2; i++) {
+    uint16_t pixel =
+        entry.ThumbnailData[i * 2] | (entry.ThumbnailData[i * 2 + 1] << 8);
```
KKhanhH, line 227: This doesnt account for host endianness

KKhanhH: Tho not sure if we account for that in the rgb calculations either

newo-2001: Is the endianness of the save-file defined in the file format or does it match the host endianness? And if the endianness of the host matters, what does the `LE` in `ReadLE` do? I had a look at the implementation but the constexpr stuff goes a bit over my head. I'd love to test on a big-endian machine but I also don't see an easy way to do that besides spinning up a full VM.

KKhanhH: Savefile format should be consistent with PS4, so it's LE. The LE and BE functions basically just read it as LE or BE (so the source) , and swaps only if the host endianess doesn't match.

newo-2001: But if the bytes on disk are always LE, then reading byte-by-byte on a big-endian system, then manually constructing the 16-bit number using bitshifts should just work, because I believe bitshifting is independent of endianness.

KKhanhH: Oh yeah, sorry you're right. Should be good then

KKhanhH (approved):

src/games/cclcc/savesystem.cpp
newo-2001: No problem, better be safe then sorry, you at least had me double check.

PR #215, opened 2025-07-28
Frame bgeffs

PringlesGang:
Implements the background effects that cover the entire frame, as seen in C;C (not LCC)

### DIFF
--- a/src/background2d.cpp
+++ b/src/background2d.cpp
@@ -431,7 +431,7 @@ void Background2D::Render(const int layer) {
         nonSpriteShader |= ChaBgEff.Shader != +ShaderProgramType::Sprite;
       }
 
-      renderBgEffs &= nonSpriteShader;
+      if (!nonSpriteShader) renderBgEffs = false;
     }
 
     LastRenderedBackground = renderBgEffs ? this : nullptr;


src/background2d.cpp
```diff
@@ -77,22 +77,42 @@ bool Background2D::LoadSync(uint32_t bgId) {
     BgTexture.Load(stream);
     delete stream;
 
-    BgEffsLoaded = false;
-    if (Profile::UseBgEffects && BgEffTextureIdMap.contains(bgId)) {
+    if (Profile::UseBgFrameEffects && BgEffTextureIdMap.contains(bgId)) {
+      std::for_each(FrameBgEffs.begin(), FrameBgEffs.end(),
+                    [](auto bgEff) { bgEff.Loaded = false; });
```
KKhanhH: auto&, this has no effect as is

src/background2d.cpp
```diff
@@ -77,22 +77,42 @@ bool Background2D::LoadSync(uint32_t bgId) {
...
       const std::array<int, 4>& textureIds = BgEffTextureIdMap[bgId];
 
-      for (size_t i = 0; i < MaxBgEffCount; i++) {
+      for (size_t i = 0; i < FrameBgEffs.size(); i++) {
+        if (textureIds[i] == 0) continue;
+
         Io::Stream* bgEffStream;
         err = Io::VfsOpen("bgeffect", textureIds[i], &bgEffStream);
 
         if (err != IoError::IoError_OK) return false;
```
KKhanhH: should an error in one bgeffect load prevent the rest from loading actually?

src/background2d.cpp
```diff
@@ -344,11 +402,89 @@ void Background2D::UpdateState(const int bgId) {
 void Background2D::Render(const int layer) {
   if (Status != LoadStatus::Loaded || !OnLayer(layer) || !Show) return;
 
-  LastRenderedBackground = this;
+  if (Profile::UseBgChaEffects || Profile::UseBgFrameEffects) {
+    bool renderBgEffs = false;
+
+    for (size_t i = 0; i < BgEffsLayers.size(); i++) {
+      renderBgEffs |= BgEffsLayers[i] != 0 && BgEffsLayers[i] > Layers[i];
```
KKhanhH, line 417: I think this can be replaced with an any_of

src/background2d.cpp
```diff
@@ -344,11 +402,89 @@ void Background2D::UpdateState(const int bgId) {
...
+            FrameBgEffs.begin(), FrameBgEffs.end() - 1, [](const auto bgEff) {
+              return bgEff.Shader != +ShaderProgramType::Sprite;
+            });
+      }
+
+      if (Profile::UseBgChaEffects) {
+        nonSpriteShader |= ChaBgEff.Shader != +ShaderProgramType::Sprite;
+      }
+
+      renderBgEffs &= nonSpriteShader;
```
KKhanhH: just use an if, it's more clear

src/renderer/opengl/renderer.cpp
```diff
@@ -832,14 +832,22 @@ void Renderer::CaptureScreencap(Sprite& sprite) {
   sprite.Bounds.Width = sprite.Sheet.DesignWidth;
   sprite.Bounds.Height = sprite.Sheet.DesignHeight;
 
-  Window->SwapRTs();
+  int prevReadBuffer;
+  int drawBuffer;
   int prevTextureBinding;
+
+  glGetIntegerv(GL_DRAW_FRAMEBUFFER_BINDING, &prevReadBuffer);
```
KKhanhH: should this be GL_READ_FRAMEBUFFER_BINDING

KKhanhH (changes requested):

src/background2d.cpp
PringlesGang: Currently we don't even do anything with the load result, so it would still act as if everything was loaded fine. Regardless, as all these bgeffects are defined explicitly as being used by the binary, I think we should report an error when even one of them fails

src/background2d.cpp
PringlesGang: I need the `i` for both `BgEffsLayers[i]` and `Layers[i]` :/

src/background2d.cpp
PringlesGang: You mean like `if (!nonSpriteShader) renderBgEffs = false;`?

src/background2d.cpp
KKhanhH: Hm, i guess you could technically get the index if you get a difference of the addresses, though thats not exactly safe if data type changes...

src/background2d.cpp
KKhanhH: Yeah

src/background2d.cpp
KKhanhH: Not entirely nothing, because it's an early return and .Loaded wouldn't get set on the members of every effect, so it won't do the texture submission. You can move the return false to the end of the load for error reporting

PringlesGang: Fair enough

KKhanhH (approved):

PR #217, opened 2025-08-04
Chlcc/save system

PringlesGang:
Implements a large part of C;H LCC's save system:
- Saves/loads config settings
- Saves/loads thumbnails
- Creates the save file when not found
- Set delusion trigger state upon loading into/out of delusion triggers
- Have the save menu update its entry immediately upon saving
- Set default system data upon first boot
- Fix save menu highlight alignment upon re-entering
- Set and check save file checksums

Also fixes a few general saving/loading issues:
- Have save _creation_ become a part of `WriteSaveFile`, have a separate method for initializing system data, to more closely resemble what is done in the binary
- Split `InstSave` into `InstSave` and `InstSaveOld`, and guard some things in `InstAutoSave` as well
- Improves `SW_PLAYTIME` and `SW_TOTALPLAYTIME` use
- Show corrupted save slots as such


src/games/cclcc/optionsmenu.cpp
```diff
@@ -393,17 +393,17 @@ void OptionsMenu::ResetToDefault() {
       break;
     }
     case PageType::Sound: {
-      std::copy(std::begin(Default::GroupVolumes),
-                std::end(Default::GroupVolumes), Audio::GroupVolumes);
+      std::copy(Default::GroupVolumes.begin(), Default::GroupVolumes.end(),
```
KKhanhH: Btw since we're on 20 now, std::ranges::copy(src, destItr) is also a thing. Don't have to change it, just an fyi

src/games/cclcc/savemenu.cpp
```diff
@@ -361,6 +363,11 @@ void SaveMenu::Render() {
   }
 }
 
+void SaveMenu::RefreshCurrentEntryInfo() {
+  if (!CurrentlyFocusedElement) return;
+  dynamic_cast<SaveEntryButton*>(CurrentlyFocusedElement)->RefreshInfo();
```
KKhanhH: There's not really much point in using dynamic_cast if it's just immediately dereferenced, since the point is being able to check dynamic types by comparing to nullptr, I would just static_cast here if you already know what it should be.

src/games/cclcc/savesystem.cpp
```diff
@@ -80,64 +83,69 @@ SaveError SaveSystem::CheckSaveFile() {
...
-  IoError err = Io::PhysicalFileStream::Create(
-      SaveFilePath, &stream,
-      CF::CREATE_IF_NOT_EXISTS | CF::CREATE_DIRS | CF::WRITE);
-  if (err != IoError_OK) {
-    ImpLog(LogLevel::Error, LogChannel::IO,
-           "Failed to open save file for writing\n");
-    return SaveError::Failed;
-  }
+void SaveSystem::InitializeSystemData() {
+  std::fill(SystemData.begin(), SystemData.end(), 0x00);
```
KKhanhH, line 90: Can we test if it would be faster to remove prefill with 0 and replace seeks with write 0s?

src/games/cclcc/savesystem.cpp
```diff
@@ -581,14 +601,19 @@ void SaveSystem::SaveSystemData() {
...
+      SaveFilePath, &stream,
+      CF::CREATE_IF_NOT_EXISTS | CF::CREATE_DIRS | CF::WRITE | CF::READ);
   if (err != IoError_OK) {
     ImpLog(LogLevel::Error, LogChannel::IO,
            "Failed to open save file for writing\n");
     return SaveError::Failed;
   }
 
+  std::vector<uint8_t> emptyData(SaveFileSize, 0x00);
+  Io::WriteArrayLE<uint8_t>(emptyData.data(), stream, emptyData.size());
```
KKhanhH: similar thing as my systemdata comment, but it might be easier to see results here since entire file is 10megs

src/games/cclcc/savemenu.cpp
```diff
@@ -116,6 +116,8 @@ void SaveMenu::Show() {
 
           EntryGrid[p][i][j] = saveEntryButton;
           MainItems[p]->Add(saveEntryButton);
+
+          saveEntryButton->RefreshInfo();
```
KKhanhH, line 120: Btw I did add an init function to menu.h that's meant to be used for script calls that initialize a menu's data before a show call. It maybe useful to move button creation kinda stuff to there.

src/games/chlcc/savemenu.cpp
```diff
@@ -527,6 +465,12 @@ void SaveMenu::UpdateTitles() {
                 460.0f * (MenuTransition.Progress * 4.0f - 3.0f) / 3.0f);
 }
 
+void SaveMenu::RefreshCurrentEntryInfo() {
+  if (!CurrentlyFocusedElement) return;
+  dynamic_cast<SaveEntryButton*>(CurrentlyFocusedElement)
```
KKhanhH: dynamic cast

src/games/chlcc/savesystem.cpp
```diff
@@ -17,8 +19,145 @@ namespace CHLCC {
...
+using namespace Impacto::Profile::ConfigSystem;
 
-SaveFileEntry* WorkingSaveEntry = 0;
+SaveFileEntry* WorkingSaveEntry = nullptr;
+
+constexpr std::array<float, 4> TextSpeeds = {0x100 / 60.0f, 0x300 / 60.0f,
+                                             0x600 / 60.0f, 0xfff0000 / 60.0f};
+constexpr uint8_t TextSpeedToSettingIndex(const float speed) {
+  for (uint8_t i = 0; i < std::ssize(TextSpeeds) - 1; i++) {
+    const float average = (TextSpeeds[i] + TextSpeeds[i + 1]) / 2.0f;
```
KKhanhH: std::midpoint if you'd like

src/games/cclcc/savesystem.cpp
```diff
@@ -80,64 +83,69 @@ SaveError SaveSystem::CheckSaveFile() {
            "Failed to get save file permissions, error: \"{:s}\"\n",
            ec.message());
     return SaveError::Failed;
-  } else if ((!checkPermsBit(perms, Io::FilePermissionsFlags::owner_read) ||
-              !checkPermsBit(perms, Io::FilePermissionsFlags::owner_write))) {
+  } else if (!checkPermsBit(perms, Io::FilePermissionsFlags::owner_read) ||
```
KKhanhH: I wrote this before cpp20, but this can probably be replaced with an using enum Io::FilePermissionsFlags and (perms & owner_read) != none ||  (perms & owner_write) != none and killing the lambda

src/games/chlcc/savesystem.cpp
```diff
@@ -17,8 +19,145 @@ namespace CHLCC {
...
+  for (uint8_t i = 0; i < std::ssize(AutoSpeeds) - 1; i++) {
+    const float average = (AutoSpeeds[i] + AutoSpeeds[i + 1]) / 2.0f;
+    if (speed <= average) return i;
+  }
+
+  return std::ssize(AutoSpeeds) - 1;
+}
+
+void SaveSystem::InitializeSystemData() {
+  std::fill(SystemData.begin(), SystemData.end(), 0x00);
```
KKhanhH: same as the other comment

src/games/chlcc/savesystem.cpp
```diff
@@ -17,8 +19,145 @@ namespace CHLCC {
...
+  Io::WriteLE<Uint8>(&stream, 0x00);  // TODO: Pad type
+  Io::WriteLE(&stream, Default::TriggerStopSkip);
+
+  stream.Seek(0x79e, SEEK_SET);
+  assert(Default::VoiceMuted.size() >= 32);
+  for (size_t i = 0; i < 32; i++) {
+    Io::WriteLE(&stream, !Default::VoiceMuted[i]);
+  }
+
+  assert(Default::VoiceVolume.size() >= 20);
```
KKhanhH: could use static_assert unless we change it to vector

src/games/chlcc/savesystem.cpp
```diff
@@ -55,6 +229,114 @@ void SaveSystem::SaveSystemData() {
   stream.Seek(0x3b06, SEEK_SET);  // TODO: Actually write system data
```
KKhanhH: is this todo still needed

src/games/chlcc/savesystem.cpp
```diff
@@ -194,82 +518,106 @@ void SaveSystem::FlushWorkingSaveEntry(SaveType type, int id,
...
-      }
-      time_t rawtime;
-      time(&rawtime);
-      entry->SaveDate = CurrentDateTime();
+  if (WorkingSaveEntry != nullptr && entry != nullptr &&
+      !(GetSaveFlags(type, id) & WriteProtect)) {
+    *entry = *WorkingSaveEntry;
+    if (type == SaveType::Quick) entry->SaveType = autoSaveType;
+
+    time_t rawtime;
```
KKhanhH: time_t stuff isn't needed anymore is it?

src/games/chlcc/savesystem.cpp
```diff
@@ -194,82 +518,106 @@ void SaveSystem::FlushWorkingSaveEntry(SaveType type, int id,
...
+  IoError err = Io::PhysicalFileStream::Create(
+      SaveFilePath, &stream,
+      CF::CREATE_IF_NOT_EXISTS | CF::CREATE_DIRS | CF::WRITE | CF::READ);
   if (err != IoError_OK) {
     ImpLog(LogLevel::Error, LogChannel::IO,
            "Failed to open save file for writing\n");
     return SaveError::Failed;
   }
 
+  std::vector<uint8_t> emptyData(SaveFileSize, 0x00);
```
KKhanhH: same thing as the other comment

KKhanhH: definitely don't keep the vector though, it would be better to just use a constexpr static array or something, so we don't do a few megs of heap allocation

src/games/chlcc/savesystem.cpp
```diff
@@ -466,9 +826,13 @@ void SaveSystem::LoadEntry(SaveType type, int id) {
...
         }
 
-        memcpy(thd->Variables, entry->MainThreadVariables, 64);
+        memcpy(thd->Variables, entry->MainThreadVariables.data(),
+               16 * sizeof(int));
         thd->DialoguePageId = entry->MainThreadDialoguePageId;
       }
+
+      // Reset delusion trigger state
+      Impacto::Profile::DelusionTrigger::CreateInstance();
```
KKhanhH: hmm, i feel like this is supposed to be handled by inst_msinit or something.

profiles/chlcc/savedata.lua
```diff
@@ -1,6 +1,7 @@
 root.SaveData = {
     Type = SaveDataType.CHLCC,
     SaveFilePath = "games/chlcc/savedata/SYSTEM.DAT",
+    ThumbnailFilePath = "games/chlcc/savedata/THUMNAIL.DAT",
```
KKhanhH, line 4: is THUMNAIL supposed to be THUMBNAIL? or is this just a mages moment

src/profile/data/savesystem.h
```diff
@@ -16,6 +16,7 @@ inline Impacto::SaveSystem::SaveDataType Type =
     Impacto::SaveSystem::SaveDataType::None;
 
 inline std::string SaveFilePath;
+inline std::optional<std::string> ThumbnailFilePath;
```
KKhanhH, line 19: shouldn't this just be a std::string

src/vm/inst_gamespecific.cpp
```diff
@@ -977,6 +977,11 @@ VmInstruction(InstDelusionTriggerCHLCC) {
   StartInstruction;
   PopUint8(type);
   switch (type) {
```
KKhanhH, line 979: chlcc delusion triggers really do need a rework...

src/ui/widgets/chlcc/saveentrybutton.cpp
```diff
@@ -154,6 +160,78 @@ void SaveEntryButton::UpdateFocusedAlphaFade(float dt) {
...
+    const uint32_t seconds = (time % 3600) % 60;
+    AddPlayTimeText(fmt::format("{:3}:{:02}:{:02}", hours, minutes, seconds),
+                    18, RendererOutlineMode::BottomRight,
+                    {PlayTimeTextRelativePos.x + (float)((hours < 10) * 10),
+                     PlayTimeTextRelativePos.y});
+
+    AddSaveDateHintText(Vm::ScriptGetTextTableStrAddress(0, 3), 18,
+                        RendererOutlineMode::BottomRight,
+                        SaveDateHintTextRelativePos);
+    std::stringstream dateStr;
```
KKhanhH: fmt should be able to format tm

src/vm/inst_system.cpp
```diff
@@ -215,89 +215,63 @@ VmInstruction(InstCopyThreadWork) {
...
-                   // 67, 70, 71, 74, 76
-    case 0:
+  switch (type) {  // TODO: Types 1, 2, 3, 6, 7, 8, 9, 10, 11, 12, 19, 40,
+                   // 41, 45, 46, 47, 50, 51, 52, 53, 88, 89, 150, 151
+    case 0: {
       SaveSystem::SaveSystemData();
       break;
+    }
     case 1:
       break;
```
KKhanhH: I think you can save on space if you combine all the empty cases together and only break on the last one. I'm pretty sure it doesn't warn or require [[fallthrough]] on empty cases

KKhanhH (changes requested):

profiles/chlcc/savedata.lua
PringlesGang: This is a certified MAGES moment

src/profile/data/savesystem.h
PringlesGang: Not all games have a dedicated thumbnail file, so I wanted to be explicit. Don't really like sentinel values like "" for things that _truly are optional_

src/vm/inst_gamespecific.cpp
PringlesGang: You bet. Prolly a good time to do when the delusion trigger text gets done

src/games/chlcc/savesystem.cpp
PringlesGang: Don't think that is actually called upon loading a save. In the binary the internal values of the delusion trigger get reset, but I figured with impacto a more RAII-like approach would be better

src/games/chlcc/savesystem.cpp
PringlesGang: Yeah maybe putting it in static memory is a good idea

src/games/cclcc/savesystem.cpp
PringlesGang: Something like you recently did with the switch fix?

src/games/cclcc/savemenu.cpp
PringlesGang: I think at that point I'd rather just have it RAII and do this initialization in the constructor, creating a whole new instance each time it's needed

KKhanhH: The point of the init function is because menu needs to be constructed at the start of the program, and we can't put things that require script text inside that. Init is supposed to be synchronized with game script calls anyway

src/games/chlcc/savesystem.cpp
KKhanhH: Set a breakpoint in msinit and check if it gets called after a load. I'm pretty sure scripts call it, cuz i remember having to set things in there

KKhanhH: Although i guess if they set it in save calls in the binary then they don't, letting it's a cclcc thing

PringlesGang: Hm, apparently chlcc does it in InstMSinit as well

src/games/cclcc/savemenu.cpp
PringlesGang: Ahh, I see what you mean

src/games/cclcc/savesystem.cpp
PringlesGang: Well, yeah, it is faster. On release mode for the thumbnails it takes 4 ms without prefill and 8 ms with in the worst case, but like... that's still on the order of milliseconds, for something that doesn't happen very often, is always initiated by the user, and is something the user expects could take a little time.
Also I don't really like having to write something like `Io::WriteArrayLE<uint8_t>(emptyData.data(), stream, targetPos - stream->Position)` instead of just `stream->Seek(targetPos, SEEK_SET)`. I don't think adding a new function for this specific use-case is a great idea either.
Furthermore, this would mean we'd have to pay active attention as to whether this data still needs to be zeroed out or if it could already contain some written data (not that we backtrack often while saving, but still something to consider).

src/ui/widgets/chlcc/saveentrybutton.cpp
```diff
@@ -193,42 +187,63 @@ void SaveEntryButton::RefreshInfo(const SaveSystem::SaveType entryType) {
...
+      const uint32_t seconds = (time % 3600) % 60;
+      AddPlayTimeText(fmt::format("{:3}:{:02}:{:02}", hours, minutes, seconds),
+                      18, RendererOutlineMode::BottomRight,
+                      {PlayTimeTextRelativePos.x + (float)((hours < 10) * 10),
+                       PlayTimeTextRelativePos.y});
+
+      AddSaveDateHintText(Vm::ScriptGetTextTableStrAddress(0, 3), 18,
+                          RendererOutlineMode::BottomRight,
+                          SaveDateHintTextRelativePos);
+      std::stringstream dateStr;
```
KKhanhH: unused dateStr?

src/vm/inst_misc.cpp
```diff
@@ -415,6 +415,7 @@ VmInstruction(InstSaveMenuOld) {
       UI::SaveMenuPtr->ActiveMenuType =
           UI::SaveMenuPageType::_from_integral_nothrow(arg1);
       ScrWork[SW_SAVEFILESTATUS] = 0;
+      if (UI::SaveMenuPtr) UI::SaveMenuPtr->Init();
       ImpLogSlow(LogLevel::Warning, LogChannel::VMStub,
```
KKhanhH: remove the stub log

src/vm/inst_misc.cpp
```diff
@@ -442,6 +443,7 @@ VmInstruction(InstSaveMenuOld) {
     } break;
     case 10: {
       PopUint8(arg1);
+      if (UI::SaveMenuPtr) UI::SaveMenuPtr->Init();
       ImpLogSlow(LogLevel::Warning, LogChannel::VMStub,
```
KKhanhH: remove the stub log

src/games/cclcc/savesystem.cpp
KKhanhH: We could add a write function to stream that writes N things, and instead of calling seek, we call something like writeN<uint8_t>(stream, 0, count). That would effectively give it the same thing and I don't think it would be unreasonable for the stream api.

PringlesGang: Yeah that sounds like a useful function, but I still think we should only use that for inserting intended padding; not in place of skipping to a different part of the stream:
> Furthermore, this would mean we'd have to pay active attention as to whether this data still needs to be zeroed out or if it could already contain some written data (not that we backtrack often while saving, but still something to consider).

KKhanhH: I would consider these explicit padding

KKhanhH: If you have an usecase where you actually need to seek back, sure use the prefill. But we shouldn't be leaving behind performance for a hypothetical.

PringlesGang: Idk I think there _is_ a difference. For a large part of the system data we don't know yet what each byte is for, and what they should contain. For the thumbnails I _know_ it's RGB24 followed by `width * height` padding, so I explicitly fill with 0. For system data we're still unsure about a lot of it, so I'd rather prefill and only explicitly write thereafter for stuff we _know_ should hold specific data (or even be empty)
I can also see this causing bugs in development, because in a later section of the function you'd be overwriting data with 0 that should've held new information you wrote earlier in the function

As for the jumping around the save file, for cclcc that doesn't happen so much, but for chlcc I'm jumping back a couple times to the start to set the global checksums

And again, I still don't like having to `Io::WriteN(stream, 0, targetPos - stream->Position)` instead of `stream->Seek(targetPos, SEEK_SET)`. Writing 0s should only be done as padding—when you know the length of the block; not when you know your next target position

I just don't think such a small performance increase is worth the technical debt. You'd really only passively notice this delay if you were writing to tape storage

I'd be willing to split the prefilling up such that the system data en save entries are done separately, so that a large part of the entries (which make up most of the save file) could be padded instead of prefilled, but I really think the system data is better off being prefilled until we've covered _all_ of the system data and know for _certain_ what should be 0 and what shouldn't

KKhanhH: Yeah, fair enough. I'm fine with just padding the entries, since thats where the major time sink is (and we should be mostly certain of what goes into entry data now).

KKhanhH (approved):

PR #218, opened 2025-08-05
More warning fixes

PringlesGang:
Fixes more clang warnings
Mostly int vs size_t
Also added dt to all `UpdateInput` methods


src/background2d.cpp
```diff
@@ -34,8 +34,8 @@ void Background2D::InitFrameBuffers() {
 }
 
 void Background2D::Init() {
-  for (int i = 0; i < Backgrounds.size(); i++) {
-    Backgrounds2D[i] = &Backgrounds[i];
+  for (size_t i = 0; i < Backgrounds.size(); i++) {
+    Backgrounds2D[static_cast<int>(i)] = &Backgrounds[i];
```
KKhanhH: should this be indexed as int

src/data/savesystem.cpp
```diff
@@ -152,14 +152,7 @@ uint8_t GetSaveFlags(SaveType type, int id) {
...
 tm const& GetSaveDate(SaveType type, int id) {
-  static tm t{
-      .tm_sec = 0,
-      .tm_min = 0,
-      .tm_hour = 0,
-      .tm_mday = 1,
-      .tm_mon = 0,
-      .tm_year = 0,
-  };
+  static tm t{};
```
KKhanhH: Mday is out of range here

src/games/cclcc/musicmenu.h
```diff
@@ -35,7 +35,7 @@ class MusicTrackButton : public Widgets::Button {
   std::array<ProcessedTextGlyph, 6> LockedText;
   std::array<ProcessedTextGlyph, 4> NumberText;
   Widgets::Label ArtistName;
-  int Position = 0;
+  [[maybe_unused]] int Position = 0;
```
KKhanhH: If its not used anywhere you can remove it

src/games/cclcc/savesystem.cpp
```diff
@@ -652,14 +652,7 @@ uint8_t SaveSystem::GetSaveFlags(SaveType type, int id) {
...
 tm const& SaveSystem::GetSaveDate(SaveType type, int id) {
-  static const tm t{
-      .tm_sec = 0,
-      .tm_min = 0,
-      .tm_hour = 0,
-      .tm_mday = 1,
-      .tm_mon = 0,
-      .tm_year = 0,
-  };
+  static const tm t{};
```
KKhanhH: Mday is out of bounds

src/games/chlcc/savesystem.cpp
```diff
@@ -300,14 +300,7 @@ uint8_t SaveSystem::GetSaveFlags(SaveType type, int id) {
...
 tm const& SaveSystem::GetSaveDate(SaveType type, int id) {
-  static const tm t{
-      .tm_sec = 0,
-      .tm_min = 0,
-      .tm_hour = 0,
-      .tm_mday = 1,
-      .tm_mon = 0,
-      .tm_year = 0,
-  };
+  static const tm t{};
```
KKhanhH: Mday

src/games/mo6tw/savesystem.cpp
```diff
@@ -586,14 +586,7 @@ uint8_t SaveSystem::GetSaveFlags(SaveType type, int id) {
...
 tm const& SaveSystem::GetSaveDate(SaveType type, int id) {
-  static const tm t{
-      .tm_sec = 0,
-      .tm_min = 0,
-      .tm_hour = 0,
-      .tm_mday = 1,
-      .tm_mon = 0,
-      .tm_year = 0,
-  };
+  static const tm t{};
```
KKhanhH: Mday

src/games/mo6tw/titlemenu.cpp
```diff
@@ -344,7 +344,7 @@ void TitleMenu::Update(float dt) {
         if (MemoriesItems->IsShown && !MainItems->IsShown) {
           MainItems->IsShown = true;
         }
-      }
+      } break;
```
KKhanhH, line 347: Check that this wasnt an intentional fallthrough

src/profile/games/cclcc/savemenu.cpp
```diff
@@ -39,10 +39,10 @@ void Configure() {
   SaveEntryPrimaryColor = EnsureGetMember<uint32_t>("SaveEntryPrimaryColor");
   SaveEntrySecondaryColor =
       EnsureGetMember<uint32_t>("SaveEntrySecondaryColor");
-  for (int i = 0; i < sizeof(SaveMenuTypeNames) / sizeof(*SaveMenuTypeNames);
+  for (size_t i = 0; i < sizeof(SaveMenuTypeNames) / sizeof(*SaveMenuTypeNames);
```
KKhanhH: Change sizeof stuff to std::size too

KKhanhH (changes requested):
Only half done, dont rebase

src/game.cpp
```diff
@@ -298,23 +298,23 @@ static void RenderMain() {
       Renderer->SetFramebuffer(renderTarget);
     }
 
-    for (int bgId = 0; bgId < Backgrounds.size(); bgId++) {
+    for (size_t bgId = 0; bgId < Backgrounds.size(); bgId++) {
```
KKhanhH: maybe we should just use std::ssize rather than .size, we'd need a lot less static_casting that way

KKhanhH (changes requested):
it doesn't seem like apple clang warnings are on, that or github is wrong

src/background2d.cpp
PringlesGang: Yes. That is a map with ints for keys

src/game.cpp
PringlesGang: The real issue is that a lot of these variables are unnecessarily signed, but yeah std::ssize is probably a good idea for now

src/games/mo6tw/titlemenu.cpp
PringlesGang: Cases 10 and 11 have to do with separte submenu entries on the title menu: 10 for Memories and 11 for System; 10 is not supposed to fallthrough to 11

src/data/savesystem.cpp
KKhanhH: this approach kinda sets tm_mday on every call of get date, I think you should place it in a lambda initialization, and probably make t const too.

KKhanhH: actually does this not work:  const static tm t{.tm_mday = 1};

PringlesGang: The lambda initialization is a great idea! The reason I replaced the brace initialization in the first place is because `tm` doesn't have default values set for its members, meaning if you set _one_ of them, the compilers will warn that you need to set all of them (which is a warning I'd rather not turn off). We cannot simply set all of them, because g++ also adds `tm_gmtoff` and `tm_zone` members, which would give errors on other compilers like MSVC.

PR #230, opened 2025-08-11
Switch improvements and fixes & logging/arg stuff

KKhanhH:
File based argument support (profile.txt is basically space delimited cli args now)
Log File
Fix writing on switch


src/io/physicalfilestream.cpp
```diff
@@ -10,20 +10,22 @@ namespace Impacto {
...
-  if (fileExists == IoError_Fail) {
-    ErrorCode = IoError_Fail;
+  ErrorCode = PathExists(SourceFileName);
+  if (ErrorCode == IoError_Fail) {
     return {};
   }
+  const bool fileExists = [this] {
+    bool exists = ErrorCode != IoError_NotFound;
+    return exists;
+  }();
```
PringlesGang: Why is this a lambda? Why not just `const bool fileExists = ErrorCode != IoError_NotFound;`?

KKhanhH: Leftover code, i used it before to assign errorcode

PR #234, opened 2025-08-17
Music menu scroll

Quessian-Phaedrus-Sternabite:
Added partial scroll functionality and button prompts to music menu.

### DIFF
--- a/src/games/chlcc/musicmenu.cpp
+++ b/src/games/chlcc/musicmenu.cpp
@@ -8,7 +8,7 @@
 #include "../../background2d.h"
 #include "../../vm/interface/input.h"
 #include "../../profile/game.h"
-
+#include "../../inputsystem.h"
 #include "../../ui/widgets/chlcc/trackselectbutton.h"
 
 namespace Impacto {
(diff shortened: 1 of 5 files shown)


profiles/chlcc/vfs.lua
```diff
@@ -1,6 +1,7 @@
 root.Vfs = {
     Mounts = {
         ["script"] = {"games/chlcc/gamedata/script.cpk"},
+        ["scriptdbg"] = {"games/chlcc/gamedata/scriptdbg.cls"},
```
KKhanhH: Don't merge this

src/games/chlcc/musicmenu.cpp
```diff
@@ -274,6 +279,42 @@ void MusicMenu::UpdateInput(float dt) {
       SwitchToTrack(-1);
     }
 
+    const uint32_t btnUp = PAD1R1;
```
KKhanhH: these need to include pad up/down or scrolling with buttons won't work

profiles/chlcc/scriptinput.lua
```diff
@@ -1,16 +1,16 @@
---[[root.Input = {
+root.Input = {
```
KKhanhH: Let's not merge this, these controls are pretty messed up and should be fixed completely in a different PR

src/games/chlcc/musicmenu.cpp
```diff
@@ -285,8 +326,8 @@ void MusicMenu::UpdateInput(float dt) {
     } else if (CurrentUpperBound - button->Id >= 16) {
       CurrentLowerBound = button->Id;
       CurrentUpperBound = button->Id + 15;
-    } else
-      return;
+    }  // else
```
KKhanhH: if these have no effect/break things then you can remove

KKhanhH (changes requested):
A couple of things:
Scrolling doesn't work when you hover over a music button
When you move mouse above/below music button range it scrolls, limit that.

Beyond just scrolling if you wanna tackle it in this PR or another:
Playmode should be clickable
song end causes segfault due to unguarded positioninseconds call

KKhanhH:
> not sure how to fix the hovering issue.
I think you'll need to set hovered to false of the currentlyfocusedelement prior to advance focus. it's keeping mouse hover state, and reverting it back. Possibly could make this change global as well by changing menu::advancefocus, if it doesn't break everything then it wouldn't be bad to do.

KKhanhH:
Any menu with a hoverable element probably. I would test by hovering and then moving selection with arrow keys, or moving mouse between stuff

src/games/chlcc/musicmenu.cpp
```diff
@@ -274,6 +279,42 @@ void MusicMenu::UpdateInput(float dt) {
...
+    if (QueuedMove.has_value() &&
+        (MainItems->MoveAnimation.State != +AnimationState::Playing)) {
+      float deltaY = 0;
+      const bool dirDown = *QueuedMove == FocusDirection::FDIR_DOWN;
+      deltaY += dirDown ? TrackOffset.y : -TrackOffset.y;
+      TurboMoved = DirectionButtonHoldHandler.IsTurbo || upScroll || downScroll;
+      const float animationSpeed =
+          TurboMoved ? MusicDirectionalFocusTimeInterval : 0.3f;
+
+      MainItems->Move({0.0f, -deltaY}, animationSpeed);
```
KKhanhH: remove this move, it doesn't actually do anything

PR #235, opened 2025-08-17
C;HLCC: Various UI/Input Fixes

Enorovan:
- [x] Correctly fixes double transition of menus
- [x] Sub-menus properly go back to System Menu
- [x] Add Mousewheel/Q-E/L1-R1 Input for menus with pages
  - [x] Save Menus
  - [x] Album Menu
  - [x] Trophy Menu
- [x] Fix Menus appearing only once in System
- [x] Have first Movie Menu Entry focused on Show
- [x] Have first Album Menu Entry focused on Show
- [x] Have the CrossFade Animation instead of the dots when Showing and Hiding from SystemMenu
  - [x] Create a new Duration
- [x] Add Save Menus selection markers


PringlesGang (changes requested):
- Sound menu doesn't even open for me. Just kinda weirdly softlocks the game
- In the album menu you can still open a cg when hiding/showing
- Erin shouldn't pop in again when opening a submenu from the system menu
- Backing out in submenus in the system menu is still kinda sluggish at times

Enorovan:
> The buttons are still selectable during in and out. I think you might have to set and reset the buttons' `Enabled` bool on Shown and Hide

Selectable or showing as selected? Because latter is on purpose to match PS3. (Which menus?)

KKhanhH:
> Also forgot to mention that in the music menu, if you hover your mouse just on the top/bottom edges of the list, it can select an entry that isn't fully in view yet, causing the list to jump so that it _is_ fully visible, which could then cascade into selecting a _new_ barely visible entry, meaning you get a weird rapid-scrolling effect. I think either music entries selected through mouse hover shouldn't move the list when the entry is slightly off screen, or entries that are not fully on-screen just shouldn't be selectable with hovering the mouse

Music stuff will probably be handled by the other PR

src/games/chlcc/systemmenu.cpp
```diff
@@ -138,6 +138,7 @@ void SystemMenu::Update(float dt) {
     UpdateSmoothSelection(dt);
     MenuLoop.Update(dt);
     UpdateRunningSelectedLabel(dt);
+    MainItems->UpdateInput(dt);
```
KKhanhH, line 141: remove this since rn mainitems->Update calls updateinput

PringlesGang (changes requested):
- Movie menu still has the selection issues on hide, and highlight issues on show
- Save/load menu only sporadically shows for me?
- I also don't see the save/load entry marker?

PringlesGang (approved):
nvm lgtm lmao

Enorovan:
> nvm lgtm lmao

Homie didn't fetch

image

PR #239, opened 2025-08-26
Global pad input hold/turbo hold

KKhanhH:
- Add pad input functions and variables for hold repeats
 - Probably deprecate turboonholdhandler in favor of new padinputbutton variables for better timing consistency with MAGES games.
 - Update input checks to use pad repeat wherever necessary
 
Places to add PAD hold:
- [x] CCLCC Music Menu
- [x] CCLCC Movie Menu
- [x] CCLCC Tips Menu
- [x] CCLCC Save Menu
- [x] CCLCC Backlog
- [x] CCLCC Options Menu
- [x] CCLCC System Menu
- [x] CCLCC Album
- [x] CHLCC Music Menu
- [x] CHLCC Tips Menu
- [x] CHLCC Album
- [x] CHLCC Save
- [x] CHLCC Backlog
- [x] CHLCC System
- [x] CHLCC Config

Suggestions welcome


KKhanhH:
Places to add PAD hold:
- [x] CCLCC Music Menu
- [ ] CCLCC Tips Menu
- [ ] CCLCC Save Menu
- [ ] CCLCC Backlog
- [ ] CCLCC Options Menu
- [ ] CHLCC Music Menu
- [ ] CHLCC Tips Menu

Suggestions welcome

src/vm/interface/input.cpp
```diff
@@ -25,9 +25,14 @@ void UpdatePADcustomType(int type) {
   }
 }
 
+static std::array<float, 32> PADIsDownTime{};
+static float padRepeatClock = 0.0f;
+static float padAccel1Clock = 0.0f;
+static float padAccel2Clock = 0.0f;
+
```
PringlesGang, line 32: I think you should capitalize the first letter, as they're not really regular local variables

src/games/cclcc/musicmenu.h
```diff
@@ -66,10 +65,10 @@ class MusicMenu : public LibrarySubmenu {
   std::optional<FocusDirection> QueuedMove;
   std::vector<size_t> ShuffleTrackIndices;
   MusicBGs BGWidget;
-  TurboOnHoldHandler DirectionButtonHoldHandler;
   Animation NowPlayingFadeAnimation;
   Widgets::Label NowPlayingTrackName;
   MusicModeButton ModeButton;
+  float HoldTimer = false;
```
PringlesGang: wtf lol

src/games/cclcc/musicmenu.cpp
```diff
@@ -306,12 +301,15 @@ void MusicMenu::UpdateInput(float dt) {
...
-    const int directionShouldFire = DirectionButtonHoldHandler.ShouldFire();
+    const uint32_t directionShouldFire =
+        PADinputButtonRepeatDown & (btnUp | btnDown);
     const bool directionMovement =
         (bool)(directionShouldFire & btnUp || upScroll) ^
         (bool)(directionShouldFire & btnDown || downScroll);
 
+    if ((PADinputButtonWentDown & (btnUp | btnDown)) != 0) HoldTimer = 0.0f;
+    if (directionShouldFire) HoldTimer += dt;
+
```
PringlesGang, line 312: This doesn't seem right? Shouldn't it update `HoldTimer` so long as either button _is_ pressed down?
`if (PADinputButtonIsDown & (btnUp | btnDown)) HoldTime += dt;`?

src/vm/interface/input.cpp
```diff
@@ -100,13 +105,80 @@ static void UpdateFromPADCode(int PADcode, PADInputType type) {
...
+      PADIsDownTime[i] = 0.0f;
+      continue;
+    }
+
+    PADIsDownTime[i] += dt;
+
+    if (0 < PADIsDownTime[i] && PADIsDownTime[i] < 2 * frameTime) {
+      PADinputButtonRepeatDown |= PADcode;
+      PADinputButtonRepeatAccelDown |= PADcode;
+    }
```
PringlesGang: I think it's safer to do this _after_ the `if ((PADinputButtonIsDown & PADcode) == 0)` check, but _before_ the `PADIsDownTime[i] += dt` and then just check `if (PADIsDownTime[i] <= std::numeric_limits<float>::epsilon())`
This could fuck up if `dt` is ever greater than 2/60

src/vm/interface/input.cpp
```diff
@@ -100,13 +105,80 @@ static void UpdateFromPADCode(int PADcode, PADInputType type) {
...
+
+      } else {
+        if (accel2Cycle) PADinputButtonRepeatAccelDown |= PADcode;
+      }
+    }
+
+    if (PADIsDownTime[i] >= 30 * frameTime) {
+      if (repeatCycle) {
+        PADinputButtonRepeatDown |= PADcode;
+        if (PADIsDownTime[i] < 90 * frameTime) {
```
PringlesGang: Nitpicky as fuck, but this would skip setting `AccelDown` in case `PADisDownTime[i]` ever just so happens to be _exactly_ `90 * frameTime`, so either one needs to handle the equal case

src/vm/interface/input.cpp
```diff
@@ -100,13 +105,80 @@ static void UpdateFromPADCode(int PADcode, PADInputType type) {
...
+    padRepeatClock = 0.0f;
+  }
+  if (padAccel1Clock >= 2 * frameTime) {
+    accel1Cycle = true;
+    padAccel1Clock = 0.0f;
+  }
+  if (padAccel2Clock >= 1 * frameTime) {
+    accel2Cycle = true;
+    padAccel2Clock = 0.0f;
+  }
```
PringlesGang, line 134: I wonder if this should set the clocks to `padRepeatClock = std::fmod(padRepeatClock, 4 * frameTime)` instead, so it's actually properly time-based, instead of technically tick-based...
Could also see that as a needless computation for something nobody'll ever notice :p

PringlesGang (approved):
Can the `TurboOnHoldHandler` go now? Or do you think there may still be some use for it at some point?

src/vm/interface/input.cpp
KKhanhH: That computation alone would probably take longer than the theoretical time difference

src/games/cclcc/musicmenu.cpp
KKhanhH: This is basically just manual accel, if it does the repeat long enough then it shifts into turbo move with a defined animation time

src/vm/interface/input.cpp
KKhanhH: changing this to just check for PADinputButtonWentDown & PADcode, looks a lot cleaner this way

Issue #153, opened 2025-04-14
Implement all StringTokens (in particular center doesn't work)


Enorovan:
![Image](https://github.com/user-attachments/assets/9257309a-95e7-4878-a9e2-5311c16b0112)

Issue #134, opened 2025-04-10
Implement automatic linebreak through system strings


Enorovan:
It's supposed to use
```
0:/—().,!?  、。.,?!〜”ー)〕]}〉》」』】☆★♪々ぁぃぅぇぉっゃゅょァィゥェォッャュョ[...]-
1:@/,“‘((〔[{〈《「『【
```
in system

PR #240, opened 2025-08-26
String token fixes

PringlesGang:
- [x] Adds chlcc linebreak on puncutation
- [x] Simplifies linebreak on punctuation profile settings
- [x] Fixes `DialoguePage` center alignment
- [x] Impelements `STT_PrintInParallel` string token type #153
- [ ] Implements `STT_AutoAdvance` string token type #153
- [ ] Adds colors to system messages
- [ ] #134


profiles/common/charset.lua
```diff
@@ -0,0 +1,35 @@
...
+        local charsetIndex = 0;
+        for targetBytePos, targetCodePoint in utf8.codes(root.CharsetInternal.CharsetStr) do
+            if sourceCodePoint == targetCodePoint then
+                root.Charset.Flags[charsetIndex] = root.Charset.Flags[charsetIndex] | flag;
+            end
+            
+            charsetIndex = charsetIndex + 1;
+        end
+    end
+end
```
KKhanhH: Internalstr should be longer so i think we can probably flip for loop nesting levels and add an early break when found since codepoints shouldn't be the same twice

src/text.cpp
```diff
@@ -186,56 +186,195 @@ uint8_t ProcessedTextGlyph::Flags() const {
...
+      const float progressAdded = progressLeft * secondsLeftFractionCompleted;
+      dt = progressAdded;
+    }
+  }
+
   UpdateImpl(dt);
 }
 
-float TypewriterEffect::CalcOpacity(int glyph) {
+std::pair<size_t, size_t> TypewriterEffect::GetParallelBlock(
```
KKhanhH: Let's return a struct so there's no chance of mixing up variable order. Structured bindings will still work the same.

src/text.cpp
```diff
@@ -842,26 +1029,42 @@ std::pair<int, float> TextLayoutPlainLineHelper(
...
+        break;
 
-    ptg.DestRect.X = currentX;
-    ptg.DestRect.Y = pos.y;
-    ptg.DestRect.Width = std::floor((fontSize / font->BitmapEmWidth) *
-                                    font->AdvanceWidths[ptg.CharId]);
-    ptg.DestRect.Height = fontSize;
+      case STT_SetColor: {
+        if (253 <= token.Val_Expr && token.Val_Expr <= 255) {
+          token.Val_Expr = ScrWork[2161 + (255 - token.Val_Expr)];
```
KKhanhH: Lua the script variable

src/text.cpp
```diff
@@ -774,6 +954,13 @@ void DialoguePage::Move(glm::vec2 relativePos) {
       Name[i].DestRect.Y += relativePos.y;
     }
   }
+  for (size_t rubyChunkId = 0; rubyChunkId < RubyChunkCount; rubyChunkId++) {
```
KKhanhH: you could probably make all the ruby loops a for range if you use the span constructor

src/text.cpp
```diff
@@ -539,11 +696,15 @@ void DialoguePage::AddString(Vm::Sc3VmThread* ctx, Audio::AudioStream* voice,
...
         LastWordStart = Glyphs.size();
+        BuildingRubyBase = true;
+        break;
+      }
+      case STT_AutoForward: {
+        AutoForward = AutoForwardType::SyncVoice;
         break;
       }
-      case STT_AutoForward:
       case STT_AutoForward_1A: {
```
KKhanhH: we can rename these case labels if we know what they are now

src/text.cpp
```diff
@@ -186,56 +186,195 @@ uint8_t ProcessedTextGlyph::Flags() const {
...
+
+  const size_t parallelBlockGlyphNo = glyph - parallelBlockFirstGlyph;
+
+  // We start displaying a glyph after the previous one is 25% opaque, hence
+  // totalDisplayTime = glyphCount * (0.25 * glyphDisplayTime) +
+  //                    glyphDisplayTime * 0.75
+
+  constexpr float singleGlyphDuration = 1.0f;
+  constexpr float glyphPropagateDuration = singleGlyphDuration * 0.25f;
+  const float totalDuration = parallelBlockGlyphCount * glyphPropagateDuration +
```
KKhanhH: instead of hard coding 0.75 you should make it dependent on propagate duration

src/text.cpp
```diff
@@ -186,56 +186,195 @@ uint8_t ProcessedTextGlyph::Flags() const {
...
-    timePerGlyph = DurationIn;
+  const auto [startProgress, endProgress] = GetGlyphWritingProgresses(glyph);
+
+  if (IsCancelled) {
+    // On cancellation, all non-opaque glyphs start appearing simultaneously
+
+    // Opaque glyphs remain opaque
+    if (ProgressOnCancel >= endProgress) return 1.0f;
+
+    // Transparant glyphs fade in according to the progress made
```
KKhanhH: Transparent

KKhanhH (changes requested):

profiles/common/charset.lua
PringlesGang: I agree with the early break, but shouldn't the `characterString` loop still be the outer one?
We know for certain we'll need to iterate every character in there to set the flags, but those can be found early in the charset string—which doesn't necessarily ever have to be fully traversed

KKhanhH: Ah, I didn't look at the strings that closely. If thats the case then yeah, do that.

KKhanhH (approved):
Bump version too

PR #245, opened 2025-08-30
Chlcc/tips notification

PringlesGang:
- Implements the C;H LCC tips notification
- Fixes unlocked tips not properly loading from system data
- Allows for adding Out animations to `SequencedAnimations`


src/animation.h
```diff
@@ -27,26 +31,43 @@ class Animation {
...
+    else
+      StartOut(reset);
+  }
+
+  constexpr static float GetInitialProgress(AnimationDirection direction) {
+    return direction == AnimationDirection::In ? 0.0f : 1.0f;
+  }
+  constexpr static float GetFinishedProgress(AnimationDirection direction) {
+    return 1.0f - GetInitialProgress(direction);
+  }
```
KKhanhH: Not sure that these make sense for loop modes.

src/games/chlcc/tipsnotification.cpp
```diff
@@ -0,0 +1,136 @@
...
+    // Start display animation
+    const auto tipNameAdr = NotificationQueue.front();
+    const auto tipsScrBufId = TipsSystem::GetTipsScriptBufferId();
+    TipName.SetText({.ScriptBufferId = tipsScrBufId, .IpOffset = tipNameAdr},
+                    TextFontSize, RendererOutlineMode::BottomRight,
+                    static_cast<int>(TipNameColorIndex));
+    NotificationQueue.pop();
+
+    TipsAnimation.StartIn(true);
+    TipsAnimation.Update(dt);
```
KKhanhH: Is the update necessary

src/games/chlcc/tipsnotification.cpp
```diff
@@ -0,0 +1,136 @@
...
+  TextPartAfter.Render();
+
+  Renderer->DisableScissor();
+}
+
+void TipsNotification::AddTip(const int tipId) {
+  FirstInQueue |= TipsAnimation.State == +AnimationState::Stopped &&
+                  NotificationQueue.empty();
+
+  auto record = TipsSystem::GetTipRecord(tipId);
```
KKhanhH: Make this auto* so it's obvious that its a pointer

src/sequencedanimation.h
```diff
@@ -7,18 +7,23 @@ struct ChildAnimation {
...
+                    AnimationDirection direction = AnimationDirection::In) {
+    AddAnimation(animation, startTime, animation.GetDuration(direction),
+                 direction);
   }
 
-  void AddAnimation(Animation& animation) {
-    AddAnimation(animation, DurationIn);
+  void AddAnimation(Animation& animation,
+                    AnimationDirection direction = AnimationDirection::In) {
+    AddAnimation(animation, DurationIn, direction);
```
KKhanhH, line 26: Should Duration depend on direction here?

KKhanhH (changes requested):

src/sequencedanimation.h
PringlesGang: This `DurationIn` is the _sequence's_ duration, so that it places the child at the end of the sequence, so this shouldn't depend on the child's direction

src/animation.h
PringlesGang: Yeah you're right, forgot looping was a thing lol
I removed them wholesale; a repeated ternary isn't too bad for 0 and 1

KKhanhH (approved):
LGTM, bump up version too

PR #246, opened 2025-09-02
Chlcc/config menu

PringlesGang:
Implements the chlcc config menu #201


src/ui/widgets/cclcc/optionsentry.cpp
```diff
@@ -73,6 +73,11 @@ void OptionsEntry::UpdateInput(float dt) {
   if (Selected && PADinputButtonWentDown & PAD1B) {
     Selected = false;
     Audio::Channels[Audio::AC_SSE]->Play("sysse", 3, false, 0.0f);
+
+    // Hack to prevent closing the menu entirely
```
KKhanhH: You could also use AllowsScriptInput instead and check for that in the instruction

src/ui/optionsmenu.cpp
```diff
@@ -125,15 +133,15 @@ void OptionsMenu::UpdateEntryMovementInput(float dt) {
 void OptionsMenu::UpdateInput(float dt) {
   UpdatePageInput(dt);
 
-  if (GetControlState(CT_Back)) {
+  if (IsFocused && GetControlState(CT_Back)) {
```
KKhanhH: Is this supposed to actually be handled by the menu or is it supposed to be scripts? most menu inputs are script based

src/ui/widgets/chlcc/optionsslider.cpp
```diff
@@ -0,0 +1,146 @@
...
+                             const std::function<void(OptionsEntry*)> highlight,
+                             const std::optional<Sprite> mutedSprite)
+    : OptionsEntry(hoverBounds, highlight),
+      OldProgress(value),
+      Slider(0,
+             topRight + SliderBarTopRightOffset + SliderBarFillOffset -
+                 glm::vec2(bar.ScaledWidth(), 0.0f),
+             min, max, &value, ScrollbarDirection::SBDIR_HORIZONTAL,
+             fill.Bounds.GetSize()),
+      BarSprite(bar),
```
KKhanhH, line 28: Are the barsprite and fillsprite necessary or can you use the scrollbar's tracksprite and fillsprite instead?

src/ui/widgets/chlcc/optionsbutton.h
```diff
@@ -0,0 +1,43 @@
...
+namespace Impacto {
+namespace UI {
+namespace Widgets {
+namespace CHLCC {
+
+template <typename T>
+class OptionsButton : public OptionsEntry {
+ public:
+  OptionsButton(T& value, std::span<const T> optionsValues,
+                std::span<const Sprite*> optionsSprites, glm::vec2 topRight,
```
KKhanhH: Since these aren't being checked for nullability, i would make them reference_wrappers instead

src/ui/widgets/chlcc/optionsbutton.cpp
```diff
@@ -0,0 +1,102 @@
...
+  TopRight += relativePos;
+  EntryButton.Move(relativePos);
+}
+
+template <typename T>
+size_t OptionsButton<T>::GetCurrentOptionId() {
+  const auto found = std::ranges::find(OptionsValues, Value);
+
+  if (found == OptionsValues.end()) {
+    ImpLogSlow(LogLevel::Error, LogChannel::General,
```
KKhanhH: Maybe it wouldn't be a bad idea to set Value if it's in an invalid range

src/ui/widgets/chlcc/optionsbutton.h
```diff
@@ -0,0 +1,43 @@
...
+  void UpdateValue() override { OptionId = GetCurrentOptionId(); };
+
+  void Move(glm::vec2 relativePos) override;
+
+ private:
+  size_t GetCurrentOptionId();
+
+  T& Value;
+
+  std::vector<T> OptionsValues;
```
KKhanhH: Does this need to be a vector? if using profile variables the lifetime should be ok with a span

src/games/chlcc/optionsmenu.h
```diff
@@ -5,32 +5,70 @@
 #include "../../ui/widgets/button.h"
 #include "../../ui/widgets/carousel.h"
 #include "../../ui/widgets/label.h"
+#include "../../ui/widgets/chlcc/optionsentry.h"
 
 namespace Impacto {
 namespace UI {
 namespace CHLCC {
 
+using namespace Impacto::UI::Widgets::CHLCC;
```
KKhanhH: i don't think this should be in the header

src/games/chlcc/savesystem.cpp
```diff
@@ -42,6 +42,36 @@ constexpr uint8_t AutoSpeedToSettingIndex(const float speed) {
...
+constexpr uint8_t AutoQuickSaveSettingToIndex(const uint8_t setting) {
+  const bool onScene = setting & AutoQuickSaveType::OnScene;
+  const bool onTrigger = setting & AutoQuickSaveType::OnTrigger;
+
+  constexpr uint8_t onTriggerIndex = 0;
+  constexpr uint8_t onSceneIndex = 1;
+  constexpr uint8_t onTriggerAndSceneIndex = 2;
+  constexpr uint8_t neverIndex = 3;
+
+  return onScene ? (onTrigger ? onTriggerAndSceneIndex : onSceneIndex)
```
KKhanhH: if elses is probably clearer here

src/games/chlcc/optionsmenu.cpp
```diff
@@ -11,39 +14,256 @@
...
+
+  HighlightedEntriesPerPage.reserve(Pages.size());
+  for (std::unique_ptr<Group>& page : Pages) {
+    HighlightedEntriesPerPage.push_back(page->GetFirstFocusableChild());
+  }
+
+  SelectedLabelPos = HighlightedEntriesPerPage[CurrentPage]->Bounds.GetPos();
+}
+
+std::unique_ptr<Widgets::Group> OptionsMenu::CreateTextPage(
```
KKhanhH: It looks like all 3 the createpage functions can be made static to the file, just need to add argument for the menu pointer.

src/games/chlcc/optionsmenu.cpp
```diff
@@ -11,39 +14,256 @@
...
+            value, values, sprites,
+            highlightBounds.GetPos() + glm::vec2(highlightBounds.Width, 0.0f),
+            highlightBounds, highlight),
+        FDIR_DOWN);
+  };
+
+  // Basic settings
+  {
+    std::array<const Sprite*, 2> sprites{&SettingDoSprite, &SettingDontSprite};
+    addButton.template operator()<bool>(0, TriggerStopSkip,
```
KKhanhH: could also wrap the array with std::span\<const bool\> and let template deduction take over, that might be cleaner than template operator()

src/games/chlcc/optionsmenu.cpp
```diff
@@ -11,39 +14,256 @@
...
+        new OptionsButton<T>(
+            value, values, sprites,
+            highlightBounds.GetPos() + glm::vec2(highlightBounds.Width, 0.0f),
+            highlightBounds, highlight),
+        FDIR_DOWN);
+  };
+
+  // Basic settings
+  {
+    std::array<const Sprite*, 2> sprites{&SettingDoSprite, &SettingDontSprite};
```
KKhanhH: If it's always the same sprite variable, you could make the array constexpr static, then on the button side it would only need to store a span of sprite pointers,

src/games/chlcc/optionsmenu.cpp
```diff
@@ -11,39 +14,256 @@
...
+void OptionsMenu::Hide() {
+  if (State == Shown) {
+    SetFlag(SF_SUBMENUEXIT, true);
+    ShowPageAnimation.StartOut();
+  }
+
+  UI::OptionsMenu::Hide();
+
+  if (State == Hiding) {
+    Pages[CurrentPage]->IsShown = true;
```
KKhanhH, line 250: we really do need to decouple hide from isshown somehow in group so we can keep hide animations without having to do what feels like hacks

src/vm/inst_system.cpp
```diff
@@ -988,20 +1001,116 @@ VmInstruction(InstAutoSave) {
...
+      if (ScrWork[SW_AUTOSAVERESTART] != 3 &&
+          (autoQuickSave & AutoQuickSaveType::OnTrigger)) {
+        quickSave(3);
+      }
+
+      ScrWork[SW_AUTOSAVERESTART] = 0;
+    } break;
+
+    case 0xff: {
+      SetFlag(1206, 1);
```
KKhanhH: SF_SAVECAPTURE it looks like

src/vm/inst_system.cpp
```diff
@@ -988,20 +1001,116 @@ VmInstruction(InstAutoSave) {
...
+    case 5: {  // EnableAutoSave
+      SaveSystem::SaveMemory();
+
+      SetFlag(SF_AUTOSAVEENABLE, 1);
+      ScrWork[SW_AUTOSAVERESTART] = 0;
+    } break;
+
+    case 10: {  // SetCheckpointId
+      BlockThread;
+      SetFlag(1206, 1);
```
KKhanhH: SF_SAVECAPTURE

KKhanhH (changes requested):

src/ui/optionsmenu.cpp
PringlesGang: It's supposed to be handled by the menu in C;C LCC (because that menu actually makes use of the back button internally). C;H LCC _does_ do this by script, yeah. And since C;C LCC already handles this in its derived method, I'll just remove it here

src/ui/widgets/chlcc/optionsslider.cpp
PringlesGang: `Slider` assumes the bounds of the bar to be the bounds of the fill, which is not the case here. I have changed it so that the fill sprite is handled by the Slider, but left the bar outside

src/games/chlcc/optionsmenu.cpp
PringlesGang: Agreed. Separate PR tho

KKhanhH (approved):
Versioning and all that

PR #253, opened 2025-09-09
Widget move fixes

PringlesGang:
Changes the `Widget` class's Move API such that only `Move(glm::vec2 offset)` ever needs to be overridden—with `MoveTo` and the timed Move(To)s being supplied automatically

- Adds optional anchor point for `MoveTo` to set to the desired position instead of the usual `Bounds.GetPos()`
- Removes all superfluous (and sometimes wrong) `MoveTo` overrides
- Removes `RenderingBounds` check in `Group::Render` as it does not take out-of-`Bounds` rendering into account (and should really only be checked by the renderer anyway)
- Adds a few useful methods to `RectF` and `Rect`
- ~~Free small C;H LCC trophy menu refactor I did while debugging an issue I introduced to it \o/ (feel free to reject as the issue was not actually in the trophy menu itself)~~

### DIFF
--- a/src/ui/widget.h
+++ b/src/ui/widget.h
@@ -34,15 +34,20 @@ class Widget {
   // TODO: Text movement in widgets with text
   virtual void Move(glm::vec2 relativePosition) { Bounds += relativePosition; }
   void Move(glm::vec2 relativePosition, float duration);
-  void MoveTo(glm::vec2 pos) { Move(pos - Bounds.GetPos()); }
+  void MoveTo(glm::vec2 pos) {
+    Move(pos - MoveToAnchor.value_or(Bounds.GetPos()));
+  }
   void MoveTo(glm::vec2 pos, float duration) {
-    Move(pos - Bounds.GetPos(), duration);
+    Move(pos - MoveToAnchor.value_or(Bounds.GetPos()), duration);
   }
 
   virtual Widget* GetFocus(FocusDirection dir);
   virtual void SetFocus(Widget* widget, FocusDirection dir);
 
   RectF Bounds = RectF(0.0f, 0.0f, 0.0f, 0.0f);
+
+  // Will move this point to the desired location with MoveTo
+  std::optional<glm::vec2> MoveToAnchor;
 
   glm::vec2 MoveTarget;
   glm::vec2 MoveOrigin;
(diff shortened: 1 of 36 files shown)


src/ui/widgets/mo6tw/scenelistentry.h
```diff
@@ -14,8 +14,10 @@ class SceneListEntry : public Widgets::Button {
                  Widgets::Label* unlockedText, Sprite const& highlight,
                  bool isLocked);
   void Render() override;
+
+  using Widget::Move;
+  using Widget::MoveTo;
```
KKhanhH: Is using moveto necessary?

src/ui/widget.h
```diff
@@ -32,16 +32,23 @@ class Widget {
...
+  void MoveTo(glm::vec2 pos, float duration) {
+    Move(pos - MoveToAnchor.value_or(Bounds.GetPos()), duration);
+  }
 
   virtual Widget* GetFocus(FocusDirection dir);
   virtual void SetFocus(Widget* widget, FocusDirection dir);
 
   RectF Bounds = RectF(0.0f, 0.0f, 0.0f, 0.0f);
 
+  // Will move this point to the desired location with MoveTo
```
KKhanhH, line 49: I'm a bit confused on how movetoanchor is supposed to work, could you explain what the intent is?

KKhanhH (changes requested):

src/ui/widget.h
PringlesGang: Normally `MoveTo` moves the widget such that `Bounds.GetPos()` becomes the given position, but `MoveToAnchor` is set, `MoveTo` will instead move the widget such that the _anchor_ becomes the given position
This is useful, for example, in a `DialoguePage`, where the name is part of the bounds (and can thus define the top left of it), but when using `MoveTo` you'd rather it relate to the top left of the _actual dialogue_; so `MoveToAnchor` is set to be the top left of the dialogue

src/ui/widgets/mo6tw/scenelistentry.h
PringlesGang: Fair point. It can't be overridden in the first place

KKhanhH (approved):

PR #257, opened 2025-09-26
[CHLCC][CCLCC] Save Locking

Gunoshozo:
CHLCC:
- Fix lock icon not moving back after moving out
- Move lock icon position to a .lua file
- Add script flags for "Saving to a locked save", "Locked all Q.Saves", "Lock State changed" and update them in the UI::CHLCC::SaveMenu class
- Add ability to toggle lock on SaveEntryButton and render button accordingly

CCLCC:
- Fix additional render of SceneTitleLabel for SaveEntryButton when in locked state
- Move SlotLockedSpritePosition and NoDataSpritePosition to lua profile and also correct the values
- Add ability to toggle lock on SaveEntryButton and render button accordingly
- Add script flags for "Saving to a locked save", "Locked all Q.Saves" (game script does not care for "Lock State changed") and update them in the UI::CCLCC::SaveMenu class
- Add sounds for lock action on regular and empty slot, and for switching between save slots

CCLCC & CHLCC:
- Added LockedQuickSaveSlots static field to the SaveSystem class to count initial number of locked quick saves so that other menus could access that value, should be useful for blocking "quick save" button in system menu (which itself requires even more changes and I decided to not go there, since this PR was already getting too big)

CCLCC, CHLCC, MO6TW:
- Add SetSaveFlags virtual method to SaveSystem and added implementations based on the GetSaveFlags

Other:
- Remove redundant IsLocked from CCLCC::SaveEntryButton, using Button::IsLocked like in CHLCC::SaveEntryButton (this allows to also improve memory layout for CCLCC::SaveEntryButton for additional -8 bytes)
- Improve memory layout of Impacto::UI::Widgets::Button and Impacto::UI::Menu (-8 bytes for every kind of button and menu in the app)

CCLCC video comparison with native ps4
https://youtu.be/-GJysBTyjJ0

CHLCC video comparison with native ps3
https://youtu.be/z_8AGs4iwxw


profiles/cclcc/scriptvars.lua
```diff
@@ -36,6 +36,9 @@ sv.SW_SVSCRNO4 = 3609;
 sv.SW_DELUSION_STATE = 6412;
 sv.SW_YESNO_PRI = 6437;
 
+sv.SF_SAVINGLOCKEDSAVE = 1245;
```
KKhanhH: 1245:SF_SAVEPROTECTED

profiles/cclcc/scriptvars.lua
```diff
@@ -36,6 +36,9 @@ sv.SW_SVSCRNO4 = 3609;
...
+sv.SF_ALLQUICKSAVESLOCKED = 1246;
```
KKhanhH: 1246:SF_SAVEALLPROTECTED

profiles/cclcc/scriptvars.lua
```diff
@@ -36,6 +36,9 @@ sv.SW_SVSCRNO4 = 3609;
...
+
```
KKhanhH: 1247: SF_SAVEPROTECTCHANGED

Move these all into common as well

profiles/chlcc/scriptvars.lua
```diff
@@ -152,6 +152,9 @@ sv.SW_MOVIEPRI2 = 3400;
 sv.SW_MOVIEALPHA2 = 3421;
 sv.SW_DELUSION_STATE = 3430;
 --FlagWork
+sv.SF_SAVINGLOCKEDSAVE = 1245;
```
KKhanhH: move into common

src/games/cclcc/savesystem.cpp
```diff
@@ -26,6 +26,8 @@ using namespace Impacto::Profile::ScriptVars;
 using namespace Impacto::Profile::Vm;
 using namespace Impacto::Profile::ConfigSystem;
 
+int SaveSystem::LockedQuickSaveSlots = 0;
```
KKhanhH: I think this would be a lot better as a regular member of the base savesystem class, it's basically gonna be the same in every game anyway. Just need to add a getter and probably setter for it.. 

Also you can init static members in a header with the inline keyword I believe.

src/ui/widgets/chlcc/saveentrybutton.h
```diff
@@ -53,6 +53,7 @@ class SaveEntryButton : public Widgets::Button {
 
   void RefreshInfo(SaveSystem::SaveType entryType);
 
+  void ToggleLock(const SaveSystem::SaveType entryType);
```
KKhanhH: Don't declare func params as const in header

src/ui/widgets/cclcc/saveentrybutton.h
```diff
@@ -37,20 +38,18 @@ class SaveEntryButton : public Widgets::Button {
  private:
   int Index;
   int Page;
+  int SaveStatus;
```
KKhanhH, line 41: You'll have to update the initializer list order too in the constructor definition when you reorder members

KKhanhH (changes requested):
Thanks for the PR, it looks good, just a few minor changes. You'll need to fix up the builds and rebase on master as well, I also recommend having format on save on since we do have a clang-format file

KKhanhH (approved):

PR #258, opened 2025-10-01
C;H LCC script crash fixes

PringlesGang:
- Fixes crash on butterfly scene in Ayase route
  - Gracefully handles empty script expressions
  - Adds stub for InstMesSync
- Fixes crash upon returning to the title screen in C;H LCC
- Properly resets selected element when exiting title screen

### DIFF
--- a/src/vm/expression.cpp
+++ b/src/vm/expression.cpp
@@ -94,16 +94,14 @@ class ExpressionParser {
 };
 
 int ExpressionEval(Sc3VmThread* thd, int* result) {
-  ExpressionParser* parser = new ExpressionParser(thd);
+  std::unique_ptr<ExpressionParser> parser =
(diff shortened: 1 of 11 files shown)


src/vm/expression.cpp
```diff
@@ -94,16 +94,14 @@ class ExpressionParser {
...
-
   std::unique_ptr<ExpressionNode> rootPtr =
       std::unique_ptr<ExpressionNode>(root);
 
-  *result = rootPtr->Evaluate(thd);
-
-  delete parser;
+  *result = root == nullptr ? 0 : rootPtr->Evaluate(thd);
 
   return 0;
```
KKhanhH: We should maybe make this just return the result, this return code seems useless

KKhanhH (approved):

PR #276, opened 2025-10-10
Fix skip bug plus add missing sound to CCLCC SystemMenu.

dendi2008:
This PR will: 
- Fix skip bug (issue #267). 
- Add the missing Selection sound that plays when switching between buttons (Backlog, Save, etc.) in CCLCC System Menu using a gamepad or keyboard.

### DIFF
--- a/src/games/cclcc/systemmenu.cpp
+++ b/src/games/cclcc/systemmenu.cpp
@@ -4,6 +4,7 @@
 #include "../../renderer/renderer.h"
 #include "../../ui/ui.h"
 #include "../../vm/interface/input.h"
+#include "../../inputsystem.h"
 #include "../../profile/ui/systemmenu.h"
 #include "../../profile/game.h"
 #include "../../ui/widgets/cclcc/sysmenubutton.h"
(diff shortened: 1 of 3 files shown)


src/ui/optionsmenu.cpp
```diff
@@ -72,6 +74,14 @@ void OptionsMenu::Hide() {
...
       CurrentlyFocusedElement = nullptr;
     }
+
+    if (Impacto::MesSkipMode &
+        (Impacto::SkipModeFlags::SkipRead | Impacto::SkipModeFlags::SkipAll)) {
+      Impacto::MesSkipMode &= Impacto::SkipModeFlags::Auto;
+      Impacto::MesSkipMode |= (Impacto::Profile::ConfigSystem::SkipRead
+                                   ? Impacto::SkipModeFlags::SkipRead
+                                   : Impacto::SkipModeFlags::SkipAll);
+    }
```
PringlesGang: This does work, but I don't think this is quite the right place to handle this logic. I think this can better be done in `ChkMesSkip` with the rest of the skip logic

(Also, you're already in the `Impacto` namespace, so you don't have to specify it anymore)

dendi2008: > This does work, but I don't think this is quite the right place to handle this logic. I think this can better be done in `ChkMesSkip` with the rest of the skip logic
> 
> (Also, you're already in the `Impacto` namespace, so you don't have to specify it anymore)

Done.

src/ui/optionsmenu.cpp
```diff
@@ -172,4 +172,4 @@ void OptionsMenu::Highlight(Widget* toHighlight) {
 }
 
 }  // namespace UI
-}  // namespace Impacto
\ No newline at end of file
+}  // namespace Impacto
```
KKhanhH: Accidental commit?

dendi2008: Yeah. But I fixed that.

Issue #264, opened 2025-10-10
Fix controller dpad input for scrollbars


PringlesGang:
Holding the dpad for a scrollbar does not keep moving the scrollbar like it does with keyboard
(E.g. backlog menu, config sliders)

PR #277, opened 2025-10-12
Fix controller dpad input for scrollbars and sliders.

dendi2008:
This PR will fix controller dpad input for scrollbars and sliders. (Issue #264 )

### DIFF
--- a/src/vm/interface/input.cpp
+++ b/src/vm/interface/input.cpp
@@ -45,6 +45,9 @@ static void UpdateFromPADCode(int PADcode, PADInputType type) {
   const auto& kbDownArr = type == PADInputType::WentDown
                               ? Input::KeyboardButtonWentDown
                               : Input::KeyboardButtonIsDown;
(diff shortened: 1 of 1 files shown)


src/vm/interface/input.cpp
```diff
@@ -55,8 +55,11 @@ static void UpdateFromPADCode(int PADcode, PADInputType type) {
       KBcodes != PADToKeyboard.end() &&
       std::any_of(KBcodes->second.begin(), KBcodes->second.end(),
                   [&kbDownArr](auto KBcode) { return kbDownArr[KBcode]; });
-  const bool isGpDown = GPcode != PADToController.end() &&
-                        Input::ControllerButtonWentDown[GPcode->second];
+  const auto& gpDownArr = type == PADInputType::WentDown
```
KKhanhH: Move this next to the other down arrs

dendi2008: > Move this next to the other down arrs

Done.

PR #278, opened 2025-10-12
Fix word wrapping

PringlesGang:
Will fix word wrapping issues for sc3 strings

- [x] Read word starting/ending punctuation from script
- [x] Properly implement word wrapping logic for EN

### DIFF
--- a/src/text.h
+++ b/src/text.h
@@ -26,6 +26,66 @@ BETTER_ENUM(SkipModeFlags, uint8_t, SkipRead = (1 << 0), SkipAll = (1 << 1),
 
 // TODO: think about / profile memory access patterns
 
+enum StringTokenType : uint8_t {
+  STT_LineBreak = 0x00,
+  STT_CharacterNameStart = 0x01,
+  STT_DialogueLineStart = 0x02,
+  STT_Present = 0x03,
+  STT_SetColor = 0x04,
+  STT_Present_Clear = 0x08,
+  STT_RubyBaseStart = 0x09,
+  STT_RubyTextStart = 0x0A,
+  STT_RubyTextEnd = 0x0B,
+  STT_SetFontSize = 0x0C,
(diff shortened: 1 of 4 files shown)


src/text.h
```diff
@@ -26,6 +26,64 @@ BETTER_ENUM(SkipModeFlags, uint8_t, SkipRead = (1 << 0), SkipAll = (1 << 1),
...
+ public:
+  StringTokenType Type;
+
+  uint16_t Val_Uint16;
+  int Val_Expr;
+
+  uint8_t Flags{};
+
+  static void AddFlags(Vm::BufferOffsetContext scrCtx, uint8_t flags);
+  static inline void AddFlags(uint16_t glyphId, uint8_t flags) {
```
KKhanhH: You don't need inline for anything defined in a class/struct

src/text.h
```diff
@@ -26,6 +26,64 @@ BETTER_ENUM(SkipModeFlags, uint8_t, SkipRead = (1 << 0), SkipAll = (1 << 1),
...
+    if (StringToken::FlagsMap.contains(glyphId)) {
+      StringToken::FlagsMap[glyphId] |= flags;
+    } else {
+      StringToken::FlagsMap.emplace(glyphId, flags);
+    }
+  }
+  static inline uint8_t GetFlags(uint16_t glyphId) {
+    return FlagsMap.contains(glyphId) ? FlagsMap[glyphId] : 0;
```
KKhanhH: Use .find, I don't think compilers can optimize out this check, especially since operator[] is mutable.

src/text.h
PringlesGang: `FlagsMap` _does_ seem to need it. Otherwise the linker complains about it being an unresolved external symbol

KKhanhH: Yes, variables behave differently than functions. It's also the same as how constexpr functions/variables behave with/without inline

PR #279, opened 2025-10-13
Cpk load fix

KKhanhH:
Fixes issue preventing coalesc3 generated cpks from being loaded

### DIFF
--- a/src/io/cpkarchive.cpp
+++ b/src/io/cpkarchive.cpp
@@ -65,6 +65,7 @@ void DecryptUtfBlock(uint8_t* utfBlock, uint64_t size) {
 }
 
 static std::string ReadString(int64_t stringsOffset, Stream* utfStream) {
+  assert(utfStream);
   int64_t stringAddr = stringsOffset + ReadBE<uint32_t>(utfStream);
 
   int64_t retAddr = utfStream->Position;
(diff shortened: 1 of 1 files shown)


src/io/cpkarchive.h
```diff
@@ -3,28 +3,14 @@
 #include "vfsarchive.h"
 #include "memorystream.h"
 #include <vector>
+#include <variant>
 #include <ankerl/unordered_dense.h>
```
PringlesGang, line 6: These can move to the implementation file

src/io/cpkarchive.cpp
```diff
@@ -53,126 +62,152 @@ void DecryptUtfBlock(uint8_t* utfBlock, uint64_t size) {
...
+  utfStream->Seek(stringAddr, RW_SEEK_SET);
+
+  char ch;
+  std::string output;
+  while ((ch = ReadU8(utfStream)) != 0x00) {
+    output.push_back(ch);
+  }
+  utfStream->Seek(retAddr, RW_SEEK_SET);
+  return output;
+}
```
PringlesGang, line 82: Maybe assert that `utfStream != nullptr`? Or just make it a ref

src/io/cpkarchive.cpp
```diff
@@ -53,126 +62,152 @@ void DecryptUtfBlock(uint8_t* utfBlock, uint64_t size) {
...
     std::vector<ankerl::unordered_dense::map<std::string, CpkCell, string_hash,
-                                             std::equal_to<>>>* rows) {
+                                             std::equal_to<>>>& rows) {
   const uint32_t utfMagic = 0x40555446;
-  UtfStream = new MemoryStream(utfBlock, utfSize, true);
-  if (ReadBE<uint32_t>(UtfStream) != utfMagic) {
-    DecryptUtfBlock(utfBlock, utfSize);
-    UtfStream->Seek(0, RW_SEEK_SET);
-    if (ReadBE<uint32_t>(UtfStream) != utfMagic) {
+  auto* utfStream = new MemoryStream(utfBlock.data(), utfBlock.size(), false);
```
PringlesGang: Just make it a unique ptr. You're manually deleting it at the end of the scope anyway

src/io/cpkarchive.cpp
```diff
@@ -53,126 +62,152 @@ void DecryptUtfBlock(uint8_t* utfBlock, uint64_t size) {
...
+  }
+  utfStream->Seek(retAddr, RW_SEEK_SET);
+  return output;
+}
+
+static bool ReadUtfBlock(
+    std::vector<uint8_t>& utfBlock,
     std::vector<ankerl::unordered_dense::map<std::string, CpkCell, string_hash,
-                                             std::equal_to<>>>* rows) {
+                                             std::equal_to<>>>& rows) {
```
PringlesGang, line 87: `utfBlock` can be a span

src/io/cpkarchive.cpp
KKhanhH: Making it a ref would be inconsistent with every other function that takes a stream pointer, which is basically all of them.

src/io/cpkarchive.cpp
KKhanhH: imo it's fine here as is since it's not a public function and it's a modifiable vector

PR #282, opened 2025-10-17
Some additions.

dendi2008:
This PR will:
- Add Space bar input support.
- Add WASD keys input support.
- Add missing Selecting sound for YES/NO button in CCLCC SystemMessageBox (sound plays only when user selects button with his keyboard or gamepad).
- Update code (play Selecting sound for CCLCC SystemMenu buttons.)

### DIFF
--- a/src/games/cc/sysmesbox.cpp
+++ b/src/games/cc/sysmesbox.cpp
@@ -7,6 +7,7 @@
 #include "../../profile/scriptvars.h"
 #include "../../mem.h"
 #include "../../renderer/renderer.h"
+#include "../../vm/interface/input.h"
 
 namespace Impacto {
 namespace UI {
@@ -16,6 +17,7 @@ using namespace Impacto::UI::Widgets;
 using namespace Impacto::Profile::ScriptVars;
 using namespace Impacto::Profile::SysMesBox;
 using namespace Impacto::Profile::CC::SysMesBox;
+using namespace Impacto::Vm::Interface;
 
 static float BoxAnimCount = 0.0f;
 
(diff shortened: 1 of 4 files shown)


src/games/cc/sysmesbox.cpp
```diff
@@ -108,6 +110,16 @@ void SysMesBox::Hide() {
...
+void SysMesBox::UpdateInput(float dt) {
+  if (!IsFocused) return;
+  Menu::UpdateInput(dt);
+  if (ChoiceCount == 2) {
+    if (PADinputButtonWentDown & (PAD1LEFT | PAD1RIGHT)) {
+      Audio::Channels[Audio::AC_SSE]->Play("sysse", 1, false, 0);
+    }
+  }
+}
+
```
PringlesGang, line 122: I don't think we wanna needlessly limit ourselves to either the amount of choices, nor the direction of the input (even if as far as I'm aware the C;Cs only ever have at most two choices, laid out horizontally)

A little dirty, but I think something like
```cpp
void SysMesBox::UpdateInput(float dt) {
    if (!IsFocused) return;

    const auto* const prevSelected = CurrentlyFocusedElement;
    Menu::UpdateInput(dt);
    if (CurrentlyFocusedElement && prevSelected == CurrentlyFocusedElement) {
        Audio::Channels[Audio::AC_SSE]->Play("sysse", 1, false, 0.0f);
    }
}
```
would do the trick

PringlesGang (changes requested):

src/games/cc/sysmesbox.cpp
dendi2008: Thanks for suggestion. I fixed my commit.

PR #285, opened 2025-10-20
Optimize text rendering

PringlesGang:
Only send glyphs to the renderer that are actually in the viewport

### DIFF
--- a/src/renderer/renderer.cpp
+++ b/src/renderer/renderer.cpp
@@ -173,18 +173,34 @@ void BaseRenderer::DrawProcessedText(std::span<const ProcessedTextGlyph> text,
   }
 }
 
+static std::vector<size_t> GetVisibleGlyphIds(
+    std::span<const ProcessedTextGlyph> text) {
+  const static RectF viewport{0.0f, 0.0f, Profile::DesignWidth,
+                              Profile::DesignHeight};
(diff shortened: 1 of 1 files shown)


src/renderer/renderer.cpp
```diff
@@ -136,11 +136,18 @@ void BaseRenderer::DrawQuad(const CornersQuad& dest, const glm::vec4 color) {
...
-                                     Font* font, float opacity,
-                                     RendererOutlineMode outlineMode,
-                                     bool smoothstepGlyphOpacity,
-                                     SpriteSheet* maskedSheet) {
+void BaseRenderer::DrawProcessedText(
+    std::span<const ProcessedTextGlyph> sourceText, Font* font, float opacity,
+    RendererOutlineMode outlineMode, bool smoothstepGlyphOpacity,
+    SpriteSheet* maskedSheet) {
+  std::vector<ProcessedTextGlyph> text;
+  std::copy_if(sourceText.begin(), sourceText.end(), std::back_inserter(text),
```
KKhanhH: rather than copying glyphs into a new vector, maybe use a vector of indices to the span that are filtered? similar to chlcc tips menu

KKhanhH (approved):

PR #287, opened 2025-10-22
Lock Backlog & QuickSave Buttons

Gunoshozo:
C:HLCC:
- No blocking backlog button. Judging by the script game does not even check that flag.
- I've implemented blocking of QS button slightly different from the original. Original for some reason allows you to save even when you can't and there will be no effect. I'm blocking the button and setting flags if all slots are blocked and if user Qsaved on current line.

Def proof (though could not open menu fast enough and show empty backlog):
1st - original, 2nd - impacto
https://youtu.be/SjAbbO1Y7vk

C;CLCC:
- Backlog is blocked if either of flags `SF_BACKLOG_NOLOG` or `SF_MESREVDISABLE` is set. Flag `SF_BACKLOG_NOLOG` is set on initialization and on clearing of the backlog. Flag `SF_BACKLOG_NOLOG` is reset when entry is added to backlog
- QS is blocked if Qsaved on the current line or all QS slots are blocked (`SF_SAVEALLPROTECTED`), `SF_SAVEALLPROTECTED` is reset after loading the game, so it's reclculated on SystemMenu::Init. + on each new line SF_SAVEDISABLE is reset only if there's a free slot.

Def proof:
1st - original, 2nd - impacto
https://youtu.be/YHgfKKCpVwc


src/data/savesystem.h
```diff
@@ -131,13 +131,16 @@ class SaveSystemBase {
   Sprite& GetWorkingSaveThumbnail() { return WorkingSaveThumbnail; }
   int GetLockedQuickSaveCount();
   void SetLockedQuickSaveCount(int value);
+  bool HasQSavedOnCurrentLine();
```
PringlesGang: Please mark the method as const

src/games/cclcc/systemmenu.h
```diff
@@ -38,6 +38,20 @@ class SystemMenu : public Menu {
...
+    Backlog,
+    QuickSave,
+    QuickLoad,
+    Save,
+    Load,
+    TipsList,
+    Config,
+    Help,
+    ReturnTitle
+  };
```
PringlesGang: Please turn this (and the chlcc one) into enum classes for automatic namespacing

src/games/cclcc/systemmenu.cpp
```diff
@@ -245,6 +249,25 @@ void SystemMenu::Render() {
...
+}
+
+inline void SystemMenu::LockMenuButton(size_t buttonIndex, bool state) {
+  if (buttonIndex >= MainItems->Children.size()) {
+    return;
+  }
+  SysMenuButton* menuButton =
+      static_cast<UI::CCLCC::SysMenuButton*>(MainItems->Children[buttonIndex]);
+  menuButton->IsLocked = state;
 }
```
PringlesGang: I don't think such a method is really necessary when you can just
`MainItems->Children[MenuItems::QuickSave]->IsLocked = state`

(This all also holds for the chlcc version)

src/ui/backlogmenu.cpp
```diff
@@ -57,6 +57,7 @@ BacklogMenu::BacklogMenu() {
   FadeAnimation.LoopMode = AnimationLoopMode::Stop;
   FadeAnimation.DurationIn = FadeInDuration;
   FadeAnimation.DurationOut = FadeOutDuration;
+  SetFlag(SF_BACKLOG_NOLOG, true);
```
PringlesGang: This should probably only be set in `BacklogMenu::Update`, and actually be set according to `MainItems.Children.empty()`

PringlesGang (changes requested):
Thank you for the PR, it looks really nice! Still some notes tho:

- I think you're confusing `SF_SAVEDISABLE` for `SF_AUTOSAVEENABLE` in most cases? That flag should already be properly implemented in most places
- `SF_MESREVDISABLE` still needs to be set in `InstMSinit`, in case a save is loaded

Gunoshozo:
Resolved merge conflicts

PR #289, opened 2025-10-22
SSE being interrupted when another SSE starts playing fix.

dendi2008:
This PR will fix SSE (System sound effect) being interrupted when another SSE (System sound effect) starts playing. 
Example of SSE being interrupted when another SSE starts playing: In the CCLCC Main Menu, if you click on the locked Extra button, it will play an buzzer sound. If you then immediately click on the Load button, the buzzer sound gets cut off by an entering sound.

### DIFF
--- a/src/games/cc/backlogmenu.cpp
+++ b/src/games/cc/backlogmenu.cpp
@@ -20,7 +20,7 @@ void BacklogMenu::MenuButtonOnClick(Widgets::BacklogEntry* target) {
   UI::BacklogMenu::MenuButtonOnClick(target);
 
   if (target->AudioId == -1)
-    Audio::Channels[Audio::AC_SSE]->Play("sysse", 4, false, 0.0f);
+    Audio::PlayInGroup(Audio::ACG_SE, "sysse", 4, false, 0.0f);
 }
 
 void BacklogMenu::Show() {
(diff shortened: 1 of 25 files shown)


dendi2008:
> Also, rather than make channelids a vector, you could make it an array since the channel ids are known at compile time, and make channelgroups a tuple. For tuple you will need to use a switch to use a runtime group id though.

Can I also make this in my another refactor commit?

KKhanhH:
> > Also, rather than make channelids a vector, you could make it an array since the channel ids are known at compile time, and make channelgroups a tuple. For tuple you will need to use a switch to use a runtime group id though.
> 
> Can I also make this in my another refactor commit?

I would amend the current commit for this change, and make the refactor commit solely for changing play sse stuff to playgroup sse

dendi2008:
> Yeah, that's pretty much what i meant. Although, instead of needing a SSEChannelProxy in the first place, it's probably better to just make AC_SSE into AC_SSE0 as a regular channel, find and replace of function calls related to `[AC_SSE]` into separate group play functions (on another refactor commit). This makes it more generic if we want the same for other channels.
> 
> Also, rather than make channelids a vector, you could make it an array since the channel ids are known at compile time, and make channelgroups a tuple. For tuple you will need to use a switch to use a runtime group id though.
> 
> ```c++
> template <AudioChannelId... Ids>
> struct ChannelGroupDef {
>     static constexpr std::array<AudioChannelId, sizeof...(Ids)> ChannelIds{ Ids... };
>     size_t NextIndex = 0;
> };
> 
> inline constexpr auto ChannelGroups = std::tuple{
>     ChannelGroupDef<AC_BGM0, AC_BGM1>{},
>     ChannelGroupDef<AC_SE0, AC_SE1>{},
> }; 
> ```

Done.

KKhanhH (approved):

PR #292, opened 2025-10-25
Small refactor of gamespecific ui

KKhanhH:
- [x] Move gamespecific ui features like delusion triggers, map, etc, so they have a common entry point.
- [x] Get rid of the base delusion trigger and base mapsystem classes. 
- [x] Separate button guide from mapsystem, it should be a cc gamespecific common thing
- [x] Rename a few public variables/namespaces for consistency between gamespecific sstuff

### DIFF
--- a/src/profile/ui/gamespecific.cpp
+++ b/src/profile/ui/gamespecific.cpp
@@ -22,17 +22,23 @@ void Configure() {
   // Get common gamespecific UI elements
   if (GameSpecificType == +GameSpecificType::CCLCC) {
     UIButtonGuideSprites = GetMemberVector<Sprite>("UIButtonGuideSprites");
+    UIButtonGuideEndDisp = EnsureGetMember<RectF>("UIButtonGuideEndDisp");
   }
   Pop();
 
-  // Configure game-specific subsystems
-  if (GameSpecificType == +GameSpecificType::CHLCC) {
-    Profile::CHLCC::DelusionTrigger::Configure();
-  } else if (GameSpecificType == +GameSpecificType::CC) {
-  } else if (GameSpecificType == +GameSpecificType::CCLCC) {
-    Profile::CCLCC::MapSystem::Configure();
-    Profile::CCLCC::YesNoTrigger::Configure();
-    Profile::CCLCC::DelusionTrigger::Configure();
+  switch (GameSpecificType) {
+    case +GameSpecificType::CHLCC:
+      CHLCC::DelusionTrigger::Configure();
+      break;
+    case +GameSpecificType::CC:
+      break;
+    case +GameSpecificType::CCLCC:
+      CCLCC::MapSystem::Configure();
+      CCLCC::YesNoTrigger::Configure();
+      CCLCC::DelusionTrigger::Configure();
+      break;
+    case +GameSpecificType::None:
+      break;
   }
 }
 
--- a/src/ui/gamespecific.cpp
+++ b/src/ui/gamespecific.cpp
@@ -1,9 +1,14 @@
 #include "gamespecific.h"
+#include "../profile/game.h"
 #include "../profile/vm.h"
 #include "../profile/ui/gamespecific.h"
 #include "../profile/games/cclcc/delusiontrigger.h"
-
 #include "../profile/games/chlcc/delusiontrigger.h"
+
+#include "../games/chlcc/delusiontrigger.h"
+#include "../games/cclcc/delusiontrigger.h"
+#include "../games/cclcc/yesnotrigger.h"
+#include "../games/cclcc/mapsystem.h"
 
 using namespace Impacto::Profile::GameSpecific;
 using namespace Impacto::Profile::ScriptVars;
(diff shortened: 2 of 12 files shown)


KKhanhH:
primary change in the first commit, rest are just like renames. I'm not sure if this API would be more sensible for the gamespecific ui features.

src/profile/ui/gamespecific.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+  // Configure game-specific subsystems
+  if (GameSpecificType == +GameSpecificType::CHLCC) {
+    Profile::CHLCC::DelusionTrigger::Configure();
+  } else if (GameSpecificType == +GameSpecificType::CC) {
+  } else if (GameSpecificType == +GameSpecificType::CCLCC) {
+    Profile::CCLCC::MapSystem::Configure();
+    Profile::CCLCC::YesNoTrigger::Configure();
+    Profile::CCLCC::DelusionTrigger::Configure();
+  }
```
PringlesGang, line 42: These should be `switch` statements

src/ui/gamespecific.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include "../games/cclcc/delusiontrigger.h"
+#include "../games/chlcc/delusiontrigger.h"
+#include "../games/cclcc/yesnotrigger.h"
+#include "../games/cclcc/mapsystem.h"
```
PringlesGang: These should all be moved to the implementation file

src/ui/gamespecific.cpp
```diff
@@ -0,0 +1,108 @@
...
+        std::make_unique<CHLCC::DelusionTrigger>();
+  } else if (Profile::GameSpecific::GameSpecificType ==
+             +GameSpecificType::CCLCC) {
+    Impacto::CCLCC::DelusionTrigger::Implementation =
+        std::make_unique<Impacto::CCLCC::DelusionTrigger>();
+    Impacto::CCLCC::YesNoTrigger::YesNoTriggerPtr =
+        std::make_unique<Impacto::CCLCC::YesNoTrigger>();
+    Impacto::UI::CCLCC::MapSystemCCLCC::Implementation =
+        std::make_unique<Impacto::UI::CCLCC::MapSystemCCLCC>();
+  }
```
PringlesGang: Across this file these should be switch statements, and preferably even call into static functions for each game, for easier maintainability

src/ui/gamespecific.cpp
```diff
@@ -0,0 +1,108 @@
...
+  } else if (Profile::GameSpecific::GameSpecificType ==
+             +GameSpecificType::CCLCC) {
+    if (ScrWork[SW_MAP_PRI] == static_cast<int>(layer) &&
+        ScrWork[SW_MAP_ALPHA]) {
+      Impacto::UI::CCLCC::MapSystemCCLCC::Implementation->Render();
+    }
+
+    if (static_cast<uint32_t>(ScrWork[SW_DELUSION_PRI]) == layer)
+      Impacto::CCLCC::DelusionTrigger::Implementation->Render();
+    if (Impacto::CCLCC::YesNoTrigger::YesNoTriggerPtr &&
```
PringlesGang: I think at this point we can just make it a requirement for CCLCC to have the Yes/No Trigger implemented (not to mention that `Init` already necessarily instantiates the pointer), so I don't think there's a need to check against its existence anymore

src/ui/gamespecific.cpp
```diff
@@ -0,0 +1,108 @@
...
+        static_cast<uint32_t>(ScrWork[SW_YESNO_PRI]) == layer) {
+      Impacto::CCLCC::YesNoTrigger::YesNoTriggerPtr->Render();
+    }
+  }
+}
+
+void RenderCCButtonGuide() {
+  if (!GetFlag(SF_UIHIDDEN) && !GetFlag(2487)) {
```
PringlesGang, line 79: Do we know what 2487 does?

src/ui/gamespecific.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+        CornersQuad dest = {
+            glm::vec2{guideXWidth, 988.0f},        // top left
+            glm::vec2{guideXWidth, 1042.0f},       // bottom left
+            glm::vec2{guideXWidth + 60, 988.0f},   // top right
+            glm::vec2{guideXWidth + 60, 1042.0f},  // bottom right
+        };
+        Renderer->DrawSprite(guideSprite2, dest, glm::mat4(1.0f), tints,
+                             glm::vec3(0.0f), false);
+      }
```
PringlesGang, line 103: Should this be based on the design dimensions in any way? Or is that outta scope for this PR?

src/ui/gamespecific.cpp
KKhanhH: Don't see it being set or anything anywhere, i assume it's like ui elements related since it gets checked in saveicon and stuff too, but flags are harder to find

Issue #219, opened 2025-08-06
Add window icons to SDL_Window from profiles


Enorovan:
So that we can custom it per game instead of having the default one

PR #300, opened 2025-10-29
Add support to window icons to SDL_Window from profiles.

dendi2008:
This PR will add support to window icons to SDL_Window from profiles. (issue #219). You will need to add icons files (standard .png or .jpg and etc.), because I only added support to them.

Example:
{96D934EF-CD8A-4BEA-A14B-DAECB7CC5955}

### DIFF
--- a/src/profile/game.cpp
+++ b/src/profile/game.cpp
@@ -16,6 +16,7 @@ void LoadGameFromLua() {
   LayerCount = EnsureGetMember<int>("LayerCount");
   GameFeatures = EnsureGetMember<int>("GameFeatures");
   WindowName = EnsureGetMember<char const*>("WindowName");
+  WindowIconPath = TryGetMember<std::string>("WindowIconPath");
   DesignWidth = EnsureGetMember<float>("DesignWidth");
   DesignHeight = EnsureGetMember<float>("DesignHeight");
 
--- a/src/renderer/dx9/window.cpp
+++ b/src/renderer/dx9/window.cpp
@@ -95,6 +95,8 @@ void DirectX9Window::Init() {
   ImpLog(LogLevel::Debug, LogChannel::General,
          "Window size (screen coords): {:d} x {:d}\n", WindowWidth,
          WindowHeight);
+
+  SetWindowIcon(SDLWindow);
 }
 
 void DirectX9Window::SetDimensions(int width, int height, int msaa,
--- a/src/renderer/window.cpp
+++ b/src/renderer/window.cpp
@@ -1,0 +1,66 @@
+#include "window.h"
+#include "../io/assetpath.h"
+#include "../profile/game.h"
+#include "../log.h"
+#include <stb_image.h>
+#include <memory>
+#include <vector>
+
+namespace Impacto {
+
+void SetWindowIcon(SDL_Window* window) {
+  if (!Profile::WindowIconPath.has_value()) {
+    return;
+  }
+
+  Io::AssetPath asset;
+  asset.FileName = Profile::WindowIconPath->c_str();
+  Io::Stream* streamPtr;
+  IoError err = asset.Open(&streamPtr);
+  if (err != IoError_OK) {
+    ImpLog(LogLevel::Warning, LogChannel::General,
+           "Could not open window icon file {:s}\n",
+           Profile::WindowIconPath->c_str());
+    return;
+  }
+  std::unique_ptr<Io::Stream> stream(streamPtr);
+  size_t fileSize = stream->Meta.Size;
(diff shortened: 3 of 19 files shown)


src/renderer/dx9/window.cpp
```diff
@@ -95,6 +98,40 @@ void DirectX9Window::Init() {
...
+            SDL_FreeSurface(surface);
+          }
+          stbi_image_free(image);
+        }
+      } else {
+        free(fileData);
+      }
+    }
+  }
 }
```
KKhanhH: Since this is identical on every window, i would suggest moving it to a common file, maybe window.cpp at the base level

src/renderer/dx9/window.cpp
```diff
@@ -95,6 +98,40 @@ void DirectX9Window::Init() {
...
          WindowHeight);
+
+  if (Profile::WindowIconPath && strlen(Profile::WindowIconPath) > 0) {
+    Io::AssetPath asset;
+    asset.FileName = Profile::WindowIconPath;
+    Io::Stream* stream;
+    IoError err = asset.Open(&stream);
+    if (err == IoError_OK) {
+      size_t fileSize = stream->Meta.Size;
+      uint8_t* fileData = (uint8_t*)malloc(fileSize);
```
KKhanhH: Could you use vector or unique_ptr<uint8_t[]> instead?

src/profile/game.cpp
```diff
@@ -16,6 +16,8 @@ void LoadGameFromLua() {
   LayerCount = EnsureGetMember<int>("LayerCount");
   GameFeatures = EnsureGetMember<int>("GameFeatures");
   WindowName = EnsureGetMember<char const*>("WindowName");
+  WindowIconPath = nullptr;
+  TryGetMember<char const*>("WindowIconPath", WindowIconPath);
```
PringlesGang: Please make `WindowIconPath` a `std::optional<std::string>` and do
`WindowIconPath = TryGetMember<std::string>("WindowIconPath");`
Makes the conditional check a bit cleaner

dendi2008: Done

src/renderer/dx9/window.cpp
dendi2008: Done.

src/renderer/dx9/window.cpp
dendi2008: Done.

src/renderer/window.cpp
```diff
@@ -0,0 +1,44 @@
...
+            image, width, height, 32, width * 4, 0x000000FF, 0x0000FF00,
+            0x00FF0000, 0xFF000000);
+        if (surface) {
+          SDL_SetWindowIcon(window, surface);
+          SDL_FreeSurface(surface);
+        }
+        stbi_image_free(image);
+      }
+    }
+  }
```
PringlesGang: This could probably use some logging in case something goes wrong (`err != IoError_OK`, `image == nullptr`, `surface == nullptr`). Other than that, looks really good!

dendi2008: I added logging for cases you wrote. I also added logging for a read failure case.

src/renderer/window.cpp
```diff
@@ -0,0 +1,65 @@
...
+  Io::AssetPath asset;
+  asset.FileName = Profile::WindowIconPath->c_str();
+  Io::Stream* stream;
+  IoError err = asset.Open(&stream);
+  if (err != IoError_OK) {
+    ImpLog(LogLevel::Warning, LogChannel::General,
+           "Could not open window icon file {:s}\n",
+           Profile::WindowIconPath->c_str());
+    return;
+  }
```
PringlesGang: You're not deleting the stream here in the error case, so either delete it before the return or change stream to a unique pointer
Apart from that, all good!

PringlesGang (changes requested):

src/renderer/window.cpp
dendi2008: Fixed. I changed stream to a unique pointer.

PringlesGang: You're only creating the unique pointer _after_ the error case would've already returned, so in that case you'd still cause a memory leak

KKhanhH: Actually on error cases stream pointers are already deleted in their stream implementations so it should be fine.

KKhanhH (approved):

src/renderer/window.cpp
PringlesGang: Oh cool nevermind then

PringlesGang (approved):

Issue #220, opened 2025-08-06
Add custom arrow and pointer icons per game


Enorovan:
And implement pointer logic when hovering clickable things

(You can test it by grabbing the C;C ones from C;C Steam using ResourceHacker)

PR #302, opened 2025-10-31
Add custom arrow and pointer icons per game support.

dendi2008:
This PR will add custom arrow and pointer icons per game support and will implement pointer logic when hovering clickable things. (issue #220) Custom arrow and pointer icons will use .png, or .jpg and etc. files formats.

Examples:
{3898A843-F389-4FC5-BAEE-5B4D550FF74F}
{1A1587DD-32FD-4598-AF54-7C072DCAC93F}

### DIFF
--- a/src/renderer/window.cpp
+++ b/src/renderer/window.cpp
@@ -5,6 +5,8 @@
 #include <stb_image.h>
 #include <memory>
 #include <vector>
+#include <optional>
+#include "../inputsystem.h"
 
 namespace Impacto {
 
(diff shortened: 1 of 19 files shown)


src/ui/widgets/backlogentry.cpp
```diff
@@ -81,6 +82,7 @@ void BacklogEntry::UpdateInput(float dt) {
           entryHoverBounds.ContainsPoint(Input::CurMousePos) &&
           HoverBounds.Y <= Bounds.Y &&
           (Bounds.Y + Bounds.Height) <= (HoverBounds.Y + HoverBounds.Height);
+      if (Hovered) RequestPointerCursor();
     } else if (Input::CurrentInputDevice == Input::Device::Touch &&
                Input::TouchIsDown[0] &&
                Input::PrevMousePos != Input::CurTouchPos) {
```
PringlesGang, line 86: This one (and the ones in the other widgets) are superfluous, as these would all trigger the `RequestPointerCursor` call further down as well

dendi2008:
> The implementation looks fine, but it's not very extendable to other kinds of cursors (text selection, loading, moving, etc.). Maybe you could adjust the API slightly to work based on enums, where for now you just have `Default` and `Pointer`?
> 
> Also, I'm not sure if this would work, but maybe you could move the `RequestPointerCursor` stuff into `Widget::Update`? Would reduce the technical debt for having to consider the pointer each time

Done

src/renderer/window.cpp
```diff
@@ -63,4 +64,109 @@ void SetWindowIcon(SDL_Window* window) {
...
+           SDL_GetError());
+  }
+
+  SDL_FreeSurface(surface);
+  stbi_image_free(image);
+  return cursor;
+}
+
+void InitCursors() {
+  SystemArrowCursor = SDL_CreateSystemCursor(SDL_SYSTEM_CURSOR_ARROW);
```
KKhanhH: Is this being used for anything?

KKhanhH: if it's just being used for storing the default cursor then I think SDL_GetDefaultCursor would be better

src/renderer/window.cpp
```diff
@@ -63,4 +64,109 @@ void SetWindowIcon(SDL_Window* window) {
...
+    CurrentCursor = CursorArrow;
+  }
+}
+
+void RequestCursor(CursorType type) { RequestedCursorType = type; }
+
+void ApplyCursorForFrame() {
+  SDL_Cursor* desired = CursorArrow;
+  if (Input::CurrentInputDevice == Input::Device::Mouse) {
+    if (RequestedCursorType == CursorType::Pointer && CursorPointer) {
```
KKhanhH: I would make Request an optional or add a none option so that you could reset request after setting cursor, so that you aren't setting the cursor every frame.

dendi2008: Done.

src/renderer/window.cpp
dendi2008: > if it's just being used for storing the default cursor then I think SDL_GetDefaultCursor would be better

Yeah, that's would be better. Thanks.

src/renderer/window.cpp
KKhanhH: I guess you might still have to store the current type or something, since it'll still get called by update functions actually

dendi2008: > I guess you might still have to store the current type or something, since it'll still get called by update functions actually

Should I revert my changes?

KKhanhH: I think it's probably fine, just add a current and don't assign to request if current matches. You'd probably also want to assign current when input changes as well.

dendi2008: > I think it's probably fine, just add a current and don't assign to request if current matches. You'd probably also want to assign current when input changes as well.

Done. If I understood you correctly.

src/renderer/window.cpp
```diff
@@ -63,4 +65,146 @@ void SetWindowIcon(SDL_Window* window) {
...
+
+void ApplyCursorForFrame() {
+  SDL_Cursor* actualCursor = SDL_GetCursor();
+
+  if (Input::CurrentInputDevice != PreviousInputDevice) {
+    CurrentCursor = actualCursor;
+    PreviousInputDevice = Input::CurrentInputDevice;
+  }
+
+  if (Input::CurrentInputDevice == Input::Device::Mouse) {
```
PringlesGang: I think it should also preserve the custom cursor if the current input device is a keyboard

dendi2008: > I think it should also preserve the custom cursor if the current input device is a keyboard

I can change that. But in Mages games if current input device is a keyboard (for example, if you press W or S to change selected option in C;C main menu) the custom cursor changes to the system (OS) cursor.

PringlesGang: I still think that's weird behaviour, so I think it's best we diverge from the MAGES. engine here
@KKhanhH @Enorovan thoughts?

Enorovan: I agree

PringlesGang: Mouse and keyboard, yeah. Ideally I'd say on any other input device, the cursor should just become invisible, but you can just do whatever for now

KKhanhH: I think that's fine. Cursor doesn't stay like that when you alt tab or something does it?

dendi2008: > I think that's fine. Cursor doesn't stay like that when you alt tab or something does it?

Well, the system cursor changes to the custom cursor when the cursor is inside the game (impacto) window. If you alt tab from this window the custom cursor changes back to the system cursor.

dendi2008: So, I'm making that when the user uses another input device, the cursor just disappears, right?

PringlesGang: We discussed it a bit more internally. We think it's best the custom cursor just kinda stays regardless of what the current input is, so long as the cursor is only a pointer if the current input device is a mouse

dendi2008: > the custom cursor just kinda stays regardless of what the current input is, so long as the cursor is only a pointer if the current input device is a mouse

For example, when I hovered over a button in the CCLCC menu with the mouse, the custom cursor changed from an arrow to a pointer. Then, I changed the selected button with the keyboard. Should the pointer cursor change to an arrow?

KKhanhH: Yeah, it should to indicate that hover state isnt being checked right now

dendi2008: Done.

KKhanhH (approved):

PR #303, opened 2025-11-03
[CCLCC][CHLCC] Bug fixes.

dendi2008:
This PR will:

- Fix bug where clicking a button with mouse after returning from a submenu would activate it (button, for example, Load button in CCLCC Title Menu and etc.) even if the mouse wasn't hovering button.

### DIFF
--- a/src/games/chlcc/systemmenu.cpp
+++ b/src/games/chlcc/systemmenu.cpp
@@ -18,6 +18,7 @@ using namespace Impacto::Profile::SystemMenu;
 using namespace Impacto::UI::Widgets::CHLCC;
 
 void SystemMenu::MenuButtonOnClick(Widgets::Button* target) {
+  target->Hovered = false;
   if (static_cast<SystemMenuEntryButton*>(target)->IsLocked) {
     // Yep, that's similar to how it's done in the binary
     // Binary checks for button state and if it's locked, PADone is modified
--- a/src/games/cc/titlemenu.cpp
+++ b/src/games/cc/titlemenu.cpp
@@ -34,17 +34,20 @@ using namespace Impacto::Vm::Interface;
 using namespace Impacto::UI::Widgets::CC;
 
 void TitleMenu::MenuButtonOnClick(Widgets::Button* target) {
+  target->Hovered = false;
   ScrWork[SW_TITLECUR1] = target->Id;
   SetFlag(SF_TITLEEND, 1);
   AllowsScriptInput = true;
 }
 
 void TitleMenu::ContinueButtonOnClick(Widgets::Button* target) {
+  target->Hovered = false;
   CurrentSubMenu = ContinueItems;
   AllowsScriptInput = false;
 }
 
 void TitleMenu::ExtraButtonOnClick(Widgets::Button* target) {
+  target->Hovered = false;
   CurrentSubMenu = ExtraItems;
   AllowsScriptInput = false;
 }
--- a/src/games/cclcc/savemenu.cpp
+++ b/src/games/cclcc/savemenu.cpp
@@ -24,6 +24,7 @@ using namespace Impacto::UI::Widgets::CCLCC;
 Widget* EntryGrid[Pages][RowsPerPage][EntriesPerRow];
 
 void SaveMenu::MenuButtonOnClick(Widgets::Button* target) {
+  target->Hovered = false;
   Impacto::SaveSystem::SaveType saveType =
       *ActiveMenuType == +SaveMenuPageType::QuickLoad
           ? SaveSystem::SaveType::Quick
--- a/src/games/cclcc/systemmenu.cpp
+++ b/src/games/cclcc/systemmenu.cpp
@@ -20,6 +20,7 @@ using namespace Impacto::Profile::SystemMenu;
 using namespace Impacto::UI::Widgets::CCLCC;
 
 void SystemMenu::MenuButtonOnClick(Widgets::Button* target) {
+  target->Hovered = false;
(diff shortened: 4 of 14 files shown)


KKhanhH:
I think setting hover to false after a click or after entering a submenu would be a lot better here, these changes are pretty computationally expensive and affects a LOT of buttons

dendi2008:
> I think setting hover to false after a click or after entering a submenu would be a lot better here, these changes are pretty computationally expensive and affects a LOT of buttons

Yeah, that's a lot better. Done.

src/games/chlcc/systemmenu.cpp
```diff
@@ -164,7 +165,27 @@ void SystemMenu::Update(float dt) {
     UpdateSmoothSelection(dt);
     MenuLoop.Update(dt);
     UpdateRunningSelectedLabel(dt);
+    Widget* prevFocusedElement = CurrentlyFocusedElement;
     MainItems->UpdateInput(dt);
+    if (PADinputMouseWentDown & PAD1A) {
```
KKhanhH: I'm not so sure this section should be necessary, intended behavior of button by default is already to only allow click callback only when hovering. Maybe it's triggering because of double update calls?

dendi2008: Well, I think buttons triggering because:
```39:44:src/ui/widgets/button.cpp
if (OnClickHandler && HasFocus &&
    ((Hovered &&
      Vm::Interface::PADinputMouseWentDown & Vm::Interface::PAD1A) ||
     (Vm::Interface::PADinputButtonWentDown & Vm::Interface::PAD1A))) {
  OnClickHandler(this);
}
```
(in button.cpp)
```71:86:src/ui/widgets/group.cpp
for (const auto& el : Children) {
  if (el->GetType() == WT_NORMAL) {
    el->UpdateInput(dt);
    if (el->Enabled && el->Hovered && el->Bounds.Intersects(HoverBounds) &&
        (Input::CurrentInputDevice == Input::Device::Mouse ||
         Input::CurrentInputDevice == Input::Device::Touch)) {
      if (MenuContext->CurrentlyFocusedElement &&
          el != MenuContext->CurrentlyFocusedElement)
        MenuContext->CurrentlyFocusedElement->HasFocus = false;
      el->HasFocus = true;
      MenuContext->CurrentlyFocusedElement = el;
    }
  }
}
```
(in group.cpp)

Group::UpdateInput sets HasFocus to true when a button is hovered. HasFocus stays true even when you unhover button. And in CCLCC(or CHLCC) Backlog button already hovered when user entering System Menu. So, HasFocus stays true and button is hovered and if user will click using mouse it should trigger button. (I also noticed that clicking on an empty space could open Backlog menu, even when Backlog button was locked. The same for Quick Save button.)

KKhanhH: HasFocus staying true is correct, but hovered being true when over an empty space sounds like a bug somewhere else.
We should be checking why hover is true when on an empty space, if its not updating or if the bounds are incorrect or something. Also if its triggering while locked then we should just put that check in the onclick handler, same with setting hover to false.

src/ui/widgets/group.cpp
```diff
@@ -83,6 +84,22 @@ void Group::UpdateInput(float dt) {
       }
     }
   }
+
+  if ((Input::CurrentInputDevice == Input::Device::Mouse ||
```
KKhanhH: Let's not do this, this probably breaks a bit of stuff since clicks arent always handled by a singular group.

src/games/chlcc/systemmenu.cpp
dendi2008: > HasFocus staying true is correct, but hovered being true when over an empty space sounds like a bug somewhere else. We should be checking why hover is true when on an empty space, if its not updating or if the bounds are incorrect or something. Also if its triggering while locked then we should just put that check in the onclick handler, same with setting hover to false.

I also noticed that when I clicked on an empty space, I entered the last menu (for example, I entered in Load Menu, then returned to the System Menu, clicked on an empty space, and I entered again in Load Menu. The same for other menus).

KKhanhH: Its possible the focused flag isnt correct. Try first try removing the group updateinput calls and leave just update calls (update calls updateinput and guards it with flags)

dendi2008: > Its possible the focused flag isnt correct. Try first try removing the group updateinput calls and leave just update calls (update calls updateinput and guards it with flags)

It's seems that problem is not in focused flag. I even completly removed updateinput group. This bug still was there. I think that this bug needs more investigation. I think it's better, for now, to remove this fix from PR and create Issue or revert to my original fix (in CCLCC and CHLCC System Menu files.), because it will take a lot of time to investigate this bug.

Gunoshozo:
> I think setting hover to false after a click or after entering a submenu would be a lot better here

Maybe this change also should be applied to the `Impacto::UI::Widgets::ClickArea`?

src/games/chlcc/systemmenu.cpp
KKhanhH: Yeah just create an issue for it, I can take a look at it. I think I've seen something similar before

dendi2008: > Yeah just create an issue for it, I can take a look at it. I think I've seen something similar before

I created an issue for this bug and removed my fix for this bug from my commit. (Also I checked again and when I clicked on an empty space and Backlog button was locked I did not enter in Backlog Menu. But for Quick Save button it's still the same).

PR #306, opened 2025-11-06
Yes/No Trigger Improvements

Gunoshozo:
Changes:
- Blur & Masking for bg
- Mouse support
- Star uses Animation class for rotation
- Minor refactor & fixes to YesNoData profile initialization
- Reorder fields and set u8 as underlying type for enums for better memory alignment & density of YesNoTrigger

Dev Proof:
[vita3k](https://www.youtube.com/watch?v=7u1AY8IcR98)
[impacto](https://www.youtube.com/watch?v=DGO3FiNQXUI)

- Mask in RenderDoc: <a href="https://github.com/user-attachments/assets/1c6d046e-6806-465d-8ea2-e26ded1fd475" target="_blank">
  
</a>

- Regarding `-46.0f` Y offset. It technically accurate to the console bahaviour but it causes visual artifacts at the start of the zoom and it is most likely a bug. I have rescaled value range so it will be 0 at the start and returns to full -46.0f at the full zoom. Here's couple examples from ps4 and vita:
<a href="https://github.com/user-attachments/assets/24e92562-5853-49b1-bebc-9c7c2ab235b9" target="_blank">
  
</a>
<a href="https://github.com/user-attachments/assets/12fa2018-4910-4b53-ab6b-20b3d2808a90" target="_blank">
  
</a>

I hope that the amount of comments is not super excessive


src/games/cclcc/yesnotrigger.h
```diff
@@ -38,23 +39,28 @@ class YesNoTrigger {
   int Load(uint8_t* data);
   int Save(uint8_t* data);
 
-  YesNoState State = YesNoState::None;
+  Sprite ActiveBackground;
+  Sprite ActiveBlur;
+  Animation StarAnimation;
+  YesNoPositions* DispPosArr = nullptr;
+  Widgets::ClickArea* YesClickArea;
```
KKhanhH: These can probably just be value objects

src/games/cclcc/yesnotrigger.cpp
```diff
@@ -219,24 +212,66 @@ void YesNoTrigger::Render() {
...
-  float alpha = static_cast<float>(ScrWork[6433] << 3);
-  if (alpha > 255) {
-    alpha = 255;
-  } else if (alpha < 0) {
-    alpha = 0;
-  }
+  // TODO: refactor to be normalized to the design space
+  // 3840.0 and 2160.0 are coupled to PS4 sprite dimensions, probably will cause
+  // issues for psVita resources
+  const glm::vec2 BgSize = {3840.0f, 2160.0f};
```
KKhanhH: you could just use designwidth and designheight of the background sheet

src/games/cclcc/yesnotrigger.cpp
```diff
@@ -219,24 +212,66 @@ void YesNoTrigger::Render() {
...
+  float bgWidth = BgSize.x / BgSpriteScale;
```
KKhanhH: might as well make variables const while you're at it

KKhanhH (changes requested):

src/games/cclcc/yesnotrigger.h
Gunoshozo: Done

src/games/cclcc/yesnotrigger.cpp
Gunoshozo: Done

src/games/cclcc/yesnotrigger.cpp
Gunoshozo: Damn, totaly forgot that you can access those directly
Done

src/games/cclcc/yesnotrigger.cpp
```diff
@@ -148,6 +148,17 @@ void YesNoTrigger::Update(float dt) {
...
         DispSel = true;
+        if (Input::CurrentInputDevice == Input::Device::Mouse) {
+          if (YesClickArea.Hovered && Selection != YesNoSelect::YES &&
+              !IsChoiceBlocked(YesNoSelect::YES)) {
+            Selection = YesNoSelect::YES;
+            Audio::PlayInGroup(Audio::ACG_SE, "sysse", 1, false, 0.0f);
+          } else if (NoClickArea.Hovered && Selection != YesNoSelect::NO) {
+            Selection = YesNoSelect::NO;
+            Audio::PlayInGroup(Audio::ACG_SE, "sysse", 1, false, 0.0f);
+          }
```
PringlesGang: Does NO not need an `IsChoiceBlocked` check?

src/games/cclcc/yesnotrigger.cpp
```diff
@@ -275,30 +301,34 @@ void YesNoTrigger::Render() {
...
-                         chipTint);
+
+    YesClickArea.Bounds = yesChipDest;
+    NoClickArea.Bounds = noChipDest;
+    Renderer->DrawSprite(*ActiveYesChip, yesChipDest, chipTint);
+    Renderer->DrawSprite(*ActiveNoChip, noChipDest, chipTint);
   }
-  glm::vec4 maskTint = glm::vec4(1.0f, 1.0f, 1.0f, 160 / 256.0f);
-  Renderer->DrawSprite(YesNoBgMask, Rect(0, 0, 1920, 1080), maskTint);
+  const glm::vec4 maskTint = glm::vec4(1.0f, 1.0f, 1.0f, 160 / 256.0f);
```
PringlesGang: This can be constexpr

src/games/cclcc/yesnotrigger.cpp
```diff
@@ -350,6 +407,37 @@ void YesNoTrigger::Hide() {
...
+
+void YesNoTrigger::AreaClick(Widgets::ClickArea* area) {
+  if (IsChoiceBlocked(static_cast<YesNoSelect>(area->Id))) {
+    Audio::PlayInGroup(Audio::ACG_SE, "sysse", 4, false, 0.0f);
+    return;
+  }
+  ChooseSelected();
+  area->Hovered = false;
+}
+
```
PringlesGang, line 419: No need for `YesNoTrigger::IsChoiceBlocked` to be inline. It's already only ever in a single translation unit

src/games/cclcc/yesnotrigger.cpp
```diff
@@ -148,6 +148,21 @@ void YesNoTrigger::Update(float dt) {
     case YesNoState::MainInput:
       if (AllowInput) {
         DispSel = true;
+        if (Input::CurrentInputDevice == Input::Device::Mouse) {
+          if (NoClickArea.Hovered) {
+            // folded that way on purpose,if two areas are both hovered
+            // they will be switching constantly
```
Gunoshozo, line 154: otherwise this

[oops.webm](https://github.com/user-attachments/assets/306a0cc4-75ff-492f-a45c-56457fe18061)

src/games/cclcc/yesnotrigger.h
```diff
@@ -38,23 +39,28 @@ class YesNoTrigger {
   int Load(uint8_t* data);
   int Save(uint8_t* data);
 
-  YesNoState State = YesNoState::None;
+  Sprite ActiveBackground;
+  Sprite ActiveBlur;
+  Animation StarAnimation;
+  YesNoPositions* DispPosArr = nullptr;
```
PringlesGang: Could you make this a span?

src/games/cclcc/yesnotrigger.cpp
```diff
@@ -219,42 +216,75 @@ void YesNoTrigger::Render() {
...
+      glm::vec2((1.0f - BgSpriteScale) * ActiveBlur.Bounds.Width,
+                (1.0f - BgSpriteScale) * ActiveBlur.Bounds.Height);
+  // apply scaled BgSpritePos offset
+  blurTranslation -= BgSpritePos * BgSpriteScale * 0.5f;
+  // correction for "- 46.0f" from ActiveBackground
+  blurTranslation.y += 46.0f * (BgSpriteScale - 1.0f) * 0.75f;
+
+  const glm::mat4 blurTransform = TransformationMatrix(
+      glm::vec3(0.0f), blurScale, glm::vec3(0.0f), glm::quat(),
+      {blurTranslation.x, blurTranslation.y, 0});
```
PringlesGang: Why aren't you using the 2D overload?

src/games/cclcc/yesnotrigger.cpp
```diff
@@ -219,42 +216,75 @@ void YesNoTrigger::Render() {
...
            static_cast<int>(BgSpritePos.y + bgYOffset),
            static_cast<int>(bgWidth), static_cast<int>(bgHeight));
 
-  Renderer->DrawSprite(ActiveBackground, Rect(0, 0, 1920, 1080), bgtint);
+  Renderer->DrawSprite(ActiveBackground,
+                       RectF(0, 0, BgSize.x * 0.5f, BgSize.y * 0.5f), bgtint);
+
+  // blur's sprite is 2 times smaller than the bg
+  const glm::vec3 blurScale =
+      glm::vec3(2 * BgSpriteScale, 2 * BgSpriteScale, 1.0f);
```
PringlesGang: Use the `ActiveBackground` and `ActiveBlur`'s scaled bounds to compute this dynamically

src/games/cclcc/yesnotrigger.cpp
```diff
@@ -219,42 +216,75 @@ void YesNoTrigger::Render() {
...
-  float bgXOffset = (3840.0f - bgWidth) / 2.0f;
-  float bgYOffset = (2160.0f - bgHeight) / 2.0f - 46.0f;
-  float alpha = static_cast<float>(ScrWork[6433] << 3);
-  if (alpha > 255) {
-    alpha = 255;
-  } else if (alpha < 0) {
-    alpha = 0;
-  }
-  glm::vec4 bgtint = glm::vec4(1.0f, 1.0f, 1.0f, alpha / 255.0f);
+  const glm::vec2 BgSize = ActiveBackground.Sheet.GetDimensions();
```
PringlesGang: Local variables should be lower case

src/games/cclcc/yesnotrigger.cpp
KKhanhH: Choice block is for the route entry into the <FINAL_ROUTE> so it shouldn't need to I believe

src/games/cclcc/yesnotrigger.cpp
KKhanhH: Yeah, don't define member functions inline (or pretty in general unless in header free function). If we ever need to reference it elsewhere it breaks things, and compiler/linker can optimize it

src/games/cclcc/yesnotrigger.cpp
Gunoshozo: 2x is a mistake in the comment, it's 4x difference actually.
ActiveBg is 2x of target resolution
BlurBg is 0.5x of target resolution
Changed it to the 
```cpp
BgSpriteScale * 0.5f *ActiveBackground.Sheet.GetDimensions() / ActiveBlur.Bounds.GetSize()
```
so a blur will get scaled to the target resoltion at BgSpriteScale == 1 and so on. 
Also using design dimensions since ActiveBackground's scaled bounds are influenced by transformations and regular bounds for ActiveBlur since it's scaled bounds are not influenced by transformations.

src/games/cclcc/yesnotrigger.cpp
Gunoshozo: Done

src/games/cclcc/yesnotrigger.h
Gunoshozo: Done, also corrected sizeof(YesNoPositions) -> sizeof(void*) in `YesNoTrigger::Save` and `YesNoTrigger::Load`

src/games/cclcc/yesnotrigger.cpp
Gunoshozo: Done

src/games/cclcc/yesnotrigger.cpp
Gunoshozo: Done, I can't keep falling into the the same trap

src/games/cclcc/yesnotrigger.cpp
Gunoshozo: Done

Issue #270, opened 2025-10-10
Stop lip movements when character changes


PringlesGang:
When advancing a voiced line into a non-voiced line with a different character on screen, the lips of the new character move as if they were the one who spoke the prior line

Issue #263, opened 2025-10-10
Fix character blinking (and lip movements?)


PringlesGang:
They look and move more jagged than they should

PR #311, opened 2025-11-08
Eye/lip fix

PringlesGang:
- [x] Fixes `InstMes` characterId and animationId mixup
- [x] Fixes character blinking and lip movement animations (#263)
- [x] Fixes lip movement continuing for the next character when character changes (#270)


src/vm/inst_dialogue.cpp
```diff
@@ -210,7 +211,13 @@ VmInstruction(InstMes) {
   thread->IpOffset = oldIp;
   UI::BacklogMenuPtr->AddMessage(
       {.ScriptBufferId = thread->ScriptBufferId, .IpOffset = line}, audioId,
-      characterId);
+      acted ? animationId : characterId);
+
+  if (!(type & 0b1000)) {
+    SetFlag(1213 + thread->DialoguePageId, true);
```
KKhanhH: do you know what 1213 is

PringlesGang: I have a guess, but nothing concrete

KKhanhH: Oh i think it's SF_CHAANIME

PringlesGang: Where'd you get that name?

KKhanhH: cross referencing scriptdbg from memoff and scriptdbg from magesscripttools

PR #312, opened 2025-11-09
Fix submenus from system menu holding before closing

PringlesGang:
When closing a submenu (e.g. backlog, config, tips, save, load) opened from the system menu in C;H LCC, the game holds for a few seconds before actually closing. This PR fixes that

When opening a menu from the system menu, `SW_SYSMENUCT` is already at 10000, because the system menu is fully transitioned in. Instead, the game then uses `SW_SYSTEMMENUCHG` to mark the progress of the transition animation

When opening, it sets `SW_SYSTEMMENUCHG` to 0, and increments it in a loop until it reaches 64, at which point it will move to that menu's main instruction
When closing, it will then decrement it again in a loop until it reaches 0

When not opening/closing through the system menu, it will simply use `SW_SYSMENUCT`, leaving `SW_SYSTEMMENUCHG` untouched at 0

### DIFF
--- a/src/games/chlcc/backlogmenu.cpp
+++ b/src/games/chlcc/backlogmenu.cpp
@@ -150,21 +150,25 @@ void BacklogMenu::Render() {
 }
 
 void BacklogMenu::Update(float dt) {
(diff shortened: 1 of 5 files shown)


KKhanhH (approved):
Gonna trust that you tested it and tried spamming open/close with no issues

PR #317, opened 2025-11-12
Add support for opening the in-game Backlog menu by scrolling up and close it when fully scrolled down, and scrolling down more.

dendi2008:
This PR will add support for opening the in-game Backlog menu by scrolling up and close it when fully scrolled down, and scrolling down more (with mouse). It will also close Backlog if scrolling is unavaible (for example, when Backlog has only 1-2 lines.)

### DIFF
--- a/src/profile/game.cpp
+++ b/src/profile/game.cpp
@@ -27,6 +27,8 @@ void LoadGameFromLua() {
   ResolutionHeight = EnsureGetMember<int>("ResolutionHeight");
   Fullscreen = EnsureGetMember<bool>("Fullscreen");
   Subtitles = EnsureGetMember<char const*>("Subtitles");
+  CloseBacklogWhenReachedEnd =
+      TryGetMember<bool>("CloseBacklogWhenReachedEnd").value_or(true);
 
   bool res = TryGetMember<bool>("LayFileBigEndian", LayFileBigEndian);
   if (!res) LayFileBigEndian = false;
--- a/src/ui/backlogmenu.cpp
+++ b/src/ui/backlogmenu.cpp
@@ -1,5 +1,7 @@
 #include "backlogmenu.h"
 
+#include <limits>
+
 #include "ui.h"
 #include "../profile/game.h"
 #include "../profile/vm.h"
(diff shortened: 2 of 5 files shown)


Enorovan:
> > Could this be the occasion to implement in the menus that they close when fully scrolled down, and you scroll down more?
> 
> I can implement this. But are there any other menus besides Backlog that should close when fully scrolled down and you scroll down more?

Nope, though maybe it should be in the BacklogMenu base class.

dendi2008:
> Should probably be a variable in `config.lua` tho, as we also a have a toggle for that in the launcher of a few of our patches (and I personally don't like it :blini:)

Should I also implement this or only implement in the Backlog menu that meny close when fully scrolled down, and you scroll down more (in BacklogMenu base class)?

dendi2008:
> Yeah I'd say just add a boolean to the profile for the down scrolling, and then implement it on the base class of the backlog menu You could also do it in another PR if you'd prefer that

Done.

src/profile/game.cpp
```diff
@@ -27,6 +27,8 @@ void LoadGameFromLua() {
   ResolutionHeight = EnsureGetMember<int>("ResolutionHeight");
   Fullscreen = EnsureGetMember<bool>("Fullscreen");
   Subtitles = EnsureGetMember<char const*>("Subtitles");
+  CloseBacklogWhenReachedEnd =
+      EnsureGetMember<bool>("CloseBacklogWhenReachedEnd");
```
PringlesGang: I'd do
```cpp
CloseBacklogWhenReachedEnd = TryGetMember<bool>("CloseBacklogWhenReachedEnd").value_or(true);
```
to make it an optional profile setting, with a default

src/ui/backlogmenu.cpp
```diff
@@ -205,6 +205,20 @@ void BacklogMenu::UpdateInput(float dt) {
...
   UpdateScrollingInput(dt);
+
+  static bool AtBottomPrev = false;
+
+  if (!(State == Shown && IsFocused)) {
+    AtBottomPrev = false;
+  } else if (Profile::CloseBacklogWhenReachedEnd) {
+    const float epsilon = 0.5f;
+    const bool atBottomNow = !MainScrollbar->Enabled ||
+                             (PageY <= (MainScrollbar->EndValue + epsilon));
```
PringlesGang, line 216: `std::numeric_limits<float>::epsilon()`

PringlesGang (dismissed):

src/ui/backlogmenu.cpp
dendi2008: Done.

src/profile/game.cpp
dendi2008: Done.

src/ui/backlogmenu.cpp
```diff
@@ -205,6 +207,21 @@ void BacklogMenu::UpdateInput(float dt) {
   MainScrollbar->UpdateInput(dt);
   UpdatePageUpDownInput(dt);
   UpdateScrollingInput(dt);
+
+  static bool AtBottomPrev = false;
```
KKhanhH: Practically it doesn't really make a difference since it's a class that only exists once, but logically it's part of the menu's state so I think making it a private member variable makes more sense

dendi2008: > making it a private member variable makes more sense

Done.

PR #318, opened 2025-11-14
Chlcc/delusions

KKhanhH:
Adds the wavey delusion trigger text for CHLCC

### DIFF
--- a/src/games/chlcc/delusiontrigger.cpp
+++ b/src/games/chlcc/delusiontrigger.cpp
@@ -322,9 +322,6 @@ void DelusionTrigger::UpdateShown(float dt) {
 }
 
 void DelusionTrigger::Update(float dt) {
-  if (!GetFlag(SF_DELUSIONACTIVE) && State == Shown) {
-    Reset();
-  }
   if (State == Showing) {
     UpdateShowing(dt);
   } else if (State == Hiding) {
@@ -494,7 +491,7 @@ void DelusionTextSystem::ScrollLine(size_t lineIdx) {
   line[MaxCharsPerLine - 1] = nullptr;
 
   // Find the last non-empty glyph position, fill after it
-  int lastGlyph = MaxCharsPerLine;
+  int lastGlyph = MaxCharsPerLine - 1;
   for (; lastGlyph >= 0; lastGlyph--) {
     if (line[lastGlyph] != nullptr) break;
   }
--- a/src/ui/gamespecific.cpp
+++ b/src/ui/gamespecific.cpp
@@ -20,7 +20,7 @@ namespace GameSpecific {
 void Init() {
   switch (Profile::GameSpecific::GameSpecificType) {
     case +GameSpecificType::CHLCC: {
-      CHLCC::DelusionTrigger::GetInstance();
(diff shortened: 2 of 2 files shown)


src/games/chlcc/delusiontrigger.cpp
```diff
@@ -336,26 +367,230 @@ void DelusionTrigger::Render() {
...
+  auto& line = GlyphLines[lineIdx];
+
+  std::shift_left(line.begin(), line.end(), 1);
+  line[MaxCharsPerLine - 1] = nullptr;
+
+  // Find the last non-empty glyph position
+  int lastGlyph = -1;
+  for (int i = 0; i < MaxCharsPerLine; ++i) {
+    if (line[i] != nullptr) lastGlyph = i;
+  }
```
PringlesGang: I think this loop could be better written going backwards
```cpp
int lastGlyph = MaxCharsPerLine;
for (; lastGlyph >= 0; lastGlyph--) {
    if (line[lastGlyph] != nullptr) break;
}
```

src/games/chlcc/delusiontrigger.cpp
```diff
@@ -336,26 +367,230 @@ void DelusionTrigger::Render() {
...
+}
+
+void DelusionTextSystem::Update(float dt) {
+  DelusionTextFade.Update(dt);
+  if (DelusionTextFade.IsOut()) {
+    if (DelusionSelectedLine) DelusionSelectedLine.reset();
+    return;
+  };
+
+  TextLineXOffset -= DelusionTextXVelocity;
```
PringlesGang: `TextLineXOffset -= DelusionTextXVelocity * dt`?

PringlesGang (changes requested):
Delusion trigger hide animation seems to be borked? It just instantly leaves for me

src/games/chlcc/delusiontrigger.cpp
KKhanhH: You mean dt * 60? Most of the delusion trigger code is frame based right now

PringlesGang: If `DelusionTextXVelocity` is in pixels per frame, then yeah. Regardless, `Update` gets called every game tick, so it should take `dt` into account

src/games/chlcc/delusiontrigger.cpp
```diff
@@ -336,26 +369,230 @@ void DelusionTrigger::Render() {
...
+  auto& line = GlyphLines[lineIdx];
+
+  std::shift_left(line.begin(), line.end(), 1);
+  line[MaxCharsPerLine - 1] = nullptr;
+
+  // Find the last non-empty glyph position, fill after it
+  int lastGlyph = MaxCharsPerLine;
+  for (; lastGlyph >= 0; lastGlyph--) {
+    if (line[lastGlyph] != nullptr) break;
+  }
```
PringlesGang: `int lastGlyph = MaxCharsPerLine - 1;` my b

PringlesGang (changes requested):
The rotation resets when it hides, so it just kinda jaggedly jumps from some rotation to 0 at the start of the transition. Maybe don't reset that?

PR #320, opened 2025-11-22
Select the button in SystemMenu that was last selected before closing System Menu.

dendi2008:
This PR will add:
- When user selects, for example, Load button in SystemMenu, closes SystemMenu (or loads a save file in SystemMenu), and then reopens SystemMenu, the Load button will be selected.

### DIFF
--- a/src/games/cclcc/systemmenu.h
+++ b/src/games/cclcc/systemmenu.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <optional>
 #include <glm/fwd.hpp>
 #include "../../ui/menu.h"
 #include "../../ui/widgets/group.h"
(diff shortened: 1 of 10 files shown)


src/games/cclcc/systemmenu.h
```diff
@@ -36,6 +36,7 @@ class SystemMenu : public Menu {
   Animation ItemsFade;
   bool ItemsFadeComplete = false;
   glm::vec2 BGPosition{};
+  int LastFocusedButtonId = -1;
```
PringlesGang: Don't really like sentinel values. They can often lead to confusion when working with them
I'd prefer something like `std::optional<int> LastFocusedButtonId;`

dendi2008: Done.

PR #325, opened 2025-11-23
Nametag animation

Gunoshozo:
- Nametag animation via external logic somewhat similar to AutoIcon and WaitIcon
- Optional parameter initialization based on enum type of animation
- Some minor alignment improvements

Vids:
[CCLCC](https://youtu.be/aeYFCLsRdXo)
[CHLCC](https://youtu.be/9nQiotmAgk8) just to check that I did not broke anything (Some additional adjustments were made to not break it)

### DIFF
--- a/src/games/cc/dialoguebox.cpp
+++ b/src/games/cc/dialoguebox.cpp
@@ -1,5 +1,6 @@
 #include "dialoguebox.h"
 
+#include "../../hud/nametagdisplay.h"
 #include "../../renderer/renderer.h"
 #include "../../profile/dialogue.h"
 #include "../../profile/game.h"
@@ -26,24 +27,63 @@ void DialogueBox::Update(float dt) {
   TextBoxEffect.Update(dt);
 }
 
-void DialogueBox::Render(DialoguePageMode mode, bool hasName, float nameWidth,
-                         uint32_t nameId, float opacity) {
+void DialogueBox::Render(DialoguePageMode mode, float nameWidth,
+                         std::optional<uint32_t> nameId, float opacity) {
   const glm::vec4 col = glm::vec4(1.0f);
   if (mode == DPM_ADV) {
     Renderer->DrawCCMessageBox(ADVBoxSprite, ADVBoxMask, ADVBoxPos, col,
                                (int)(opacity * 272.0f), 16,
                                TextBoxEffect.Progress);
-    if (hasName) {
-      Renderer->DrawSprite(NamePlateMainSprites[nameId],
-                           ADVBoxNamePlateMainPos);
-      Renderer->DrawSprite(NamePlateLabelSprites[nameId],
-                           ADVBoxNamePlateLabelPos);
+    if (!nameId.has_value()) {
+      return;
     }
-  } else {
-    glm::vec4 nvlBoxTint(0.0f, 0.0f, 0.0f, opacity * NVLBoxMaxOpacity);
-    Renderer->DrawQuad(RectF(0, 0, Profile::DesignWidth, Profile::DesignHeight),
-                       nvlBoxTint);
+
+    if (NametagDisplay::IsHiding()) {
+      // hiding old name
+      const float progressHiding = NametagDisplay::GetHidingProgress();
+      RenderNameSprites(nameId.value(), progressHiding,
+                        AnimationDirection::Out);
+    } else if (NametagDisplay::IsShowing()) {
+      // showing new name
+      float progressShowing = NametagDisplay::GetShowingProgress();
+      RenderNameSprites(nameId.value(), progressShowing,
+                        AnimationDirection::In);
+    }
+    return;
   }
+
+  glm::vec4 nvlBoxTint(0.0f, 0.0f, 0.0f, opacity * NVLBoxMaxOpacity);
+  Renderer->DrawQuad(RectF(0, 0, Profile::DesignWidth, Profile::DesignHeight),
+                     nvlBoxTint);
+}
+
+void DialogueBox::RenderNameSprites(uint32_t nameId, float progress,
+                                    AnimationDirection direction) {
+  const float invertedProgress = 1.0f - progress;
+
+  const float alpha =
+      direction == AnimationDirection::In ? progress : invertedProgress;
+
+  glm::vec4 tint = glm::vec4(1.0f, 1.0f, 1.0f, alpha);
+
+  const float translationProgress =
+      (direction == AnimationDirection::In ? invertedProgress : progress);
+
+  const glm::vec2 mainPlateTranslation = glm::vec2(
+      -NamePlateMainSprites[nameId].Bounds.Width * translationProgress, 0);
+
+  const CornersQuad mainPlateDest = NamePlateMainSprites[nameId]
+                                        .ScaledBounds()
+                                        .Translate(ADVBoxNamePlateMainPos)
+                                        .Translate(mainPlateTranslation);
+
+  const CornersQuad labelPlateDest = NamePlateLabelSprites[nameId]
+                                         .ScaledBounds()
+                                         .Translate(ADVBoxNamePlateLabelPos)
+                                         .ScaleAroundCenter({alpha, 1.0f});
+
+  Renderer->DrawSprite(NamePlateMainSprites[nameId], mainPlateDest, tint);
+  Renderer->DrawSprite(NamePlateLabelSprites[nameId], labelPlateDest, tint);
 }
 
 }  // namespace CC
--- a/src/hud/nametagdisplay.cpp
+++ b/src/hud/nametagdisplay.cpp
@@ -1,0 +1,154 @@
+#include "nametagdisplay.h"
+
+namespace Impacto {
+namespace NametagDisplay {
+
+using namespace Impacto::Profile::Dialogue;
+
+static std::optional<uint32_t> NameId;
+static std::optional<uint32_t> PrevNameId;
+
+void Init() {
+  if (NametagCurrentType == +NametagType::None) {
+    return;
+  }
+
+  switch (NametagCurrentType) {
+    case NametagType::None:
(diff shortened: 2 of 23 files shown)


src/hud/nametagdisplay.h
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#include <enum.h>
+#include <glm/glm.hpp>
```
PringlesGang: Don't see `glm/glm.hpp` being used anywhere

src/hud/nametagdisplay.cpp
```diff
@@ -0,0 +1,177 @@
...
+          (NametagAnimation.Progress < NameTagAnimProgressHideOld)) {
+        hasName = true;
+        animatedName = PrevNameId;
+      }
+      break;
+    }
+  }
+
+  return std::make_pair(hasName, animatedName);
+}
```
PringlesGang: I think this would make more sense as returning a `std::optional<uint32_t>`
(in fact, I think the whole `HasName` and `PrevHasName` could just be replaced with `std::optional<uint32_t> (Prev)NameId`)

src/hud/nametagdisplay.cpp
```diff
@@ -0,0 +1,177 @@
...
+    case NametagType::None:  // falls through
+    case NametagType::Instant: {
+      return;
+    }
+    case NametagType::FadeInPauseOut: {
+      if (!(HasName || PrevHasName) ||
+          (HasName && PrevHasName && NameId == PrevNameId)) {
+        // if name state stayed the same, still no name or same name
+        NametagAnimation.Progress = 1;
+        NametagAnimation.Stop();
```
PringlesGang: `NametagAnimation.Finish()`

src/hud/nametagdisplay.cpp
```diff
@@ -0,0 +1,177 @@
...
+      } else if (HasName && !PrevHasName) {
+        // skip hiding name if there was none
+        NametagAnimation.Progress = NameTagAnimProgressShowNew;
+        NametagAnimation.StartIn();
+      } else {
+        // start with hiding prev name
+        NametagAnimation.Progress = 0;
+        NametagAnimation.StartIn();
```
PringlesGang: `NametagAnimation.StartIn(true)`

src/hud/nametagdisplay.h
```diff
@@ -0,0 +1,33 @@
...
+                 uint32_t prevNameId, DialoguePageMode mode);
+
+std::pair<bool, uint32_t> GetNameToDraw();
+
+float GetHidingProgress();
+float GetShowingProgress();
+
+bool DialogueCanStartAppearing(DialoguePageMode mode);
+bool isHiding();
+bool isShowing();
```
PringlesGang: These can be `const`, and `isHiding()` and `isShowing()` should be in PascalCase

src/hud/nametagdisplay.cpp
```diff
@@ -0,0 +1,177 @@
...
+      return (NametagAnimation.Progress - NameTagAnimProgressShowNew) /
+             (1.0f - NameTagAnimProgressShowNew);
+    }
+    default: {
+      return 1.0f;
+    }
+  }
+}
+
+void StartHiding() { NametagAnimation.StartIn(true); }
```
PringlesGang: This and `IsHiding`/`IsShowing` should be moved to the header so they can be inlined

src/profile/dialogue.h
```diff
@@ -90,6 +91,12 @@ inline int PageCount;
 
 inline bool ColorTagIsUint8;
 
+inline NametagDisplay::NametagType NametagCurrentType =
+    NametagDisplay::NametagType::None;
+inline float NameTagDuration;
+inline float NameTagAnimProgressHideOld;
+inline float NameTagAnimProgressShowNew;
```
PringlesGang: I think these should still have default values, so even invalid state doesn't have undefined behaviour

src/games/cc/dialoguebox.cpp
```diff
@@ -33,17 +33,64 @@ void DialogueBox::Render(DialoguePageMode mode, bool hasName, float nameWidth,
...
+      NamePlateLabelSprites[nameId].Bounds.GetSize();
+  namePlateScaleOrigin.x /= 2.0f;
+
+  const glm::mat4 namePlateTransformationMatrix = TransformationMatrix(
+      namePlateScaleOrigin, namePlateScale, {0, 0}, 0, {0, 0});
+
+  Renderer->DrawSprite(NamePlateMainSprites[nameId], dest,
+                       mainPlateTransformationMatrix, tint);
+  Renderer->DrawSprite(NamePlateLabelSprites[nameId], namePlateDest,
+                       namePlateTransformationMatrix, tint);
```
PringlesGang: Transformation matrices are quite computationally expensive on the renderer; I'd really only use them if they use rotations or if they affect a _lot_ of vertices
Something like
```cpp
const RectF dest = NamePlateMainSprites[nameId].ScaledBounds()
	.Translate(ADVBoxNamePlateMainPos)
	.Translate(mainPlateTranslation);

const RectF namePlateDest = NamePlateLabelSprites[nameId].ScaledBounds()
	.Translate(ADVBoxNamePlateLabelPos)
	.ScaleAroundCenter({1.0f, alpha});

Renderer->DrawSprite(NamePlateMainSprites[nameId], dest, tint);
Renderer->DrawSprite(NamePlateLabelSprites[nameId], namePlateDest, tint);
```
works directly on the 8 total vertices, and won't have to swap the transformation matrix uniform in the shader, making it _way_ faster (and also easier to read imo)

src/text.cpp
```diff
@@ -997,6 +1023,26 @@ void DialoguePage::MoveTo(glm::vec2 pos) {
...
+  if (!HasName) {
+    return;
+  }
+
+  PrevNameId = NameId;
+  HasPrevName = true;
+  HasName = false;
+  NametagDisplay::UpdateNames(HasName, NameId, HasPrevName, PrevNameId, Mode);
+  // force animation start
+  NametagDisplay::StartHiding();
```
PringlesGang: I think it's better to put this stuff in an `if (HasName)` block than the other way around, in case we ever need to add more stuff that has to change on hide

src/text.cpp
```diff
@@ -568,17 +565,25 @@ void DialoguePage::EndRubyBase(int lastBaseCharacter) {
...
-  CurrentLineVoiced = voice != nullptr;
-  // Hold last voiced animation id
-  if (CurrentLineVoiced) AnimationId = std::max(animId, 31);
+  CurrentVoice = voice;
+  CurrentLineVoiced = CurrentVoice != nullptr;
+  if (CurrentLineVoiced) {
+    // Hold last voiced animation id
+    AnimationId = std::max(animId, 31);
+    // Stop prev voice if current is voiced, else lipsync can start early
+    Audio::Channels[Audio::AC_VOICE0]->Stop(0.0f);
```
PringlesGang: The voice line should only cut off the moment the next one begins, so maybe instead of stopping the voice line early here, we should only set the new voiced characters when the new audio begins?

src/hud/nametagdisplay.h
Gunoshozo: Do you mean make them const method? 
But there's no class here, so no `this`

PringlesGang: Oh lmao yeah you're right. Just the casing then :P

src/hud/nametagdisplay.h
Gunoshozo: Done

src/hud/nametagdisplay.cpp
Gunoshozo: Done, changed it also for text.cpp and other usages

src/hud/nametagdisplay.cpp
Gunoshozo: Done

src/hud/nametagdisplay.cpp
Gunoshozo: right...

src/hud/nametagdisplay.h
Gunoshozo: Done

src/hud/nametagdisplay.cpp
Gunoshozo: Done

src/profile/dialogue.h
Gunoshozo: Done

src/games/cc/dialoguebox.cpp
Gunoshozo: Done

src/text.cpp
Gunoshozo: Done, but would like you to check the new version)

src/hud/nametagdisplay.cpp
```diff
@@ -0,0 +1,156 @@
...
+}
+
+bool DialogueCanStartAppearing(DialoguePageMode mode) {
+  if (mode != DialoguePageMode::DPM_ADV) {
+    return true;
+  }
+  switch (NametagCurrentType) {
+    case NametagType::FadeInPauseOut: {
+      const bool hasCurrentName = NameId.has_value();
+      return (!hasCurrentName &&
```
KKhanhH: Maybe use ishiding and isshowing?

src/profile/dialogue.h
```diff
@@ -8,7 +8,6 @@
 #include "../hud/skipicondisplay.h"
 #include "../hud/dialoguebox.h"
 #include "../spriteanimation.h"
-
```
KKhanhH, line 11: not the space

KKhanhH (approved):

src/hud/nametagdisplay.cpp
Gunoshozo: `!IsHiding()`*

PR #326, opened 2025-11-24
Chlcc/erin scanline

KKhanhH:
Adds the moving scanline in computer monitor sections

Also moves the r;ne dash specific rendering code to rendermain


src/ui/gamespecific.cpp
```diff
@@ -52,14 +56,47 @@ void Update(float dt) {
...
   }
 }
 
 void RenderLayer(uint32_t layer) {
   switch (Profile::GameSpecific::GameSpecificType) {
     case +GameSpecificType::CHLCC: {
+      if (ScrWork[SW_MONITOR_SCANLINE_ENABLED] &&
+          static_cast<int>(layer) == ScrWork[SW_MONITOR_SCANLINE_LAYER]) {
+        static int offsetY = 0;
+        offsetY = (offsetY += 1) % 300;
```
PringlesGang: This should happen in `Update` and be dependent on `dt`

PR #327, opened 2025-11-24
CHLCC Delusion struct size

KKhanhH:


Gunoshozo:
While you're at it, could you also fix this gap?
image

Issue #332, opened 2025-11-26
Text(lines) should appear instantly when Message Speed is set to maximum.


dendi2008:
- [ ] Make text(lines) appear instantly when Message Speed is set to maximum.
- [ ] Add fade In animation (when new text(lines) appears) and fade Out animation (after click).

PR #334, opened 2025-11-27
Add text fade In and fade Out animation.

dendi2008:
This PR will add text fade In and fade Out animation. (issue #332).


Gunoshozo (changes requested):
Hey there 👋 
Thanks for your contribution once again, that's really cool that you continue to help with the project!

I've noticed a few subtle differences. 
For C;CLCC your implementation starts nametag animation only after dialogue text fade out, but in the original game and in the C;C they start to disappear at the same time. 
https://www.youtube.com/live/ZDj8qLcAxCQ?t=996s
For CHLCC your implementation makes nametag text fade out, while in the original it disappears/swaps instantly after dialogue text fade out.
https://youtu.be/0dR8NOQU0Fg?si=pxDyP8hdmX-E6cKT&t=1478

Here's also other games examples:
https://www.youtube.com/watch?v=o9bOcJQ-_V8
https://www.youtube.com/watch?v=7eiTZ6bJuMU
https://www.youtube.com/watch?v=fWHdnQf0Ws8
https://www.youtube.com/watch?v=zfjuvf1j2mo

You don't have to implement it for other games, it's just an example to roughly estimate how general your solution should be

Some notes regarding nametag behaviour, since it is tied to the text fade:
1. Some games (mostly memOff) also can have different behaviour for nametag if a line is voiced or not, but don't bother with it at the moment, since I don't think I saw it happen in LCCs
2. Impacto currently supports name text itself being animated, but if nametext and nameplate behave differently (like in Iwakura Aria) that will require additional support, maybe on a per game basis

Gunoshozo (changes requested):
Ok, nametag looks good

And another thing that I've just noticed is that WaitIcon is also out of sync, It should disappear/start dissapearing at the same time dialogue text starts to fade out. (CHLCC instantly, for CCLCC there's already an animation for it)
 
Looking at the code it seems to be controlled by `SF_SHOWWAITICON` flag

dendi2008:
> And another thing that I've just noticed is that WaitIcon is also out of sync, It should disappear/start dissapearing at the same time dialogue text starts to fade out. (CHLCC instantly, for CCLCC there's already an animation for it)
> 


Thanks, again. Fixed.

Gunoshozo (approved):
Nice

That's an approve from me. I have asked other devs to take a look. I guess they will check, when they will have time

PR #346, opened 2025-12-16
Smol fixes

PringlesGang:
- [x] Refactor `InstKeyOnJump` and `GetControlState` to only *check* mouse wheel for specific input groups
- [x] Display the tips notification in front of the NVL mode fade
- [x] Load in-progress movies when loading from a save file
- [x] Actually fix lip sync fr this time

Okay I lied these fixes aren't as smol as I had hoped :sob:

### DIFF
--- a/src/character2d.cpp
+++ b/src/character2d.cpp
@@ -362,7 +362,7 @@ void Character2D::UpdateState(const int chaId) {
     LipFrame = ScrWork[SW_CHA1ANIME_MOUTH + structOffset];
   } else {
     bool charSpeaking = false;
-    const uint32_t chaIndexMask = 1 << chaId & 0x1F;
+    const uint32_t chaIndexMask = 1 << (chaId & 0x1F);
 
     if (!SkipModeEnabled) {
       for (uint8_t i = 0; i < 3; i++) {


KKhanhH:
yeah I don't like the hard coded backlog check for scroll input tbh. I feel like what you're supposed to do is edit the scripts to specifically add in scroll input to stuff like backlog

PringlesGang:
I don't think it should be edited into the scripts, the official PC ports don't have it that way either afaik. I do agree that it shouldn't be hardcoded like this
I think the real solution is to add scroll up/down endpoints to `PADinputMouse`, and have them be accessible in the `PADcustom` lua profile. (At that point we should probably also expose left and right mouse buttons there as well)

profiles/chlcc/scriptvars.lua
```diff
@@ -147,10 +147,13 @@ sv.SW_EFF_CAP_PRI = 3268;
 sv.SW_EFF_CAP_BUF = 3269;
 sv.SW_EFF_CAP_PRI2 = 3270;
 sv.SW_EFF_CAP_BUF2 = 3271;
-sv.SW_MOVIEALPHA = 3411;
+sv.SW_MOVIEPRI2 = 3400;
```
KKhanhH: makes more sense to have pri first

src/vm/inst_movie.cpp
```diff
@@ -23,68 +23,134 @@ using namespace Impacto::Profile::ScriptVars;
...
+    ScrWork[SW_MOVIE_PLAYMODE + 20 * channel] = playMode + 20 * channel;
+    ScrWork[SW_MOVIE_PLAYVIEW + 20 * channel] = playView;
+    ScrWork[SW_MOVIE_LOADNO + 20 * channel] = 0xffff;
+
+    if (Video::Players[channel]->IsPlaying) Video::Players[channel]->Stop();
+    Video::Players[channel]->Play(stream, playMode == 5, playMode == 5);
+
+    SetFlag(SF_MOVIE_DRAWWAIT + channel, true);
+    SetFlag(SF_MOVIEPLAY + channel, true);
+    SetFlag(1822 + channel, false);
```
KKhanhH: SF_MOVIECANCEL

src/vm/inst_movie.cpp
```diff
@@ -100,50 +166,137 @@ VmInstruction(InstMovie) {
...
+      if (GetFlag(SF_MOVIEPLAY + playerId) && player.CancelWaitTime != 0) {
+        player.CancelWaitTime++;
+        if (player.CancelWaitTime < 8) {
+          ResetInstruction;
+          BlockThread;
+          return;
+        }
+
+        SetFlag(SF_MOVIEPLAY + playerId, false);
+        SetFlag(1856 + playerId, true);
```
KKhanhH: Did you get this from memoff? this should also be SF_MOVIECANCEL. 1856 looks like movie cancel in memoff and 1822 in ccs

src/vm/inst_movie.cpp
```diff
@@ -100,50 +166,137 @@ VmInstruction(InstMovie) {
   }
 }
 VmInstruction(InstMovieMain) {
```
KKhanhH, line 168: this instruction is a lot more complicated than what's in the chaos instructions

src/game.cpp
```diff
@@ -469,6 +467,10 @@ void Render() {
                     ->ScreenCap);
             SetFlag(SF_SYSTEMMENUCAPTURE, false);
           }
+
+          if (!GetFlag(SF_UIHIDDEN) && !GetFlag(SF_MOVIEPLAY)) {
```
KKhanhH, line 469: this is consistent on both games right?

KKhanhH (changes requested):

src/vm/inst_movie.cpp
PringlesGang: Yea I did. Couldn't pinpoint 1822 and 1856 exactly because of their differing values between CHLCC, CCLCC, and MOFDH, but I'll have another look if I agree with SF_MOVIECANCEL

src/game.cpp
PringlesGang: Yes

src/vm/inst_movie.cpp
PringlesGang: Yeah, CHLCC only handles types 0–3, but whereas MOFDH handles the second movie players through types 20–23, and has a 4 and 24 case as well, but these shouldn't show up in the LCCs. The implementation of these types are also near-identical

profiles/chlcc/scriptvars.lua
PringlesGang: I ordered them in terms of ascending ScrWork index

KKhanhH: Shouldnt it be 3400 is sw_moviepri and 3410 is pri2?

PringlesGang: No that would make too much sense :))
(I've reviewed the binary, and genuinely it looks like this is correct...)

KKhanhH: ???!??

KKhanhH: Pri at 10, pri2 at 0, alpha2 at 21, make it make sense

PringlesGang:
```lua
    SW_MOVIEPRI=6330,
    SW_ALPHAMOVIE_PRI=6331,
    SW_ALPHAMOVIE_ALPHA=6332,
    SW_MOVIE_PLAYNO=6333,
    SW_MOVIE_PLAYMODE=6334,
    SW_MOVIE_PLAYVIEW=6335,
    SW_MOVIE_LOADNO=6336,
    SW_ALPHAMOVIE_MODE=6337,
    SW_MOVIEALPHA=6338,
    SW_MOVIEPRI2=6339,
    SW_MOVIEALPHA2=6340,
    SW_MOVIEPRI3=6331,
    SW_MOVIEALPHA3=6332,
    SW_MOVIEPRI4=6341,
    SW_MOVIEALPHA4=6342,
```
sure thing bud. Another banger by Johnny Mages

PR #347, opened 2025-12-19
Background and screencapture rendering fixes

PringlesGang:
Rewrites core background and screencapture logic to more closely reflect its behaviour in the binary

CHLCC never renders its captures separately; always _through_ a background object.
CCLCC _does_ render its captures separately, along with through background objects.
This used to be handled by `Vm::Interface::UpdateBackground2D`, but that implementation was flawed, as when a background is rendered separately, `SW_CAP` variables are used, whereas when it's rendered through a background object, the corresponding `SW_BG` variables are used instead.

The distinction between background _buffer_ and background _id_ was also sometimes mishandled.

Between the two engine revisions, the `SW_BG1DISPMODE` variable also handles the background transformations differently. The class now properly changes behaviour based on the revision of the engine it is meant to emulate.


KKhanhH:
Do all instances of bgsprite need to be replaced with rendersprite then?

PringlesGang:
> Do all instances of bgsprite need to be replaced with rendersprite then?

Not all of them. `RenderSprite` is the sprite that is ultimately to be rendered by that object in that pass; `BgSprite` still holds the texture that the underlying background "normally" renders

src/background2d.h
```diff
@@ -34,6 +34,7 @@ class Background2D : public Loadable<Background2D, bool, uint32_t> {
   static void InitFrameBuffers();
 
   Sprite BgSprite;
+  Sprite* RenderSprite = nullptr;
```
KKhanhH: Would a reference wrapper make more sense? I think it should alwayd point to bgsprite or screencap, but never null

PringlesGang: If I can immediately there in the definition or constructor say `RenderSprite = BgSprite` then yes, that would make a lot more sense

KKhanhH: You should be able to, so long as you declare it after. Copy/move semantics might not be so nice but i dont think we copy this

Enorovan (approved):

src/background2d.cpp
```diff
@@ -272,13 +279,182 @@ void Background2D::LinkBuffers(const int linkCode, const int currentBufferId) {
...
+                                              capOfsStruct[SW_CAP1LY_OFS]))
+                         .Scale(resolutionScale, glm::vec2(0.0f));
+
+  return state;
+}
+
+void Background2D::SetTransformState(int dispMode, BgTransformState state) {
+  TransformState = BgTransformState();
+
+  if (ScrWorkBgStructSize >= 40) {
```
KKhanhH: it's kinda mostly the same except for two cases, maybe could have a switch statement for the common stuff and if statements for the cases that differ?

src/background2d.cpp
```diff
@@ -288,76 +464,16 @@ void Background2D::UpdateState(const int bgId) {
...
-                                ScrWork[SW_BG1POSY_OFS + structOfsOffset]) *
-                 resolutionScale / 1000.0f;
-
-      Scale = {1.0f, 1.0f};
-    } break;
+  RenderSprite = BgSprite;
+  for (int i = static_cast<int>(Profile::ScreenCaptureCount) - 1; i >= 0; i--) {
+    const int capOffset = i * Profile::Vm::ScrWorkCaptureEffectInfoStructSize;
+    if (ScrWork[SW_EFF_CAP_BUF + capOffset] == GetScriptBufferId(bgId + 1)) {
+      RenderSprite = Screencaptures[i].BgSprite;
```
KKhanhH, line 449: instead of having two sprites, what if we just have a single sprite and update the spritesheet when necessary to the screencap sheet?

KKhanhH (approved):

src/background2d.cpp
PringlesGang: That would necessitate storing the `SpriteSheet` separately in each `Background`, as currently `BgSprite` is the only thing holding on to that. That ok?

KKhanhH: Yeah that should be fine. You're already duplicating spritesheet by duplicating sprite, so there should just be less to dupe now.

PringlesGang: Idk man, I've just tried to dedupe the stuff, but the `Sprite` actually adds crucial information about the texture beyond what is already stored in the `SpriteSheet`

Mainly the `BaseScale` is important here, as for the captures and renderbuffers this represents the ratio between the window size and the design size, which is information that is not stored in the `SpriteSheet` but is crucial for how the texture is supposed to be rendered
Technically this _could_ be set every time when the `RenderSprite` would be set, but this would also necessitate resetting it to either `glm::vec2(1.0f)` every time in case no capture is selected (and this is assuming there aren't ever any backgrounds that have a different `BaseScale`); or to a separately stored base scale, still duping it

There's an argument to be made for the `Bounds`, which is needlessly duped, but I don't think that fully justifies the above drop in code quality :/

tldr, I think the two sprites truly carry different semantic meaning: `BgSprite` is the full sprite held by the class, denoting everything that the class _could_ render on that sheet, whereas `RenderSprite` denotes what part of that (or another) sheet _should_ be rendered that frame

KKhanhH: Sure, you can just leave it as bgsprite then

PR #348, opened 2025-12-21
BG Wave Effect

Gunoshozo:
- Add BGWave effect implementation
- Add complete InstBGeffectWave instruction implementation
- OpenGL support only (Flush if topology has changed)
- Support saving and loading for wave data
- Add dummies for EffWave and CHAWave


src/profile/data/bgeff.cpp
```diff
@@ -10,70 +10,72 @@ namespace BgEff {
...
-      auto arraySize = lua_rawlen(LuaState, -1);
-      if (arraySize != 5) {
-        ImpLog(LogLevel::Fatal, LogChannel::Profile,
-               "Expected five entries in shader definition defined; got {:d}\n",
-               arraySize);
-        Window->Shutdown();
+    if (TryPushMember("BgEffShaderData")) {
+      PushInitialIndex();
+      while (PushNextTableElement() != 0) {
+        AssertIs(LUA_TTABLE);
```
PringlesGang: I designed this load function to ensure that if the config states the BgEff is used, _all_ of its necessary data is correctly mapped out. Replacing all `Ensure` methods with `Try` ones kinda drops all that
Furthermore, `BgEff` is a very specific kind of effect that MAGES never bothered giving a better name, which is quite different from the wave effect (it's the effect where multiple layers of shaders are placed over the background/characters to give the illusion of actual lighting). I think the wave settings should probably just be in their own new configuration file that is separately loaded

src/renderer/opengl/renderer.cpp
```diff
@@ -333,6 +333,11 @@ void Renderer::DrawSprite(const Sprite& sprite, const CornersQuad& dest,
     return;
   }
 
+  if (LastTopologyMode != TopologyMode::Triangles) {
+    Flush();
+    LastTopologyMode = TopologyMode::Triangles;
+  }
+
```
PringlesGang: These should probably be in a helper method like `EnsureTopologyMode(TopologyMode newMode)`

src/renderer/opengl/renderer.cpp
```diff
@@ -785,8 +812,25 @@ void Renderer::Flush() {
...
+        glPrimitiveRestartIndex(0xFFFF);
+        glEnable(GL_PRIMITIVE_RESTART);
+        break;
+      }
+      default: {
+        mode = GL_TRIANGLES;
+        glDisable(GL_PRIMITIVE_RESTART);
+      }
+    }
+    glDrawElements(mode, (GLsizei)IndexBuffer.size(), GL_UNSIGNED_SHORT, 0);
```
PringlesGang, line 812: The `default` case should just be right above the `Triangles` case and fall through:
```cpp
switch (LastTopologyMode) {
    case default:
    case TopologyMode::Triangles: {
        mode = GL_TRIANGLES;
        glDisable(GL_PRIMITIVE_RESTART);
        break;
    }
```

src/renderer/renderer.h
```diff
@@ -367,6 +370,7 @@ class BaseRenderer {
     return designCoord;
   }
 
+  TopologyMode LastTopologyMode;
```
PringlesGang: This field is implementation-specific

src/background2d.cpp
```diff
@@ -731,4 +774,13 @@ void Background2D::RenderFade() {
...
 }
 
+bool IsBgWaveEffectActive() {
+  for (const auto& currentBg : Backgrounds) {
+    if (currentBg.RenderType == 14 || currentBg.RenderType == 28) {
+      return true;
+    }
+  }
+  return false;
+}
```
PringlesGang, line 763:
```cpp
bool IsBgWaveEffectActive() {
  return std::ranges::any_of(Backgrounds, [](const auto& currentBg) {
    return currentBg.RenderType == 14 || currentBg.RenderType == 28;
  });
}
```

src/effects/wave.h
```diff
@@ -0,0 +1,82 @@
...
+    const float coefficient = dt * 60.0f;
+    for (size_t i = 0; i < WaveCount; i++) {
+      WaveData[i].Phase +=
+          static_cast<int>(WaveData[i].TemporalFrequency * coefficient);
+    }
+  }
+
+  virtual void Init() {
+    WaveData = std::vector<WaveParams>(Profile::BgEff::WaveMaxCount);
+  };
```
PringlesGang: `WaveData.resize(Profile::BgEff::WaveMaxCount)` (Same thing in the other `Init`s)

src/effects/wave.h
```diff
@@ -0,0 +1,82 @@
...
+  int SpatialFrequency;
+};
+
+class Wave {
+ public:
+  std::vector<WaveParams> WaveData{};
+  std::vector<glm::vec2> WavePos{};
+  std::vector<VertexBufferSprites> Vertices{};
+  std::vector<uint16_t> Indices{};
+  uint32_t WaveCount;
```
PringlesGang: I'd initialize the `WaveCount` to 0

src/effects/wave.h
```diff
@@ -0,0 +1,82 @@
...
+    WavePos = std::vector<glm::vec2>(28 * 20);
+  }
+  void CalcPos(int startPhase) override;
+  std::pair<std::span<VertexBufferSprites>, std::span<uint16_t>> GetPrimitives()
+      override;
+};
+
+inline BGWave WaveBg = BGWave();
+inline EFFWave WaveEFF = EFFWave();
+inline CHAWave WaveCHA = CHAWave();
```
PringlesGang: Casing inconsistency with `Bg`, `EFF`, and `CHA` :p

src/effects/wave.cpp
```diff
@@ -0,0 +1,186 @@
...
+void BGWave::CalcPos(int startPhase) {
+  std::vector<float> left = std::vector(BGWaveGridSize.y, 0.0f);
+  std::vector<float> right =
+      std::vector(BGWaveGridSize.y, (float)Profile::ResolutionWidth);
+
+  std::vector<float> top = std::vector(BGWaveGridSize.x, 0.0f);
+  std::vector<float> bottom =
+      std::vector(BGWaveGridSize.x, (float)Profile::ResolutionHeight);
+
+  constexpr float from16BitToRadians = static_cast<float>(2 * M_PI / 65535.0f);
```
PringlesGang: `ScrWorkAngleToRad(int angle)` in `util.h`

src/effects/wave.cpp
```diff
@@ -0,0 +1,186 @@
...
+      // top and bottom are switched since original engine has an upside-down
+      // mesh
+      if (flags & 4) {
+        top[j] += ampedSin - amplitude;
+      }
+
+      if (flags & 8) {
+        bottom[j] += ampedSin + amplitude;
+      }
+    }
```
PringlesGang: This is technically assuming that `BGWaveGridSize.x >= BGWaveGridSize.y`. Not an unfair assumption to make, but still... To be safe, _I_ would still rewrite the loop to take the max, and guard both horizontal and vertical, but just adding an assert when loading the profile is already enough imo

src/games/cclcc/savesystem.h
```diff
@@ -81,6 +83,9 @@ class SaveSystem : public SaveSystemBase {
 
   void SetCheckpointId(int id) override;
 
+  void WaveSave(int* data);
+  void WaveLoad(int* data) const;
```
PringlesGang: Use spans to catch OOB errors at runtime in debug mode

src/effects/wave.cpp
Gunoshozo: Took the max and guarded horz and vertical vectors

Enorovan (approved):

src/effects/wave.cpp
PringlesGang: I moreso meant you can just `const float ampedSin = amplitude * std::sin(ScrWorkAngleToRad(currentPhase));`

Enorovan (approved):
- C;CLCC Post-Delusion Trigger ✅ 
- C;HLCC Pool ✅ 
- C;HLCC Common End✅ 
  - #347 fixes what could have caused a warping on the effect in particular circumstances ✅

src/games/cclcc/savesystem.h
```diff
@@ -81,6 +83,9 @@ class SaveSystem : public SaveSystemBase {
 
   void SetCheckpointId(int id) override;
 
+  void WaveSave(std::span<int> data);
+  void WaveLoad(std::span<int> data) const;
```
KKhanhH: load should be const int

src/effects/wave.h
```diff
@@ -0,0 +1,80 @@
...
+class Wave {
+ public:
+  std::vector<WaveParams> WaveData{};
+  std::vector<glm::vec2> WavePos{};
+  std::vector<VertexBufferSprites> Vertices{};
+  std::vector<uint16_t> Indices{};
+  uint32_t WaveCount = 0;
+
+  void ClearWaves();
+  void AddWave(int flags, int amplitude, int nextPhase, int initialPhase,
```
KKhanhH: could these take in a WaveParams instead? lets you use designated initializers rather than have a bunch of int params next to each other.

src/effects/wave.h
```diff
@@ -0,0 +1,80 @@
...
+    const float coefficient = dt * 60.0f;
+    for (size_t i = 0; i < WaveCount; i++) {
+      WaveData[i].Phase +=
+          static_cast<int>(WaveData[i].TemporalFrequency * coefficient);
+    }
+  }
+
+  virtual void Init() { WaveData.resize(Profile::WaveEffects::WaveMaxCount); };
+  virtual void CalcPos(int startPhase) = 0;
+  virtual std::pair<std::span<VertexBufferSprites>, std::span<uint16_t>>
```
KKhanhH: Maybe have a struct for this so it's more obvious what it returns.

```cpp
struct PrimitivePair {
  std::span<VertexBufferSprites> Vertices;
  std::span<uint16_t> Indices;
  template <typename T, typename U>
  operator std::tuple<T, U>() {
    return std::tuple<T, U>{Vertices,Indices};
  }
};
```

src/games/chlcc/savesystem.cpp
```diff
@@ -1219,5 +1227,55 @@ Sprite& SaveSystem::GetSaveThumbnail(SaveType type, int id) {
       "Tried to get thumbnail of unimplemented save entry type {}", (int)type));
 }
 
+void SaveSystem::WaveSave(std::span<uint16_t> data) {
```
KKhanhH, line 1230: endianess could be a concern but I don't think cclcc handles it either right now (and no official big endian target).

src/background2d.cpp
```diff
@@ -691,6 +692,48 @@ void BackgroundEffect2D::Render(const int layer) {
   Renderer->SetStencilMode(StencilBufferMode::Off);
 }
 
+void Background2D::RenderBgWave1() {
```
KKhanhH: bit nitpicky but you could template on a bool this since they're basically the same and if constexpr the calcpos argument. just need to explicitly instantiate the template in cpp file.

KKhanhH (changes requested):

src/effects/wave.cpp
Gunoshozo: ✅

PR #349, opened 2025-12-25
BGM fade-outs

Gunoshozo:
- remember the ratio of base gain to current actual gain to use in the fade out calculations


Enorovan (approved):
Checked both local and CI builds on Windows

PR #362, opened 2025-12-27
C;HLCC: Optional PSP Videos in Movie Menu

Enorovan:


src/games/chlcc/moviemenu.cpp
```diff
@@ -405,6 +405,14 @@ void MovieMenu::UpdateMovieEntries() {
   }
 }
 
+inline bool MovieMenu::IsExtraMoviesPresent() {
+  std::map<uint32_t, std::string> listing;
+  Io::VfsListFiles("movie", listing);
+  // There are 11 videos in the base game: the 10 ones in this menu + the smoke
+  // one. If there's more than that, it means the PSP videos are in.
+  return listing.size() > Movies + 1;
+}
```
PringlesGang: 1. Inline isn't necessary here because it's already in an implementation file, and is thus only ever compiled once
2. Maybe something like below is a better approach?
```cpp
bool MovieMenu::IsExtraMoviesPresent() {
    static std::optional<bool> isPresent;
    if (!isPresent.has_value()) {
        std::map<uint32_t, std::string> listing;
        Io::VfsListFiles("movie", listing);
        isPresent = listing.size() > Movies + 1;
    }
    return isPresent.value();
}
```
This caches the value within a static variable, making it so it only ever gets computed at most once, while not leaking the variable out of the function nor requiring you to set it explicitly in the constructor

src/ui/widgets/chlcc/moviemenuentrybutton.h
```diff
@@ -22,6 +22,8 @@ class MovieMenuEntryButton : public Widgets::Button {
 
  private:
   SpriteAnimation MovieBoxAnim;
+
+  bool* IsExtraMovieModeOn = nullptr;
```
PringlesGang: Make this a `std::reference_wrapper<bool>`

PringlesGang (changes requested):

src/games/chlcc/moviemenu.h
```diff
@@ -33,6 +51,8 @@ class MovieMenu : public Menu {
 
   void UpdateMovieEntries();
 
+  static bool IsExtraMoviesPresent();
```
KKhanhH: This doesnt need to be declared in the menu, yiu can just have it be static function in the cpp file

src/games/chlcc/moviemenu.cpp
```diff
@@ -405,6 +405,18 @@ void MovieMenu::UpdateMovieEntries() {
   }
 }
 
+bool MovieMenu::IsExtraMoviesPresent() {
+  static std::optional<bool> isPresent;
```
KKhanhH: Instead of optional, you can just make this a bool and assign it to [] {
Init code here
} ();
Then just return as usual

src/games/chlcc/moviemenu.h
```diff
@@ -9,6 +9,24 @@ namespace Impacto {
 namespace UI {
 namespace CHLCC {
 
+struct MovieButtonEntry {
+  int PhysicalId;
+  int ExtraId;
+};
+
+static const std::map<int, MovieButtonEntry> movieButtonMap = {
```
KKhanhH: Avoid static for non class members in headers, you should be using inline for globals. 
But better to actually move this into the cpp file and you can keep static in there.
(Static makes a variable private to a cpp file and if it's in a header it makes every cpp file that includes this have their own copy of the var, while inline makes them shared). 

Also since the keys are contiguous numbers, you can use an array and index into it, and make it constexpr, one less dynamic allocation and better optimization.

KKhanhH: Oh and capitalize it too

src/profile/games/chlcc/moviemenu.cpp
```diff
@@ -47,11 +49,16 @@ void Configure() {
...
+  MovieBoxExtra = EnsureGetMember<Sprite>("MovieBoxExtra");
   GetMemberArray<glm::vec2>(ThumbnailPositions, Movies, "ThumbnailPositions");
   GetMemberArray<Sprite>(MoviesThumbnails, Movies, "MoviesThumbnails");
+  MovieThumbnailExtraOp = EnsureGetMember<Sprite>("MovieThumbnailExtraOp");
   LockedThumbnail = EnsureGetMember<Sprite>("LockedThumbnail");
   ButtonPromptPosition = EnsureGetMember<glm::vec2>("ButtonPromptPosition");
   ButtonPromptSprite = EnsureGetMember<Sprite>("ButtonPromptSprite");
+  MovieButtonExtraPromptPosition =
+      EnsureGetMember<glm::vec2>("MovieButtonExtraPromptPosition");
+  MovieButtonExtraPrompt = EnsureGetMember<Sprite>("MovieButtonExtraPrompt");
```
PringlesGang: I wouldn't make these `Ensure`d, because they're inherently an extra, _optional_ thing we're adding that is not part of the base game.

Perhaps you could add a Boolean field to the profile to enable the feature in the first place, `TryGetMember<bool>` that one, default it to `false`, and then only if it's `true` do you `Ensure` the rest of the entries
(and obviously set `isPresent` in `IsExtraMoviesPresent` to `false` if this Boolean field is)

PringlesGang (changes requested):

src/profile/games/chlcc/moviemenu.h
```diff
@@ -51,6 +51,13 @@ inline Sprite LockedThumbnail;
 inline glm::vec2 ButtonPromptPosition;
 inline Sprite ButtonPromptSprite;
 
+inline std::optional<bool> MovieExtraVideosEnabled;
```
PringlesGang: Just `bool`; no `optional`. The `value_or` takes care of that

src/games/chlcc/moviemenu.cpp
```diff
@@ -18,30 +18,39 @@ using namespace Impacto::Vm::Interface;
...
+  if (!MovieExtraVideosEnabled) return false;
+  static bool isPresent = [] {
+    std::map<uint32_t, std::string> listing;
+    Io::VfsListFiles("movie", listing);
+    // There are 11 videos in the base game: the 10 ones in this menu + the
+    // smoke one. If there's more than that, it means the PSP videos are in.
+    return listing.size() > Movies + 1;
+  }();
+  return isPresent;
+}
```
PringlesGang, line 44: I'd early return `false` on the lambda tbh. Otherwise it'll still evaluate it on first call

PR #366, opened 2026-01-05
Adds additional VFX to CHLCC like bubbles and eyecatch.

KKhanhH:


src/games/chlcc/eyecatch.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include <cstdint>
+#include "../../animation.h"
```
PringlesGang: You're not using these here

src/effects/chlcc/eyecatch.cpp
```diff
@@ -0,0 +1,75 @@
...
+  if (!ScrWork[SW_EYECATCH_COUNT]) return;
+  Renderer->Clear(glm::vec4(0.0f));
+  for (int i = 0; i < 4; ++i) {
+    int scaleOffset = 2 * i;
+    for (int j = 0; j < 7; ++j) {
+      if (scaleOffset + 1 <= ScrWork[SW_EYECATCH_COUNT]) {
+        const int scale =
+            std::min((ScrWork[SW_EYECATCH_COUNT] - (scaleOffset + 1)) * 18,
+                     400) *
+            280 / 106;
```
PringlesGang, line 40: You sure this should be an int here already? It's doing some heavy early rounding here despite being casted back to a float anyway during the actual `ScaleAroundCenter`

src/effects/chlcc/bubbleseffect.h
```diff
@@ -0,0 +1,38 @@
...
+ public:
+  BubblesEffect();
+  void Init();
+  void Update(float dt);
+  void Render();
+
+  static BubblesEffect& GetInstance() {
+    static BubblesEffect impl;
+    return impl;
+  };
```
PringlesGang: If you wanna make this a singleton, it's probably best to make the constructor private (same for eyecatch)

src/games/chlcc/bubbleseffect.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+BubblesEffect::BubblesEffect() {
+  FadeAnimation.DurationIn = BubbleFadeDuration;
+}
+
+void Bubble::Render(float alphaMultipiler) {
+  const float sizeF = static_cast<float>(Size);
+  const float sine =
+      std::sin(RandAngle / 65536.0f * 2 * std::numbers::pi_v<float>);
```
PringlesGang: `std::sin(ScrWorkAngleToRad(RandAngle));`

src/effects/chlcc/bubbleseffect.cpp
```diff
@@ -0,0 +1,80 @@
...
+BubblesEffect::BubblesEffect() {
+  FadeAnimation.DurationIn = BubbleFadeDuration;
+}
+
+void Bubble::Render(float alphaMultipiler) {
+  const float sizeF = static_cast<float>(Size);
+  const float sine =
+      std::sin(RandAngle / 65536.0f * 2 * std::numbers::pi_v<float>);
+  const float xPos = Position.x + sizeF * sine * 0.5f - sizeF / 2.0f;
+  const RectF dest{xPos, Position.y - sizeF / 2.0f, sizeF, sizeF};
```
PringlesGang: This should take DesignDimensions into account. Currently it's hardcoded 720p

src/games/chlcc/bubbleseffect.cpp
```diff
@@ -0,0 +1,80 @@
...
+  if (Position.y + Size < 0.0f) {
+    Init(true);
+  }
+}
+
+BubblesEffect::BubblesEffect() {
+  FadeAnimation.DurationIn = BubbleFadeDuration;
+}
+
+void Bubble::Render(float alphaMultipiler) {
```
PringlesGang: `alphaMultipiler` here too lmao

src/effects/chlcc/eyecatch.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+EyecatchEffect::EyecatchEffect() {
+  Texture textureStarsMask{};
+  textureStarsMask.LoadSolidColor(Window->WindowWidth, Window->WindowHeight, 0);
+  SpriteSheet sheetStarsMask(static_cast<float>(Window->WindowWidth),
+                             static_cast<float>(Window->WindowHeight));
+  sheetStarsMask.Texture = textureStarsMask.Submit();
+  sheetStarsMask.IsScreenCap = true;
+  StarsMask =
+      Sprite(sheetStarsMask, 0, 0, Profile::DesignWidth, Profile::DesignHeight);
```
PringlesGang, line 25: Would this fail if you resize your window after initialization?

PringlesGang (changes requested):

PringlesGang:
Do we wanna bother having `SW_EYECATCH_COUNT` control an `Animation` like we do for the menus, for higher framerates?

KKhanhH:
> Do we wanna bother having `SW_EYECATCH_COUNT` control an `Animation` like we do for the menus, for higher framerates?

I don't think it's worth it. These effects only show up like once or twice and it's more coupling logic to use Animation

src/effects/chlcc/eyecatch.cpp
KKhanhH: Yeah this is what they also do

src/effects/chlcc/eyecatch.cpp
KKhanhH: yeah, we could probably do something like recreate the texture on capturescreencap maybe if the dimensions are different?

src/effects/chlcc/bubbleseffect.cpp
PringlesGang: The way you've changed it now, only the initial position takes the DesignDimensions into account. The initial random offset is not yet scaled properly, nor is the velocity in `Update` scaled. I think it's easiest you just leave all the internal code in 720p and then only scale the `dest` rect in `Render`

KKhanhH: Why does that scaling even matter anyway? We know it's going to be 720p design dimensions

KKhanhH: Oh forgot about switch assets

PR #367, opened 2026-01-05
Blur effect

### LINKED ISSUE #355: Blur Effect
Script: AA05 when Takumi, Rimi and Yua search for Seira

What it's supposed to look like:
Image

newo-2001:
Implements the blur effect present in C;H LCC.

Fixes #355.

The names of the ScrWork variables are not canonical names, I was unable to find those.

> [!NOTE]
> The blur is rendered to a texture of size `DesignWidth * DesignHeight`, this should really match the window size, but that can change dynamically at run-time. Doing this properly would require a system to re-submit textures on a window-resize event, which is not currently in place and exceeds the scope of this PR.


profiles/common/scriptvars.lua
```diff
@@ -57,6 +57,10 @@ root.ScriptVars = {
     SW_CAP1LY_OFS=2286,
     SW_CAP1ROTZ_OFS=2287,
     SW_CAP1ALPHA_OFS=2288,
+    SW_BLUR_PRI=2351,
```
KKhanhH: I think they use SW_FEATHERING and SW_FEATHERING_PRI. Also these should probably go into CHLCC's scriptvars, for common we can use the cclcc equivalent

src/effects/blur.cpp
```diff
@@ -0,0 +1,45 @@
...
+                         static_cast<int>(Profile::DesignHeight), 0);
+
+  BlurSprite.Sheet.IsScreenCap = true;
+  BlurSprite.Sheet.Texture = texture.Submit();
+}
+
+BlurEffect::~BlurEffect() { Renderer->FreeTexture(BlurSprite.Sheet.Texture); }
+
+void BlurEffect::Render(int iterations) {
+  if (Profile::Vm::GameInstructionSet != +Vm::InstructionSet::CHLCC) return;
```
KKhanhH: we probably don't need to limit this to just chlcc

profiles/common/scriptvars.lua
KKhanhH: if that works maybe we might also get cclcc blur for free

src/effects/blur.h
```diff
@@ -0,0 +1,23 @@
+#include "../spritesheet.h"
+
+namespace Impacto {
+namespace Effects {
+
+class BlurEffect {
+ public:
+  void Render(int iterations);
+
+  static BlurEffect& GetInstance() {
```
KKhanhH: I use getinstance pattern for the gamespecific stuff to make initialization lazy and optional, but i think for blur since it can be common it shouldn't be necessary, an inline global in the header might be okay.

newo-2001: ~~Changed it to a global, it still has to be behind a pointer or `std::optional` because submitting the texture in the constructor requires a graphics context.~~

profiles/common/scriptvars.lua
newo-2001: Looking at the binary I'm convinced that feathering is a different effect, the implementation of those look way different.

KKhanhH: We could take a look at the instances where there is feathering in cclcc and compare it how chlcc blur looms. I think it's more weird for it to have mosaic and stuff in both but not blur

newo-2001: I checked in CC, the feathering effect is also a blur, but it seems to achieve this through repeated downsampling and upsampling, not through a gaussian blur. I've renamed the variables to `SW_FEATHERING` so that we get this blur on cclcc for free, though it might look slightly different. We can always separate them in the future if we feel they look too different.

src/effects/blur.cpp
```diff
@@ -0,0 +1,49 @@
...
+  Texture texture{};
+  texture.LoadSolidColor(static_cast<int>(Profile::DesignWidth),
+                         static_cast<int>(Profile::DesignHeight), 0);
+  BlurSprite.Sheet.Texture = texture.Submit();
+}
+
+BlurEffect::~BlurEffect() { Renderer->FreeTexture(BlurSprite.Sheet.Texture); }
+
+void BlurEffect::Render(int iterations) {
+  iterations = std::min(iterations / 2, 8);
```
PringlesGang: Newer engine revision has `std::min(iterations, 32)`

PR #371, opened 2026-01-12
Automatically generate .cls files for CCLCC

AxeldeWater:
Added note about .cls file requirements for game resources.


AxeldeWater:
I created the gamedata folder with the .cls files inside them. I wonder though, how should we clarify that the movie and script folders still need to be added to the gamedata folder? Could that be commented inside the vfs.lua file?

Enorovan:
> I created the gamedata folder with the .cls files inside them. I wonder though, how should we clarify that the movie and script folders still need to be added to the gamedata folder? Could that be commented inside the vfs.lua file?

There's no need to clarify, the folders come from the game assets as they are.

PR #372, opened 2026-01-14
Fix mosaic transition bug on iGPU

AxeldeWater:
Mosaic transition animation bugs on iGPU due to `texture2D` function call. Fixed by using `texture` function call instead.


AxeldeWater:
Bad PR

PR #374, opened 2026-01-15
CHLCC: Updated musicmenu to work with BGM Flags. Issue 337.

Quessian-Phaedrus-Sternabite:
Changed playlist at the end to have 70 instead of 71 in extramenus.lua. Changed bgm checking to work with internal flags, not arbitrary ids.

### DIFF
--- a/src/games/chlcc/musicmenu.cpp
+++ b/src/games/chlcc/musicmenu.cpp
@@ -513,7 +513,7 @@ void MusicMenu::UpdateEntries() {
     auto button = static_cast<Widgets::CHLCC::TrackSelectButton*>(
         MainItems->Children[idx]);
 
-    if (!SaveSystem::GetBgmFlag((int)idx)) {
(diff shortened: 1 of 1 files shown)


profiles/chlcc/hud/extramenus.lua
```diff
@@ -237,7 +237,7 @@ root.ExtraMenus = {
         Playlist = {
             8, 14, 26, 10, 28, 12, 20, 16, 18, 24, 22, 80, 76, 32, 34,
             36, 38, 77, 41, 43, 78, 46, 49, 79, 51, 53, 56, 81, 59, 61,
-            63, 65, 82, 83, 84, 0, 1, 2, 4, 6, 85, 72, 74, 75, 71
+            63, 65, 82, 83, 84, 0, 1, 2, 4, 6, 85, 72, 74, 75, 70
```
Gunoshozo: I double checked with the binary and this whole array was correct, so that's not the issue

PR #376, opened 2026-01-15
Simultaneous ADV and NVL text

### LINKED ISSUE #359: NVL and ADV at the same time
Script: AY03

This is about MesSync

https://github.com/user-attachments/assets/554ff410-8f3c-4e1b-9981-d6b5d50075c3

PringlesGang:
Primarily mplements the logic necessary for displaying multiple `DialoguePage`s at the same time

- Properly handle multiple simultaneous `DialoguePage`s
- Implement `InstMessWindow` "Ex" type sub-instructions
- Handle wait icon across multiple `DialoguePage`s
- Skip mode is noticeably faster now

Fixes #359 

I have a much larger `DialoguePage` rework in the works that would fix a lot more stuff related to text handling (e.g. nametag display, backlog entry insertion, backlog entry rendering, reimplementing `InstMes`, etc.) but I figured this PR could still merge separately from that

Reference [this](https://gist.github.com/PringlesGang/b8d622408839f79ab1a23f08d7d19680) for the RE-d `InstMes` and `InstMesMain`

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -501,6 +501,11 @@ void Render() {
             // Dialogue pages drawn in reverse order, at least for cclcc
             for (auto pageIt = DialoguePages.rbegin();
                  pageIt != DialoguePages.rend(); pageIt++) {
+              // TODO: Consistently just use the ScrWork variable internally
+              // after tips and backlog entry pages have been refactored out of
+              // DialoguePage
+              pageIt->Mode = static_cast<DialoguePageMode>(
+                  ScrWork[SW_MESMODE0 + 10 * pageIt->Id]);
               pageIt->Render();
             }
           }
--- a/src/text.h
+++ b/src/text.h
@@ -156,7 +156,7 @@ struct TypewriterEffect : public Animation {
 struct DialoguePage {
   static void Init();
 
-  int Id;
+  int Id = 0;
   int AnimationId = 0;
   int NextAnimationId = 0;
   int CharacterId = -1;


src/vm/inst_dialogue.cpp
```diff
@@ -98,25 +101,25 @@ VmInstruction(InstMesSetID) {
...
+      dialoguePageId = ExpressionEval(thread);
       ImpLogSlow(
           LogLevel::Warning, LogChannel::VMStub,
           "STUB instruction MesSetID(type: SetPage, dialoguePageId: {:d})\n",
           dialoguePageId);
-      thread->DialoguePageId = dialoguePageId;
     } break;
   }
+
+  thread->DialoguePageId = dialoguePageId;
```
KKhanhH, line 122: are we supposed to reset dialogue page id to 0 when we hit messetid case 0?

src/vm/inst_dialogue.cpp
```diff
@@ -359,63 +380,98 @@ VmInstruction(InstMesRev) {
...
+        }
+
+        BlockThread;
       }
-      break;
-    case 2:  // AwaitShowCurrent
-      if (currentPage->FadeAnimation.State == AnimationState::Playing) {
+    } break;
+
+    case 2: {  // OpenedWait
```
KKhanhH: maybe just 
case 2:
case 3:

src/vm/inst_dialogue.cpp
```diff
@@ -359,63 +380,98 @@ VmInstruction(InstMesRev) {
...
-          "STUB instruction MessWindow(type: HideSlow, messWindowId: {:d})\n",
-          messWindowId);
+
+    case 7: {  // FastClose
+      PopExpression(dialoguePageId);
+      thread->DialoguePageId = dialoguePageId;
+
+      SetFlag(dialoguePageId + SF_MESWINDOW0OPENFL, false);
+      ScrWork[dialoguePageId + SW_MESWINDOW0ALPHA] = 0;
+      currentPage->ClearName();
```
KKhanhH: is this supposed to also clearname of dialoguepage of the new dialoguepageid?

src/vm/inst_dialogue.cpp
```diff
@@ -359,63 +380,98 @@ VmInstruction(InstMesRev) {
 }
 VmInstruction(InstMessWindow) {
   StartInstruction;
-  PopUint8(type);
+
   DialoguePage* currentPage = &DialoguePages[thread->DialoguePageId];
```
KKhanhH: instead of a pointer that gets reassigned in some of the cases later, could also do getCurPage to return ref to current dialogue page when its called.

src/vm/inst_dialogue.cpp
PringlesGang: Yes

KKhanhH (approved):

Issue #331, opened 2025-11-25
Dynamically change dock/taskbar icon per game


KKhanhH:
Setting the window icon on Windows and Mac will not change the icon on the dock, this needs a separate system call.

Linux has multiple DEs, so need to check if behavior is different per display servers and environments.
It's possible a desktop file is all that's needed on Linux.

PR #378, opened 2026-01-18
Dock/taskbar icon not changing dynamically per game on Windows fix.

dendi2008:
This PR will fix dock/taskbar icon not changing dynamically per game on Windows. (part of Issue #331).

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -125,6 +125,8 @@ static void Init() {
     Effects::Blur.Init();
   }
 
+  SetWindowIcon(Window->SDLWindow);
+
   if (Profile::GameFeatures & GameFeature::ModelViewer) {
     ModelViewer::Init();
   }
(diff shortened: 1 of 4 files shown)


dendi2008:
> Through some trial and error I've figured out the icon correctly applies only if initialized _after_ `Profile::LoadSpritesheets` (specifically [Line 54 of sprites.cpp](https://github.com/CommitteeOfZero/impacto/blob/2f90def301e4cf28309228a3ec0896f88fc7c381/src/profile/sprites.cpp#L54))

Now it's much cleaner and better, thanks.

> Also still doesn't work on Linux for some reason

Hmmm, I've tested on some Linux DE's and found that, for example, icons work on Manjaro but not on Ubuntu. So, I think, it depends on DE.

KKhanhH:
Yeah on Linux it looks like some DEs don't respect the seticon at all and require a desktop file to do it. Might be x11 vs wayland or kde vs gnome vs unity issue

src/game.cpp
```diff
@@ -105,6 +105,9 @@ static void Init() {
 
   if (Profile::GameFeatures & GameFeature::Renderer2D) {
     Profile::LoadSpritesheets();
+
+    SetWindowIcon(Window->SDLWindow);
+
```
PringlesGang: This isn't really a 2D-renderer-specific thing. I'd put it just after the entire if-block, and then it's all good

dendi2008: Fixed.

PR #383, opened 2026-01-22
Load crash fix

PringlesGang:
Fixes a use after free on `Texture.Buffer` in `Background2D::MainThreadOnLoad` by replacing the raw pointer with a `std::vector`

Also resets background state in `InstMSinit`

### DIFF
--- a/src/games/chlcc/introsequence.cpp
+++ b/src/games/chlcc/introsequence.cpp
@@ -109,7 +109,7 @@ void IntroSequence::Reset() {
                                          0);
   FallingStarsMask.Sheet.Texture = fallingStarsMaskTexture.Submit();
 
-  IntroAnimation.StartIn(true);
+  IntroAnimation.Reset();
 }
 
 void IntroSequence::Update(float dt) {
--- a/src/games/chlcc/titlemenu.cpp
+++ b/src/games/chlcc/titlemenu.cpp
@@ -274,6 +274,12 @@ void TitleMenu::Hide() {
   }
 }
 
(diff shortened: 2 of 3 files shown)


src/texture/texture.cpp
```diff
@@ -60,25 +63,19 @@ void Texture::Load1x1(uint8_t red, uint8_t green, uint8_t blue, uint8_t alpha) {
...
 void Texture::LoadSolidColor(int width, int height, uint32_t color) {
   Init(TexFmt_RGBA, width, height);
-  uint32_t* out = (uint32_t*)Buffer;
-  for (int y = 0; y < Height; y++) {
-    for (int x = 0; x < Width; x++) {
-      *out = color;
-      out++;
-    }
+
+  uint32_t* out = reinterpret_cast<uint32_t*>(Buffer.data());
```
KKhanhH: This is actually UB, we should use bit_cast to an 4 byte array and memcopy it to the buffer. Also not sure if endianess breaks this

PringlesGang: You mean something like this?
```cpp
const uint8_t[4] colorArray = std::bit_cast<const uint8_t*>(&color);
for (size_t i = 0; i < Buffer.size(); i += 4) {
  std::copy(colorArray, colorArray + 4, Buffer.begin() + i);
}
```

KKhanhH: It would need to be an aggregate like std array but yeah

KKhanhH (approved):

PR #384, opened 2026-01-22
Save Entries Refresh Info Fix

Gunoshozo:
- Lock updates during saving
- Remove redundant MainItems field 🐫

### DIFF
--- a/src/games/chlcc/savemenu.cpp
+++ b/src/games/chlcc/savemenu.cpp
@@ -60,8 +60,8 @@ SaveMenu::SaveMenu() : UI::SaveMenu() {
   // Quick Save Pages initialization
 
   for (int i = 0; i < Pages; i++) {
-    MainItems = new Widgets::Group(this);
-    MainItems->WrapFocus = false;
+    Widgets::Group* mainItems = new Widgets::Group(this);
+    mainItems->WrapFocus = false;
 
     for (int j = 0; j < EntriesPerPage; j++) {
       SaveEntryButton* saveEntryButton =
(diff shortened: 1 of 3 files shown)


src/games/chlcc/savemenu.cpp
```diff
@@ -69,7 +70,7 @@ SaveMenu::SaveMenu() : UI::SaveMenu() {
   // Quick Save Pages initialization
 
   for (int i = 0; i < Pages; i++) {
-    MainItems = new Widgets::Group(this);
+    Widgets::Group* MainItems = new Widgets::Group(this);
```
PringlesGang: camelCase

src/games/chlcc/savemenu.cpp
```diff
@@ -129,7 +130,7 @@ SaveMenu::SaveMenu() : UI::SaveMenu() {
   // Full Save Pages initialization
 
   for (int i = 0; i < Pages; i++) {
-    MainItems = new Widgets::Group(this);
+    Widgets::Group* MainItems = new Widgets::Group(this);
```
PringlesGang: camelCase

PringlesGang (commented):
I think you just shouldn't be able to change the `CurrentlyFocusedElement` as it is saving period. Does an early return in `UpdateInput` and a guard for `SavePages->at(*CurrentPage)->Update(dt)` on `GetFlag(SF_SAVEICON)` or `SaveSystem::GetLoadStatus()` work as well?
If it doesn't, then this is perfectly fine

PR #386, opened 2026-01-29
Dock/taskbar icon not changing dynamically per game on some Linux DE's fix.

dendi2008:
This PR will fix dock/taskbar icon not changing dynamically per game on some Linux DE's. (added support for .desktop files.) (Issue #331). 
Example on Ubuntu:
{7628BB6F-DE12-4907-A998-95E40B3B0C07}

Example on Manjaro:
{8C8FC4C5-C0F7-400C-AF8B-76877BAEF4B5}

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,7 @@ set(Impacto_Src
         src/util.cpp
         src/workqueue.cpp
         src/game.cpp
+        src/linuxdesktop.cpp
         src/mem.cpp
         src/modelviewer.cpp
         src/characterviewer.cpp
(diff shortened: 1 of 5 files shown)


KKhanhH:
IMO this is kind of out of scope for impacto, it would make more sense to install desktop files and icon files in the installer part of the patch.

KKhanhH:
> > it would make more sense to install desktop files and icon files in the installer part of the patch.
> 
> Yeah, I thought about it too. I just don't know if it's possible to get icon path or impacto path using installer (for Icon= and Path= in .desktop file).

I think the point would be that the installer would know and generate it with the correct paths. Ofc if you change it afterwards then no guarantees.
Also wondering if it would make more sense in that case to just set the class hints inside the desktop file as well since its just env vars and the desktop file would know what its own appid is.

PR #387, opened 2026-02-04
Break up the giant text file into multiple files in text folder

KKhanhH:
Should be no behavioral changes. Half the changes are probably adding audiosystem imports


src/text/dialoguepage.h
```diff
@@ -0,0 +1,108 @@
...
+
+namespace Impacto {
+enum DialoguePageMode : uint8_t {
+  DPM_ADV = 0,
+  DPM_NVL = 1,
+  DPM_REV = 2,
+  DPM_TIPS = 3
+};
+
+int constexpr DialogueMaxNameLength = 64;
```
PringlesGang: This can move into the cpp

src/text/text.h
```diff
@@ -95,14 +92,6 @@ struct ProcessedTextGlyph {
...
 
-enum DialoguePageMode : uint8_t {
-  DPM_ADV = 0,
-  DPM_NVL = 1,
-  DPM_REV = 2,
-  DPM_TIPS = 3
-};
-
-int constexpr DialogueMaxNameLength = 64;
 int constexpr DialogueMaxRubyChunks = 32;
```
PringlesGang: This one can just go altogether, since it's only used as the length of one `std::array`

PR #394, opened 2026-02-15
Bgm unlocks

Gunoshozo:
- Use proper Ids for music menu
- Implement music menu Init logic for CHLCC (BGM flags codepend on other BGM flags + on ending flags)


src/games/chlcc/musicmenu.cpp
```diff
@@ -141,6 +141,32 @@ MusicMenu::MusicMenu() {
...
+  if (GetFlag(SF_CLR_AYASE)) {
+    SaveSystem::SetBgmFlag(AyaseEndingBgmId, true);
+  }
+
+  const int endingFlags[7] = {SF_CLR_RIMI, SF_CLR_NANAMI, SF_CLR_YUA,
+                              SF_CLR_MIA,  SF_CLR_SENA,   SF_CLR_KOZUE,
+                              SF_CLR_SEIRA};
+  const bool gotRequiredEndings =
+      std::all_of(std::begin(endingFlags), std::end(endingFlags),
+                  [](int flag) { return GetFlag(flag); });
```
PringlesGang:
```cpp
const std::array<int, 7> endingFlags = [...];
const bool gotRequiredEndings = std::ranges::all_of(endingFlags,
        [](int flag) { return GetFlag(flag); });
```

PringlesGang (approved):

PR #395, opened 2026-02-15
[DialoguePage Refactor Step 2] Nametag refactor 

PringlesGang:
Refactors the whole nametag system for more extensibility, maintainability, and readability

Makes use of an abstract base class approach for the `NametagDisplay`
Similarly overhauls the `DialogueBox` class to make better use of the new nametag system

Adds ownership from `DialoguePage` -> `DialogueBox` -> `NametagDisplay`, as each `DialoguePage` in the MAGES. engine maintains its own textbox state
(I am aware this ownership unnecessarily allocated `DialogueBox`es for every backlog and tip entry. I pinky-promise I will finally separate these two from `DialoguePage` next refactor!)

I also cleaned up the `DialoguePage` class a bit with how it interacts with the names
Game-specific hud element implementations have also been moved from `src/games/game-name` to `src/hud/game-name`

`src/hud/nametagdisplay` and `src/hud/dialoguebox` have been completely rewritten, so I suggest reviewing those as effectively "new files," instead of looking at the git diff

Definitely also let me know if you disagree with this overall design, or if you have some better ways (part of) this could be structured


src/profile/dialogue.h
```diff
@@ -39,11 +39,10 @@ inline float ADVNameFontSize;
 // e.g. if ADVNameAlignment == TextAlignment::Right, name will *end* at
 // ADVNamePos.x
 inline glm::vec2 ADVNamePos;
-inline bool ADVBoxShowName = true;
 
 // FadeInPauseOut - CCLCC, Instant - CHLCC
 // Other games may have different behavior, implement as needed
-BETTER_ENUM(NametagType, int, None, Instant, FadeInPauseOut)
+BETTER_ENUM(NametagType, int, None, Sprite, TwoPiece, ThreePiece, CHLCC, CC)
```
Gunoshozo: Old comments :p

profiles/dash/dialogue.lua
```diff
@@ -92,4 +101,20 @@ MakeAnimation({
     SecondaryDirection = AnimationDirections.Down
 });
 
-include('dash/nametag.lua');
\ No newline at end of file
+root.Sprites["NametagLeftSprite"] = {
+    Sheet = "Menu",
+    Bounds = { X = 3207, Y = 1097, Width = 168, Height = 65 },
+    BaseScale = { X = 0.6307692307692308, Y = 0.6307692307692308 }
```
Gunoshozo: 41/65? That's a minor thing, quite old and unrelated, but I'd prefer random fractions over random floats for scale

src/hud/chlcc/dialoguebox.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+void DialogueBox::Render(const DialoguePageMode mode, const NameInfo nameInfo,
+                         glm::vec4 tint) {
+  switch (mode) {
+    case DPM_ADV: {
+      tint = {1.0f, 1.0f, 1.0f, tint.a};
+      const Sprite advBoxSprite = ScrWork[SW_MESWINDOW_COLOR] == 1
+                                      ? SecondaryADVBoxSprite
+                                      : ADVBoxSprite;
+      // On subsequent calls it's equal to 0 and *acts* as a boolean again.
```
KKhanhH: this comment is no longer relevant here

src/hud/mo6tw/dialoguebox.cpp
```diff
@@ -0,0 +1,39 @@
...
+  if (mode == DPM_ADV) {
+    Renderer->DrawSprite(ADVBoxPartLeft, ADVBoxPartLeftPos, tint);
+    Renderer->DrawSprite(ADVBoxSprite, ADVBoxPos, tint);
+    Renderer->DrawSprite(ADVBoxPartRight, ADVBoxPartRightPos, tint);
+    Renderer->DrawSprite(ADVBoxDecoration, ADVBoxDecorationPos, tint);
+
+    NametagDisplayInst->Render(nameInfo, tint);
+
+  } else {
+    assert(mode == DPM_NVL);
```
KKhanhH: why not just else if mode == DPM_NVL

src/hud/nametagdisplay.cpp
```diff
@@ -1,165 +1,93 @@
...
-    }
-    default: {
-      return 1.0f;
+void ThreePieceNametagDisplay::Render(const NameInfo nameInfo,
+                                      const glm::vec4 tint) {
+  if (Hidden || !nameInfo.RenderWindow) return;
+
+  Renderer->DrawSprite(NametagLeftSprite, NametagPosition, tint);
+
+  const RectF nameBounds = std::accumulate(
```
KKhanhH, line 64: Might be useful to expose some more stuff to the line glyph creator in the future so that it gives a coalesced bounds as it build it, so then we can get stuff like widths or start/end/center points easily.

src/hud/nametagdisplay.h
```diff
@@ -1,39 +1,50 @@
...
+  std::optional<int> NameId;
+  std::span<ProcessedTextGlyph> Name;
+};
+
+class NametagDisplay {
+ public:
+  static std::unique_ptr<NametagDisplay> Create();
+  virtual ~NametagDisplay() = default;
+
+  virtual void Render(NameInfo nameInfo, glm::vec4 tint) = 0;
```
KKhanhH: You should probably make nameinfo a reference, it is slightly big at 32bytes

src/text/dialoguepage.h
```diff
@@ -80,23 +76,23 @@ struct DialoguePage {
   void Render();
   void Hide();
   void Show();
-  bool HasName() { return this->NameId.has_value(); }
+  bool HasName() { return !Name.empty(); }
```
KKhanhH: const

src/text/text.cpp
```diff
@@ -468,11 +468,11 @@ void InitNamePlateData(Vm::Sc3Stream& stream) {
   } while (stream.PeekU16() != 0xFFFF);
 }
 
-uint32_t GetNameId(uint8_t* name, int nameLength) {
+std::optional<uint32_t> GetNameId(uint8_t* name, int nameLength) {
```
KKhanhH: maybe make this a span while you're at it

src/hud/nametagdisplay.cpp
PringlesGang: What do you mean exactly? That we can set in the profile whether the "line" displays above/below/inside the text?

KKhanhH: Oh no I'm not asking for a change here. I'm thinking it might be helpful to change text glyph layout or whatever later so that it returns the accumulated rect like here as well, so it wouldnt need recalculation here. Would be useful for stuff like figuring out margins and bounds i think

PringlesGang: Ohh yeah that does sound useful

PR #399, opened 2026-02-18
date-formats: make saveentries date formats configurable

Enorovan:


src/profile/game.cpp
```diff
@@ -29,6 +29,8 @@ void LoadGameFromLua() {
   Subtitles = EnsureGetMember<char const*>("Subtitles");
   CloseBacklogWhenReachedEnd =
       TryGetMember<bool>("CloseBacklogWhenReachedEnd").value_or(true);
+  DateFormat = DateFormatType::_from_integral_unchecked(
+      EnsureGetMember<uint8_t>("DateFormat"));
```
PringlesGang: I think such a config works best as
`TryGetMember<uint8_t>("DateFormat").value_or(+DateFormatType::YMD)`
such that it is not _required_ for the game to launch, but you can override it if you want

PR #401, opened 2026-02-22
[CCLCC] [CHLCC] Cleanup and bugfixes.

dendi2008:
This PR will:

- Delete no more needed (in some cases) `target->Hovered = false`
- Fix bug when user can hover buttons from TitleMenu after starting new game in CHLCC.

### DIFF
--- a/src/games/chlcc/titlemenu.cpp
+++ b/src/games/chlcc/titlemenu.cpp
@@ -32,7 +32,6 @@ void TitleMenu::MenuButtonOnClick(Widgets::Button* target) {
 }
 
 void TitleMenu::SecondaryButtonOnClick(Widgets::Button* target) {
-  target->Hovered = false;
   ScrWork[SW_TITLECUR2] = target->Id;
   ChoiceMade = true;
 }
@@ -292,32 +291,31 @@ void TitleMenu::Update(float dt) {
 
   if (GetFlag(SF_TITLEMODE)) {
     Show();
-  } else if (ScrWork[SW_TITLECT] < 32) {
-    ItemsFadeInAnimation.StartOut();
-    SecondaryItemsFadeInAnimation.StartOut();
+  } else if (State == Shown && !IsFocused) {
+    // when loading/starting a game from a submenu
     Hide();
   }
 
   if (State == Shown && IsFocused) {
     MainItems->Tint.a =
         glm::smoothstep(0.0f, 1.0f, PrimaryFadeAnimation.Progress);
-    MainItems->Update(dt);
     MainItems->UpdateInput(dt);
+    MainItems->Update(dt);
     LoadItems->Tint.a =
         glm::smoothstep(0.0f, 1.0f, SecondaryFadeAnimation.Progress);
-    LoadItems->Update(dt);
     LoadItems->UpdateInput(dt);
+    LoadItems->Update(dt);
     CurrentExtraItems->Tint.a =
         glm::smoothstep(0.0f, 1.0f, SecondaryFadeAnimation.Progress);
+    CurrentExtraItems->UpdateInput(dt);
     CurrentExtraItems->Update(dt);
-    SystemItems->UpdateInput(dt);
     SystemItems->Tint.a =
         glm::smoothstep(0.0f, 1.0f, SecondaryFadeAnimation.Progress);
+    SystemItems->UpdateInput(dt);
     SystemItems->Update(dt);
-    CurrentExtraItems->UpdateInput(dt);
 
-    switch (ScrWork[SW_TITLEDISPCT]) {
-      case 0: {
+    switch (static_cast<TitleDispCtState>(ScrWork[SW_TITLEDISPCT])) {
+      case TitleDispCtState::IntroAnimation: {
         if (IntroSequence.IntroAnimation.IsIn() && ScrWork[SW_TITLECT] == 0) {
           ResetIntroSequence();
         }
@@ -348,12 +346,18 @@ void TitleMenu::Update(float dt) {
           LoadItems->Hide();
         }
       } break;
-      case 1: {
+      case TitleDispCtState::PressStart: {
+        if (!GetFlag(SF_TITLEMODE)) {
+          ItemsFadeInAnimation.StartOut();
+          SecondaryItemsFadeInAnimation.StartOut();
+          Hide();
+          break;
+        }
         if (PressToStartAnimation.State == AnimationState::Stopped) {
           PressToStartAnimation.StartIn();
         }
       } break;
-      case 3: {  // Main Menu Fade In
+      case TitleDispCtState::MainEntriesFading: {
         if (MainItems->VisibilityState == Hidden && ScrWork[SW_TITLECT] == 0) {
           MainItems->Show();
           MainItems->Tint.a = 0.0f;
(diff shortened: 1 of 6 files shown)


src/games/chlcc/titlemenu.cpp
```diff
@@ -290,29 +289,31 @@ void TitleMenu::Update(float dt) {
 
   if (GetFlag(SF_TITLEMODE)) {
     Show();
-  } else if (ScrWork[SW_TITLECT] < 32) {
-    ItemsFadeInAnimation.StartOut();
-    SecondaryItemsFadeInAnimation.StartOut();
+  } else {
+    if (ScrWork[SW_TITLECT] < 32) {
+      ItemsFadeInAnimation.StartOut();
```
Gunoshozo: This one was correct before, hide should be inside `if (ScrWork[SW_TITLECT] < 32) {`
It was fixing the issue with title menu here https://github.com/CommitteeOfZero/impacto/pull/379

dendi2008: Fixed.

src/games/cc/sysmesbox.cpp
```diff
@@ -116,10 +116,10 @@ void SysMesBox::UpdateInput(float dt) {
 
   const auto* const prevSelected = CurrentlyFocusedElement;
   Menu::UpdateInput(dt);
-  ChoiceItems->UpdateInput(dt);
   if (CurrentlyFocusedElement && prevSelected != CurrentlyFocusedElement) {
     Audio::PlayInGroup(Audio::ACG_SE, "sysse", 1, false, 0.0f);
   }
+  ChoiceItems->UpdateInput(dt);
```
Gunoshozo: After a small internal discussion we decided that we actually should keep and add sounds on mouse hover for menus, popups and etc, to make UX in menus a bit nicer.
Generally if keyboard and gamepad causes a sound so a mouse should do it too.

src/games/chlcc/titlemenu.cpp
```diff
@@ -298,23 +297,23 @@ void TitleMenu::Update(float dt) {
     Hide();
   }
 
-  if (State == Shown && IsFocused) {
+  if (State == Shown && IsFocused && GetFlag(SF_TITLEMODE)) {
```
Gunoshozo: I'm not sure if that's the proper solution here, the issue here is that `ScrWork[SW_TITLECT] < 32` is not triggered reliably for some reason when starting a new game, so a title menu is not hidden 100% of the time.

For me behavior of `ScrWork[SW_TITLECT]` varies between sessions and between new game starts during same session. Sometimes it's set to 0 and sometimes it's still at 32.

Couldn't narrow it down yet, would be cool if you could figure it out. But I'm going to read game scripts anyway once I get the chance

Gunoshozo (changes requested):
Deleted a comment here

I mistook it for something else

src/games/cc/sysmesbox.cpp
dendi2008: Reverted.

src/games/chlcc/titlemenu.cpp
dendi2008: I will see what I can do, but no promises.

dendi2008: Maybe, instead of using `ScrWork[SW_TITLECT] < 32` we should use `ScrWork[SW_TITLEDISPCT] <= 4` ? It worked for me, but it still might be not proper solution.

Gunoshozo: Hmm, it seems to work on the first start only. When you go back to the title screen and start a new game, title menu is not hiding anymore

dendi2008: > Hmm, it seems to work on the first start only. When you go back to the title screen and start a new game, title menu is not hiding anymore

Weird. I've tested now (on local build). And title screen hidding  even after second start. You removed `&& GetFlag(SF_TITLEMODE` (if you had it) and edited `ScrWork[SW_TITLECT] < 32` to `ScrWork[SW_TITLEDISPCT] <= 4`, right?

dendi2008: > Weird. I've tested now. And title screen hidding even after second start.

I've also tested it on CI build and it works. Also even after second start. Could you write me steps that you did? I want to to try reproduce it exactly like you did.

Gunoshozo: Oh, It's cause I didn't notice that you have also changed `SW_TITLECT` to `SW_TITLEDISPCT`, my bad.
This seems to work better, but I'm not 100% sure if that's it
I'll ask other guys what they think about it

PringlesGang: `SW_TITLEDISPCT` is an enum for the menu's current state. This is what I've managed to figure out after some quick script RE'ing:
```cpp
enum class TitleDispCtState : uint8_t {
  IntroAnimation = 0,
  PressStart = 1,
  EmptyBackground = 2,
  MainEntriesFading = 3,
  MainEntriesControl = 4,
  // 5 and 6 go unused
  LoadSubEntriesFading = 7,
  LoadSubEntriesControl = 8,
  ExtraSubEntriesFading = 9,
  ExtraSubEntriesControl = 10,
  SystemSubEntriesFading = 11,
  SystemSubEntriesControl = 12,
};
```
As you can see, `ScrWork[SW_TITLEDISPCT] < 4` is not exactly what you want. From what I can tell, this variable is not even reset when starting a new game or loading a save.

Looks like `SF_TITLEMODE` _does_ seem to reset in those instances, but maybe also others; I haven't checked all that closely for this one.

`SW_TITLECT` is indeed the animation count variable, but its usage and bounds depend on `SW_TITLEDISPCT`, so perhaps something like
```cpp
if (GetFlag(SF_TITLEMODE)) {
  Show();
else if ((ScrWork[SW_TITLEDISPCT] == MainEntriesControl ||
          ScrWork[SW_TITLEDISPCT] == LoadSubEntriesControl) &&
         ScrWork[SW_TITLECT] < 32)
  ItemsFadeInAnimation.StartOut();
  SecondaryItemsFadeInAnimation.StartOut();
  Hide();
}
```
might work? Though frankly at that point I think this logic should move into the switch statement further below.

dendi2008: Maybe, we could add `if !GetFlag(SF_TITLEMODE) {` to switch `(ScrWork[SW_TITLEDISPCT])` cases and remove this `else if`?

PringlesGang: Yeah, that's a good idea
(Also, if you could, adding the `TitleDispCtState` enum class to the cpp file and having all comparisons to `ScrWork[SW_TITLEDISPCT]` compare against those would help a lot with readability, thanks :))

dendi2008: > Yeah, that's a good idea (Also, if you could, adding the `TitleDispCtState` enum class to the cpp file and having all comparisons to `ScrWork[SW_TITLEDISPCT]` compare against those would help a lot with readability, thanks :))

Done.

src/games/chlcc/titlemenu.cpp
```diff
@@ -348,12 +343,24 @@ void TitleMenu::Update(float dt) {
...
         }
       } break;
-      case 3: {  // Main Menu Fade In
+      case TitleDispCtState::MainEntriesFading: {
+        if (!GetFlag(SF_TITLEMODE)) {
+          ItemsFadeInAnimation.StartOut();
+          SecondaryItemsFadeInAnimation.StartOut();
+          Hide();
+          break;
+        }
```
Gunoshozo: This state is after you click any button while game prompts you to "click start", so the Main buttons are Fading In. Probably need to go under `TitleDispCtState::MainEntriesControl`

dendi2008: Done.

Issue #352, opened 2025-12-26
There's a CG Shatter VFX when you see the Rimi CG where she's sitting under a spotlight


PringlesGang:
Script: AA04, near the middle of the script, Rimi reminisces about herself

https://github.com/user-attachments/assets/b1ba3ede-a919-4b8a-bbd2-e660df4a01c2

PR #403, opened 2026-02-24
Background2D RenderExplode

PringlesGang:
Implements the exploding background vfx (#352)


src/background2d.cpp
```diff
@@ -815,6 +817,177 @@ void Background2D::RenderFade() {
...
+      renderDimensions / glm::vec2(ExplodeGridWidth, ExplodeGridHeight);
+
+  // Initialize gridpoint positions
+  std::array<glm::vec2, (ExplodeGridWidth + 1) * (ExplodeGridHeight + 1)> grid;
+
+  for (size_t y = 0; y <= ExplodeGridHeight; y++) {
+    for (size_t x = 0; x <= ExplodeGridWidth; x++) {
+      const glm::ivec2 randomOffsetLimits =
+          glm::ivec2(cellDimensions + glm::vec2(0.5f)) * 10;
+      const glm::vec2 randomOffset = glm::vec2(CALCrnd(randomOffsetLimits.x),
```
KKhanhH: put the calcrnd in a separate variable first, function eval order is kinda nondeterministic.

src/background2d.cpp
```diff
@@ -815,6 +817,177 @@ void Background2D::RenderFade() {
...
+        const glm::vec2 center = (topLeft + bottomRight) / 2.0f;
+        tri.InitialDisplayPosition = center;
+        std::ranges::transform(
+            tri.VertexOffsets, tri.VertexOffsets.begin(),
+            [center](glm::vec2 pos) { return pos - center; });
+
+        // TranslationSpeed is not based on design dimensions in the
+        // MAGES. engine either
+        tri.TranslationSpeed =
+            glm::vec2(CALCrnd(1200), CALCrnd(1200)) / 100.0f - glm::vec2(6.0f);
```
KKhanhH: same here

src/background2d.cpp
```diff
@@ -815,6 +817,177 @@ void Background2D::RenderFade() {
...
+  if (lastFadeCount == fadeCount) return;
+  lastFadeCount = fadeCount;
+
+  bool anyTriVisible = false;
+  for (ExplodeTri& tri : ExplodeTris) {
+    tri.DisplayPosition =
+        tri.InitialDisplayPosition + tri.TranslationSpeed * fadeCount;
+    tri.Rotation = glm::pow(tri.RotationSpeed, fadeCount);
+
+    [[maybe_unused]] constexpr float fadeCountMax = 64.0f;
```
KKhanhH: constexpr vars inside a function should be made static

src/background2d.cpp
PringlesGang: Does the order matter though? Isn't CALCrnd practically IID?

KKhanhH: it modifies the internal state of the randomizer, so i'd rather not risk it

PringlesGang: Well yeah, but regardless of the state inside, the drawing of the random variable is still uniformly distributed in the bounds provided. It alters the internal state, but not the semantics of the outcome; it offers a different result, but so could calling it a millisecond later

KKhanhH: non deterministic behavior and something that differs across compilers is something we should avoid. also if we wanted to test random with a different seed or something too.

PringlesGang: I suppose with your second argument you've got a point. With your first I generally agree, but in this particular case you're already explicitly giving yourself up for two IID stochastic draws, meaning their order mathematically doesn't matter—just like two independent statements being placed out-of-order by a compiler for performance doesn't matter for the execution of the application
I'll change it based on your second argument tho

src/background2d.cpp
```diff
@@ -815,6 +817,177 @@ void Background2D::RenderFade() {
...
+    constexpr float fadeOutDuration = 32.0f;
+    const float fadeOutProgress =
+        (fadeCount - tri.StartFadeOutTime) / fadeOutDuration;
+    tri.Alpha = std::clamp(1.0f - fadeOutProgress, 0.0f, 1.0f);
+    assert(tri.StartFadeOutTime + fadeOutDuration <= fadeCountMax);
```
Gunoshozo: Maybe this assert should be moved to the `ResetExplodeTris`?
Since it's called less frequently than `UpdateExplode` and it's where `tri.StartFadeOutTime` is calculated

src/background2d.cpp
```diff
@@ -815,6 +817,177 @@ void Background2D::RenderFade() {
...
+                             grid[(y + 1) * (ExplodeGridWidth + 1) + x]};
+
+        // Set the UVs for each vertex by "overlaying" the vertex grid onto the
+        // sprite's texture
+        const auto getSpritePosition = [&](glm::vec2 gridPosition) {
+          const glm::vec2 texturePos = gridPosition / renderSprite.BaseScale +
+                                       renderSprite.Bounds.GetPos();
+          const glm::vec2 normalizedTexturePos =
+              texturePos / renderSprite.Sheet.GetDimensions();
+          return normalizedTexturePos;
```
Gunoshozo: `renderSprite.Bounds.GetPos()` and `renderSprite.Sheet.GetDimensions()` are pretty much constant inside `ResetExplodeTris` context but each issues a call `ExplodeGridHeight * ExplodeGridWidth * 3 * 3` times, 
\+ I'm not sure It will be optimized out by the compiler since it's hidden inside anonymous function that is called via std::range.

Gunoshozo (approved):

PR #412, opened 2026-02-28
game-exit

Enorovan:


src/games/cclcc/titlemenu.cpp
```diff
@@ -443,6 +456,9 @@ void TitleMenu::ReturnToMenuUpdate() {
 }
 
 void TitleMenu::MainMenuUpdate() {
+  if (IsFocused) {
+    MainItems->HasFocus = true;
```
KKhanhH, line 460: Does this break anything, like moving between submenus or whatnot?

Enorovan: Branch has been stuck for so long I forgot why I even added this tbh, but it at least doesn't *break* anything.

KKhanhH: Does it break gameexit if you remove it

Enorovan: Saying no to game exit doesn't give back focus to main items. Breaks nothing else when in.

src/vm/inst_controlflow.cpp
```diff
@@ -182,7 +182,10 @@ VmInstruction(InstKeyOnJump) {
   const InputDownType downType = static_cast<InputDownType>(downTypeId);
 
   const bool inputResult = [&]() -> bool {
-    const uint32_t bitfield = (type & 2) ? PADcustom[buttonsArg] : buttonsArg;
+    const uint32_t bitfield = (type & 2) == 0 ? buttonsArg
+                              : buttonsArg < std::ssize(PADcustom)
+                                  ? PADcustom[buttonsArg]
+                                  : false;
```
PringlesGang: You're setting a uint32_t to false. Not UB, but still a little weird

PR #414, opened 2026-03-01
Conditional Titlemenu Exit Buttons

Enorovan:


src/games/cclcc/titlemenu.cpp
```diff
@@ -62,6 +62,8 @@ void TitleMenu::ExtraButtonOnClick(Widgets::Button* target) {
   ShowExtraItems();
 }
 
+void TitleMenu::ExitButtonOnClick() { Game::Shutdown(); }
```
KKhanhH: set Game::ShouldQuit to true instead

src/profile/game.h
```diff
@@ -64,6 +64,7 @@ inline bool Fullscreen;
 inline char const* Subtitles;
 inline bool CloseBacklogWhenReachedEnd = true;
 inline DateFormatDef DateFormat = +DateFormatType::YMD;
+inline bool HasTitleMenuExitButton = false;
```
PringlesGang, line 68: I don't think this should pollute global profile scope. There are definitely some games where the exit button should _not_ be optional. I'd prefer this in the game-specific title menu profile files

src/profile/games/chlcc/titlemenu.cpp
```diff
@@ -14,6 +14,13 @@ namespace CHLCC {
 namespace TitleMenu {
 
 void Configure() {
+  auto nullSprite = Sprite();
+  nullSprite.Bounds = RectF(0.0f, 0.0f, 0.0f, 0.0f);
+  ExitSprite = nullSprite;
+  TryGetMember<Sprite>("ExitSprite", ExitSprite);
+  ExitHighlightSprite = nullSprite;
+  TryGetMember<Sprite>("ExitHighlightSprite", ExitHighlightSprite);
```
PringlesGang: Here you should first `TryGetMember<bool>("HasTitleMenuExitButton").value_or(false)`, and then if the value is `true`, you should `EnsureGetMember` the two sprites

src/profile/games/cclcc/titlemenu.h
```diff
@@ -26,6 +26,7 @@ inline Sprite LibrarySprite;
 inline Sprite LibraryHighlightSprite;
 inline Sprite EndingListSprite;
 inline Sprite EndingListHighlightSprite;
+inline Sprite ExitSprite;
```
PringlesGang, line 29: Here, because it's only a single sprite that needs setting, you could even just keep it a `std::optional<Sprite> ExitSprite`, and use that as the `HasTitleMenuExitButton` bool

src/games/chlcc/titlemenu.cpp
```diff
@@ -88,6 +90,18 @@ TitleMenu::TitleMenu() {
...
 
+  // Exit menu button (Configuration/Patch driven)
+  if (HasTitleMenuExitButton) {
+    Exit = new TitleButton(
+        4, ExitSprite, ExitHighlightSprite, ItemHighlightSprite,
+        glm::vec2(
+            ((ItemHighlightOffset.x * ItemsFadeInAnimation.Progress) - 1.0f) +
+                ItemHighlightOffset.x,
+            ((ItemYBase - 1.0f) + (4 * ItemPadding))));
+    Exit->OnClickHandler = [](auto*) { return ExitButtonOnClick(); };
```
PringlesGang: This doesn't need a class method, this could just be a straight lambda `[](auto*) { Game::ShouldQuit = true; }`

src/profile/games/cclcc/titlemenu.h
Enorovan: Could I really, because it will always be in C;CLCC's profile,

PringlesGang: Then take it out of the profile :p
This is a patch thing, right? It should be in the patch profile

PR #418, opened 2026-03-02
Subtitle support

KKhanhH:
Adds subtitle support to the video player.

Currently only ass subtitles.
Embedded subtitle tracks works, as well as subtitles in separate files.
Should support multiple subtitle tracks at the same time

### DIFF
--- a/src/video/ffmpegplayer.cpp
+++ b/src/video/ffmpegplayer.cpp
@@ -541,7 +541,12 @@ void FFmpegPlayer::Update(float dt) {
   if (IsPlaying) {
     using namespace std::literals::chrono_literals;
     if (AudioStream) AudioPlayer->Process();
-    if (SubPlayer) SubPlayer->Update(MasterClock->Get());
+
+    if (Profile::GameFeatures & GameFeature::Subtitles) {
+      if (SubPlayer) SubPlayer->Update(MasterClock->Get());
+      UpdateSubtitles();
+    }
+
     Clock::Microseconds duration{};
     Clock::MonotonicTime time;
 
@@ -583,7 +588,6 @@ void FFmpegPlayer::Update(float dt) {
                      std::chrono::seconds(fps.getDenominator())) /
                  fps.getNumerator();
     }
-    if (Profile::GameFeatures & GameFeature::Subtitles) UpdateSubtitles();
 
     if (time < FrameTimer + duration) {
       return;


src/profile/game.cpp
```diff
@@ -26,7 +26,9 @@ void LoadGameFromLua() {
   ResolutionWidth = EnsureGetMember<int>("ResolutionWidth");
   ResolutionHeight = EnsureGetMember<int>("ResolutionHeight");
   Fullscreen = EnsureGetMember<bool>("Fullscreen");
-  Subtitles = EnsureGetMember<char const*>("Subtitles");
+  SubtitleConfig = SubtitleConfigType::_from_integral_unchecked(
+      TryGetMember<uint8_t>("SubtitleConfig")
+          .value_or(+SubtitleConfigType::All));
```
PringlesGang, line 31: This is a flag enum, so `Karaoke | Translation` should be a valid value. This value does not, however, have a specifically declared constant in the definition, and `_from_integral_unchecked` has undefined behaviour for non-declared constants

profiles/characterviewer/config.lua
```diff
@@ -2,4 +2,4 @@ root.Language = "Japanese";
 root.ResolutionWidth = 1280;
 root.ResolutionHeight = 720;
 root.Fullscreen = false;
-root.Subtitles = "off";
\ No newline at end of file
+root.SubtitleConfig =
```
PringlesGang: Finish the statement?

profiles/cclcc/subtitle.lua
PringlesGang, line 1: Remove this debug file

src/profile/subtitle.cpp
```diff
@@ -0,0 +1,41 @@
...
+template <>
+struct TryGetImpl<Subtitle::SubtitleTrackFile> {
+  static std::optional<Subtitle::SubtitleTrackFile> Call() {
+    if (!lua_istable(LuaState, -1)) return std::nullopt;
+    auto typeOpt = TryGetMember<uint8_t>("Type");
+    auto idOpt = TryGetMember<int>("Id");
+    auto pathOpt = TryGetMember<std::string>("Path");
+    auto configOpt = TryGetMember<uint8_t>("Config");
+
+    if (!typeOpt || !configOpt || (!pathOpt && !idOpt)) return std::nullopt;
```
PringlesGang: Perhaps also error in the case `pathOpt && idOpt`? I.e. `!(pathOpt ^ idOpt)`

src/profile/subtitle.h
```diff
@@ -0,0 +1,24 @@
...
+  SubtitleType Type;
+  SubtitleConfigType Config;
+  std::optional<int> Id;
+  std::optional<std::string> Path;
+};
+
+using SubtitleTrackFiles = std::vector<SubtitleTrackFile>;
+inline ankerl::unordered_dense::map<std::string, SubtitleTrackFiles>
+    SubtitleMappings;
+inline std::optional<std::string> SubtitleFontsDir;
```
PringlesGang: Not a big fan of having all fonts be in a single location... Not the end of the world, but if you _could_ easily refactor it to take in a vector, that'd be great

src/renderer/dx9/renderer.cpp
```diff
@@ -795,6 +795,10 @@ void Renderer::SetBlendMode(RendererBlendMode blendMode) {
       Device->SetRenderState(D3DRS_SRCBLEND, D3DBLEND_SRCALPHA);
       Device->SetRenderState(D3DRS_DESTBLEND, D3DBLEND_ONE);
       break;
+    case RendererBlendMode::Premultiplied:
+      Device->SetRenderState(D3DRS_SRCBLEND, D3DBLEND_ONE);
+      Device->SetRenderState(D3DRS_DESTBLEND, D3DBLEND_INVSRCALPHA);
+      break;
```
Enorovan, line 801: Just a self reminder to try that on my fog branch

src/subtitle/subtitlesystem.cpp
```diff
@@ -0,0 +1,161 @@
...
+  }
+}
+
+void SubtitlePlayer::AddTrack(int trackId, SubtitleType type, void* trackInfo,
+                              Profile::SubtitleConfigType config) {
+  auto& backend = Backends[GetBackendIndex(type)];
+  if (!backend) {
+    ImpLog(LogLevel::Warning, LogChannel::Subtitle,
+           "Subtitle backend not initialized for {}", type._to_string());
+  }
```
PringlesGang, line 71: This should early return

src/subtitle/ass/subtitlerenderer.cpp
```diff
@@ -0,0 +1,226 @@
...
+          uint8_t& dstB = t.Buffer[idx + 2];
+          uint8_t& dstA = t.Buffer[idx + 3];
+
+          dstR = static_cast<uint8_t>((rr + dstR * (mult - aa)) / mult);
+          dstG = static_cast<uint8_t>((gg + dstG * (mult - aa)) / mult);
+          dstB = static_cast<uint8_t>((bb + dstB * (mult - aa)) / mult);
+          dstA = static_cast<uint8_t>((aa * 255 + dstA * (mult - aa)) / (mult));
+        }
+      }
+    };
```
PringlesGang: Perhaps `glm::u8vec4 (glm::vec<4, uint8_t, defaultp>)` could help reduce the duplication here?

src/subtitle/ass/subtitlerenderer.h
```diff
@@ -0,0 +1,73 @@
...
+    CONTENT = 2
+  };
+  SubtitleRenderTrack(SubtitleRenderer& renderer, std::string_view header);
+  SubtitleRenderTrack(SubtitleRenderer& renderer,
+                      std::span<const char> fileBuffer);
+  void Push(SubtitleEntry entry);
+  void Render();
+  SpriteSheet SubtitlesSpritesheet;
+  std::vector<SubtitleEntry> Entries;
+  SubtitleRenderer* SubRenderer;
```
PringlesGang: Can this not be a `std::reference_wrapper`?

src/subtitle/ass/subtitlerenderer.cpp
```diff
@@ -0,0 +1,226 @@
...
+void SubtitleRenderer::AddTrackFile(int trackId,
+                                    std::span<const char> fileBuffer) {
+  auto [ref, inserted] = Tracks.try_emplace(trackId, *this, fileBuffer);
+  if (!inserted) {
+    ImpLog(LogLevel::Error, LogChannel::Subtitle,
+           "Attempted to insert to existing Track ID {:d}", trackId);
+  }
+}
+
+void SubtitleRenderer::SetTrackVisibility(int trackId, bool enable) {}
```
PringlesGang: Not implemented?

src/subtitle/subtitlerenderer.h
```diff
@@ -0,0 +1,53 @@
...
+  std::chrono::milliseconds StartMs{};
+  std::chrono::milliseconds Duration{};
+};
+
+class SubtitleRenderer {
+ public:
+  SubtitleRenderer(SubtitlePlayer& player, float width, float height)
+      : Player(player), Width(width), Height(height) {}
+  virtual ~SubtitleRenderer() noexcept = default;
+  virtual void AddTrack(int trackId, void* trackInfo) = 0;
```
PringlesGang: Don't quite like the idea of passing along `void*`... If the idea is that any derived classes could take the data in whatever form makes sense for them, then the interface already inherently differs between instantiations, so having it be a pure virtual function doesn't make sense in that case.
I'd just take out the base class declaration and have it be a typed method on the derived class

src/video/ffmpegplayer.cpp
```diff
@@ -198,27 +225,96 @@ void FFmpegPlayer::Play(Io::Stream* stream, bool looping, bool alpha) {
...
+        SubPlayer.emplace(VideoTexture->Width, VideoTexture->Height);
+      else
+        SubPlayer.emplace(Profile::DesignWidth, Profile::DesignHeight);
+    }
+  };
+  auto mappingsItr = SubtitleMappings.find(StreamPtr->Meta.FileName);
+
+  // Tracks embedded in video
+  for (auto& [subtitleStream, id] : embeddedSubStreams) {
+    if (!subtitleStream.isSubtitle() || !subtitleStream.isValid()) continue;
```
PringlesGang: I feel like this should be asserted instead

src/video/ffmpegplayer.cpp
```diff
@@ -429,72 +554,132 @@ void FFmpegPlayer::Update(float dt) {
...
-    const RectF dest = {0.0f, 0.0f, Profile::DesignWidth,
-                        Profile::DesignHeight};
-    const glm::vec4 tint = {1.0f, 1.0f, 1.0f, videoAlpha};
-    Renderer->DrawVideoTexture(*VideoTexture, dest, tint, IsAlpha);
+void FFmpegPlayer::UpdateSubtitles() {
+  for (auto& subtitleStream : EmbeddedSubtitleStreams) {
+    Video::AVDecodedItem<AVMEDIA_TYPE_SUBTITLE> subtitle;
+    const bool hasSubtitle = subtitleStream.FrameQueue.try_dequeue(subtitle);
+    if (!hasSubtitle) {
+      ImpLog(LogLevel::Info, LogChannel::Video, "No Subtitle");
```
PringlesGang: Should this and the other one be `ImpLogSlow`? It's possibly being called every frame

PringlesGang (approved):
To be perfectly honest most of the core logic here went far over my head, but I trust ya :)

src/profile/game.cpp
KKhanhH: well i guess it's dead code now

src/video/ffmpegplayer.cpp
KKhanhH: forgot to remove debug logs

src/subtitle/subtitlerenderer.h
KKhanhH: I didn't wanna require caller expose the backends itself and do all the validation. New approach basically just forwards the args to the subtitlerendertrack constructor, does require defining the interface correctly but should be more type safe now.

src/subtitle/ass/subtitlerenderer.cpp
```diff
@@ -0,0 +1,230 @@
...
+          (img.color >> 16) & 0xFF,
+          (img.color >> 8) & 0xFF,
+      };
+      const uint8_t a = 255 - (img.color & 0xFF);
+      for (int y = 0; y < img.h; ++y) {
+        for (int x = 0; x < img.w; ++x) {
+          const uint8_t k = img.bitmap[y * img.stride + x];
+          const int dstX = img.dst_x + x;
+          const int dstY = img.dst_y + y;
+          if (!k || dstX < 0 || dstX > t.Width || dstY < 0 || dstY > t.Height)
```
Gunoshozo:
```suggestion
          if (!k || dstX < 0 || dstX >= t.Width || dstY < 0 || dstY >= t.Height)
```
probably should also check for equality since Width and Height are integers and we start at 0

src/subtitle/ass/subtitlerenderer.cpp
```diff
@@ -0,0 +1,230 @@
...
+          const size_t idx = (dstY * t.Width + dstX) * 4;
+
+          constexpr uint32_t mult = 255 * 255;
+          const glm::u32vec4 src{
+              rgb.r * a * k,
+              rgb.g * a * k,
+              rgb.b * a * k,
+              a * k * 255,
+          };
+          const uint32_t cover = a * k;
```
Gunoshozo: I'm a little lost in this block, I think it's better to decyper the names here a bit
a - I assume is alpha, or rather inverted alpha?
k and cover - I don't get at all

src/video/ffmpegplayer.cpp
```diff
@@ -198,27 +228,98 @@ void FFmpegPlayer::Play(Io::Stream* stream, bool looping, bool alpha) {
...
+        SubPlayer.emplace(VideoTexture->Width, VideoTexture->Height);
+      else
+        SubPlayer.emplace(Profile::DesignWidth, Profile::DesignHeight);
+    }
+  };
+  auto mappingsItr = SubtitleMappings.find(StreamPtr->Meta.FileName);
+
+  // Tracks embedded in video
+  for (auto& [subtitleStream, id] : embeddedSubStreams) {
+    assert(!subtitleStream.isSubtitle() || !subtitleStream.isValid());
```
Gunoshozo: I tried to simplify the expression to better understand it and I don't think I follow:
```cpp
!(subtitleStream.isSubtitle() && subtitleStream.isValid())
```
Error if `isSubtitle` and `isValid`? Shouldn't it be inverted?

KKhanhH: Forgot to invert it cuz its an assertion

src/subtitle/ass/subtitlerenderer.cpp
KKhanhH: Idk the specifics either, cover is like alpha coverage. It's mostly code from libass

src/subtitle/ass/subtitlerenderer.h
```diff
@@ -42,8 +42,46 @@ class SubtitleRenderTrack {
...
+    };
+    GlyphKey(ass_image const& img);
+    uint64_t BitmapHash;
+    glm::ivec2 Dimensions;
+    AssImageType Type;
+
+    friend bool operator==(const GlyphKey& lhs, const GlyphKey& rhs) {
+      return lhs.BitmapHash == rhs.BitmapHash &&
+             lhs.Dimensions == rhs.Dimensions && lhs.Type == rhs.Type;
+    }
```
PringlesGang, line 60: I don't see why this needs to be a friend method?
Can't it just be `bool operator==(const GlyphKey& other) const`?

src/subtitle/ass/subtitlerenderer.h
```diff
@@ -42,8 +42,46 @@ class SubtitleRenderTrack {
...
   };
-  SpriteSheet SubtitlesSpritesheet;
+
+  struct GlyphKey {
+    enum AssImageType : uint8_t {
+      Character,
+      Outline,
+      Shadow,
+    };
+    GlyphKey(ass_image const& img);
```
PringlesGang: Here you use the struct name `ass_image` instead of the typedef name `ASS_Image` like in the rest of the usages. Confused me a little initially

src/subtitle/ass/subtitlerenderer.cpp
```diff
@@ -82,65 +80,83 @@ void SubtitleRenderTrack::Push(SubtitleEntry entry) {
...
+      sheet.Texture = itr->second;
+    } else {
+      Texture t;
+      t.Init(TexFmt_U8, img->w, img->h);
+      for (int y = 0; y < img->h; ++y) {
+        const uint8_t* srcRow = img->bitmap + y * img->stride;
+        std::span<uint8_t> dstRow =
+            std::span(t.Buffer).subspan(y * img->w, img->w);
+        std::memcpy(dstRow.data(), srcRow, img->w);
+      }
```
PringlesGang, line 116: Can you not just `std::mempcy(t.Buffer, img->bitmap, img->stride * (img->h - 1) + img->w)`?

KKhanhH: Stride might not match

PringlesGang: Ahh ofc

PR #428, opened 2026-03-15
Profile improvements

KKhanhH:
Base tryget off of optional rather than bool outparam (for no default constructibility)
Generic API for containers
All glm::vec type support

### DIFF
--- a/src/profile/profile_internal.h
+++ b/src/profile/profile_internal.h
@@ -68,9 +68,6 @@ void GetArray(std::span<T> out);
 template <typename T>
 void GetMemberArray(std::span<T> out, char const* name);
 
-template <typename T>
-std::vector<T> GetMemberVector(char const* name);
-
 void ForEachProfileArray(std::invocable<uint32_t> auto func);
 
 template <typename K>
@@ -381,19 +378,6 @@ inline void GetMemberArray(std::span<T> out, char const* name) {
   EnsurePushMemberOfType(name, LUA_TTABLE);
   GetArray(out);
   Pop();
-}
-
-template <typename T>
-inline std::vector<T> GetMemberVector(char const* name) {
-  std::vector<T> result;
-  EnsurePushMemberOfType(name, LUA_TTABLE);
-  ForEachProfileArray([&](uint32_t i) {
-    assert(i == result.size());
-    result.push_back(EnsureGetArrayElement<T>());
-  });
-
-  Pop();
-  return result;
 }
 
 inline void ForEachProfileArray(std::invocable<uint32_t> auto func) {
--- a/src/profile/data/tipssystem.cpp
+++ b/src/profile/data/tipssystem.cpp
@@ -25,7 +25,8 @@ void Configure() {
             std::make_unique<Impacto::MO6TW::TipsSystem>(MaxTipsCount);
         break;
       case TipsSystemType::CHLCC:
-        SortCategoryMapping = GetMemberVector<int>("SortCategoryMapping");
+        SortCategoryMapping =
+            EnsureGetMember<std::vector<int>>("SortCategoryMapping");
         Implementation =
             std::make_unique<Impacto::CHLCC::TipsSystem>(MaxTipsCount);
         break;
--- a/src/profile/games/cc/dialoguebox.cpp
+++ b/src/profile/games/cc/dialoguebox.cpp
@@ -14,8 +14,10 @@ void Configure() {
 }
 
 void ConfigureNametag() {
-  NametagMainSprites = GetMemberVector<Sprite>("NametagMainSprites");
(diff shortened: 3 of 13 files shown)


src/profile/profile_internal.h
```diff
@@ -107,120 +254,93 @@ const auto& GetEntityMap() {
...
+  out[0] = TryGetMember<value_type>("X");
+  if constexpr (T::length() > 1) out[1] = TryGetMember<value_type>("Y");
+  if constexpr (T::length() > 2) out[2] = TryGetMember<value_type>("Z");
+  if constexpr (T::length() > 3) out[3] = TryGetMember<value_type>("W");
+  if (!std::ranges::all_of(out, [](auto& v) { return v.has_value(); }))
+    return std::nullopt;
+
+  return [&]<std::size_t... I>(std::index_sequence<I...>) {
+    return std::optional<T>{T{*out[I]...}};
+  }(std::make_index_sequence<T::length()>{});
```
Gunoshozo: I get that part, but it's a little hard for the eyes

src/profile/profile_internal.cpp
Gunoshozo, line 75: While we are at it, maybe we can add a log here, so it's not a silent exit

src/profile/profile_internal.cpp
```diff
@@ -77,71 +77,48 @@ void PushInitialIndex() { lua_pushnil(LuaState); }
...
-    return true;
+    return lua_toboolean(LuaState, -1);
   }
   // TODO conversion?
 
-  return false;
+  return std::nullopt;
 }
 
 void ClearProfileInternal() { lua_close(LuaState); }
```
Gunoshozo: maybe also set `LuaState` as `nullptr`?

src/profile/profile_internal.h
```diff
@@ -107,120 +254,93 @@ const auto& GetEntityMap() {
...
+inline std::optional<T> TryGet() {
+  if (!lua_istable(LuaState, -1)) return std::nullopt;
+  using value_type = typename T::value_type;
+  std::array<std::optional<value_type>, T::length()> out;
+  out[0] = TryGetMember<value_type>("X");
+  if constexpr (T::length() > 1) out[1] = TryGetMember<value_type>("Y");
+  if constexpr (T::length() > 2) out[2] = TryGetMember<value_type>("Z");
+  if constexpr (T::length() > 3) out[3] = TryGetMember<value_type>("W");
+  if (!std::ranges::all_of(out, [](auto& v) { return v.has_value(); }))
+    return std::nullopt;
```
PringlesGang, line 280: Perhaps `static_assert(T::length() <= 4)`?

src/profile/profile_internal.h
```diff
@@ -107,120 +254,93 @@ const auto& GetEntityMap() {
...
+inline std::optional<T> TryGet() {
+  if (!lua_istable(LuaState, -1)) return std::nullopt;
+
+  T result;
+  size_t count = 0;
+  ForEachProfileArray([&]([[maybe_unused]] size_t i) {
+    assert(count == i);
+    result[count++] = EnsureGet<typename T::value_type>();
+  });
+  assert(count == result.size());
```
PringlesGang: I think this assert should be done in release mode as well. If there are not enough elements or too many elements, it's a profile error.

src/profile/profile_internal.h
```diff
@@ -241,7 +361,7 @@ void GetMemberArray(T* arr, size_t count, char const* name) {
 }
```
PringlesGang: This should take in a span, while we're at it.
Preferably I'd also like to see this take in string_views, but those are not guaranteed to be null-terminated...

src/profile/profile_internal.h
```diff
@@ -31,65 +31,212 @@ void EnsurePushMemberOfType(char const* name, int type);
...
+  return result.value();
+}
+
+template <typename T>
+inline T EnsureGetKey() {
+  if constexpr (is_any_of<T, char const*, std::string_view, std::string>)
+    return lua_tostring(LuaState, -2);
+  if constexpr (std::is_integral_v<T> && !std::is_same_v<T, bool>)
+    return (T)lua_tointeger(LuaState, -2);
+}
```
PringlesGang: Here, and in a handful of other places, you have a sequence of `if constexpr` statements effectively used as a switch, but if none of those are hit they just generate an "empty implementation." Maybe the last of these `else if constexpr`s could be replaced with an `else { static_assert()` to ensure they do not compile if they're not explicitly implemented?

src/profile/profile_internal.h
KKhanhH: Separated into a named lambda then called it

src/profile/profile_internal.cpp
KKhanhH: Might make a helper function at some point for these fatal exits that logs and throws/aborts.

PR #432, opened 2026-03-22
Keyboard input

KKhanhH:
Rework keyboard input to support KBcustom map for KeyboardOnJump.
SImplify scriptinput profile.
Add PADtoKB and have it map to a KBcustom id.

### DIFF
--- a/profiles/common/scriptinput.lua
+++ b/profiles/common/scriptinput.lua
@@ -169,7 +169,44 @@ root.Input.PADcustomA = {
     root.PADinput.PAD1R1,
 }
 
-root.Input.PADcustomB = {};
+root.Input.PADcustomB = {
+  root.PADinput.PAD1UP,
+  root.PADinput.PAD1DOWN,
+  root.PADinput.PAD1LEFT,
+  root.PADinput.PAD1RIGHT,
+  root.PADinput.PAD1A | root.PADinput.PAD1START,
+  root.PADinput.PAD1A,
+  root.PADinput.PAD1B,
+  root.PADinput.PAD1L2,
+  root.PADinput.PAD1R1,
+  root.PADinput.PAD1R2,
+  root.PADinput.PAD1START,
+  0,
+  root.PADinput.PAD1Y,
+  root.PADinput.PAD1R3,
+  root.PADinput.PAD1A | root.PADinput.PAD1B | root.PADinput.PAD1START,
+  root.PADinput.PAD1A,
+  root.PADinput.PAD1B,
+  root.PADinput.PAD1X,
+  root.PADinput.PAD1Y,
+  root.PADinput.PAD1R1,
+  root.PADinput.PAD1R2,
+  root.PADinput.PAD1L1,
+  0,
+  AllDir | root.PADinput.PAD1A,
+  root.PADinput.PAD1R1,
+  root.PADinput.PAD1L1,
+  root.PADinput.PAD1A,
+  root.PADinput.PAD1B,
+  root.PADinput.PAD1UP_LS | root.PADinput.PAD1UP_DIRECT,
+  root.PADinput.PAD1DOWN_LS | root.PADinput.PAD1DOWN_DIRECT,
+  root.PADinput.PAD1LEFT_LS | root.PADinput.PAD1LEFT_DIRECT,
+  root.PADinput.PAD1RIGHT_LS | root.PADinput.PAD1RIGHT_DIRECT,
+  root.PADinput.PAD1UP_RS,
+  root.PADinput.PAD1DOWN_RS,
+  root.PADinput.PAD1LEFT_RS,
+  root.PADinput.PAD1RIGHT_RS,
+};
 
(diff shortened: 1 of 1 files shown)


src/vm/inst_controlflow.cpp
```diff
@@ -247,17 +247,41 @@ VmInstruction(InstClickOnJump) {
...
+
+  if (inputResult) {
+    thread->IpOffset = labelAdr;
+  }
+
+  ImpLogSlow(LogLevel::Warning, LogChannel::VMStub,
+             "STUB instruction KeyboardOnJump(type: {:d}, buttonsArg: {:d}, "
+             "downTypeId: {:d}, "
+             "labelNum: {:d})\n",
+             type, buttonsArg, downTypeId, labelNum);
```
Gunoshozo: Is there something more to it?
Or we can change that to LogLevel::Trace, LogChannel::VM?

KKhanhH: there should not be, i just always forget to remove stub logs

KKhanhH: trace wouldn't be bad though, since we need to modify scripts and we wanna check if they worked

src/vm/interface/input.cpp
```diff
@@ -185,6 +198,21 @@ void UpdatePADHoldInput(float dt) {
   }
 }
 
+static std::array<float, SDL_NUM_SCANCODES> KBIsDownTime{};
```
PringlesGang: Could be made a static local of `UpdateKBHoldInput` to not pollute the namespace

src/vm/inst_controlflow.cpp
```diff
@@ -247,17 +247,41 @@ VmInstruction(InstClickOnJump) {
...
+  PopUint16(labelNum);
+  uint32_t labelAdr = ScriptGetLabelAddress(thread->ScriptBufferId, labelNum);
+
+  const InputDownType downType = static_cast<InputDownType>(downTypeId);
+  const bool useKBcustom = (type & 2);
+  const bool inputResult = [&]() -> bool {
+    bool* const kbBtnArr =
+        downType == InputDownType::IsDown     ? KeyboardButtonIsDown
+        : downType == InputDownType::WentDown ? KeyboardButtonWentDown
+                                              : nullptr;
```
PringlesGang: `std::span`? `std::span()` defaults to `.data() == nullptr` and `.size() == 0`

profiles/dash/scriptinput.lua
PringlesGang, line 1: Why not just remove the file?

profiles/common/scriptinput.lua
PringlesGang, line 1: I'm having a hard time figuring out what arrays store what kind of values (indices into some other array, SDL scancodes, etc.), and how I should associate these values with buttons or game actions...
Maybe add a few more comments clearly explaining what each array stores, and how one could customize it? I expect this profile section to be pretty popular for user customization

PringlesGang (changes requested):
The `PADcustomSize`s seem kinda redundant and error-prone now that we can just make the `PADcustom`s vectors and `EnsureGetMember(PADcustomA, "PADcustomA")`

profiles/common/scriptinput.lua
KKhanhH: Pad to kb might be more accurately named as padtokbcustom, but otherwise thats the only one thats using index into another array, everything else is an sdl scancode. Padcustom and kbcustom indices, I don't really know what indices correspond to for gameactions, since it's just whatever the script says, perhaps theres some constant names in the debug scripts.

This file might be more clear once i do a magic_enum refactor, then we can just reflect SDL scancodes into the lua file and use it directly in there.

src/vm/inst_controlflow.cpp
```diff
@@ -247,17 +247,41 @@ VmInstruction(InstClickOnJump) {
...
+  uint32_t labelAdr = ScriptGetLabelAddress(thread->ScriptBufferId, labelNum);
+
+  const InputDownType downType = static_cast<InputDownType>(downTypeId);
+  const bool useKBcustom = (type & 2);
+  const bool inputResult = [&]() -> bool {
+    std::span<const bool> const kbBtnArr =
+        (downType == InputDownType::IsDown     ? KeyboardButtonIsDown
+         : downType == InputDownType::WentDown ? KeyboardButtonWentDown
+                                               : std::span<const bool>{});
+    if (kbBtnArr.empty()) return true;
```
PringlesGang, line 266: Should this default to true?

profiles/common/scriptinput.lua
PringlesGang: Dunno if you wanna do the magic_enum SDL scancodes in this PR or the magic_enum then, but so long as it happens in either I'm happy :)

src/vm/inst_controlflow.cpp
KKhanhH: It's what the executable does yeah, should never happen anyway though

PR #434, opened 2026-03-24
Magic enum

KKhanhH:
Switch out implementation behind enum lua interface to use magic_enum instead of better_enum, which does not require declaring enums with a macro to create a special class, and works on native enum types.
Add templatized unary operator+ to convert enums to underlying type.
Allow TryGet to get enum types directly without needing to cast from underlying.

**This change will limit the usable compilers to Clang, GCC and MSVC.**

### DIFF
--- a/src/audio/audiocommon.h
+++ b/src/audio/audiocommon.h
@@ -23,10 +23,6 @@ enum AudioChannelId {
   AC_Count
 };
 
-constexpr inline auto format_as(AudioChannelId channel) -> int {
-  return to_underlying(channel);
-}
-
 enum AudioChannelState {
   ACS_Stopped,
   ACS_Playing,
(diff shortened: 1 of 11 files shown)


profiles/common/scriptinput.lua
```diff
@@ -87,35 +87,37 @@ root.Input.PADtoMS = {
...
+    [root.PADinput.PAD1R3]           = BTN.RIGHTSTICK,
+    [root.PADinput.PAD1L1]           = BTN.LEFTSHOULDER,
+    [root.PADinput.PAD1R1]           = BTN.RIGHTSHOULDER,
+    [root.PADinput.PAD1UP_DIRECT]    = BTN.DPAD_UP,
+    [root.PADinput.PAD1DOWN_DIRECT]  = BTN.DPAD_DOWN,
+    [root.PADinput.PAD1LEFT_DIRECT]  = BTN.DPAD_LEFT,
+    [root.PADinput.PAD1RIGHT_DIRECT] = BTN.DPAD_RIGHT,
 };
-
+local AXIS = ControllerAxis;;
```
Enorovan: `local AXIS = ControllerAxis;`

PR #435, opened 2026-03-24
Dialoguepage trifurcation

### LINKED ISSUE #249: [DialoguePage Refactor Step 3] Duplicate string parsing for the backlogs so that they don't keep the "styling" from the main DialoguePage
This should also: 
- Fix Erin's lines display in the C;HLCC backlog
- Change the Backlog baseclass to allow to snap to next entry in C;HLCC
- Not have big lines and other effects being shown in backlog (basically only alignment (and perhaps only center) should show)
- Backlogs shouldn't have the current line. It should stop one line prior

### LINKED ISSUE #283: Fix JP word-wrapping
Word-wrapping isn't quite 1:1 yet for JP

E.g.:
'々' is a word-ending character, but it shouldn't always end a line there, despite JP often being a stream of kana/kanji:

Expected:
> また、情報過多なこの時代③⑤アレがいかに衝撃的な事件であったにせよ、
> わずか数か月で人々 の記憶は化石となり、いつしか風化していこうとして
> いた。

Got:
> また、情報過多なこの時代③⑤アレがいかに衝撃的な事件であったにせよ、
> わずか数か月で人々
> の記憶は化石となり、いつしか風化していこうとしていた。

PringlesGang:
Trifurcates the `DialoguePage` class into separate instances for dialogue, backlog entries, and tip entries

Moves text parsing to a separate class, with separate directive functions for dialogue, backlog, and tips, handling what is parsed and how

- [x] Fixes #249 
- [x] Only pushes a new backlog entry at the end of `InstMesMain`, instead of at the start of `InstMes`
- [x] Fixes #283 
- [x] Give dialogue boxes and nametags more freedom in how they can be rendered
- [x] Implement placing the dialogue box through ScrWork
- [x] Fine-tune nametag visibility and positioning
- [x] Allow for text to be transformed during rendering
- [x] Perfect nametag and ruby positioning during parsing
- [x] Fix up chlcc/cclcc backlog entry rendering and formatting
- [x] Implement `InstSetMesModeFormat` and make the parser use these values
- [x] Fix DialoguePage clearing semantics
- [x] Allow for text centering for backlog entries
- [x] Add support for EN wordwrapping for added glyphs
- [x] Clean up profile
- [x] Add profile endpoints for `InstSetMesModeFormat` values
- [x] Implement Erin textbox fading
- [x] Disable cursor requests when hovering a scrollwheel area

Test English wordwrapping with https://github.com/CommitteeOfZero/impacto-patch/pull/1

English `TextModesInfo` config is very negotiable


src/ui/widgets/backlogentry.h
```diff
@@ -24,15 +24,14 @@ class BacklogEntry : public Widget {
...
+  std::optional<int> AudioId;
   int CharacterId = 0;
   float TextHeight = 0.0f;
 
   std::function<void(BacklogEntry*)> OnClickHandler;
 
  protected:
-  DialoguePage* BacklogPage;
-  int TextLength = 0;
+  std::unique_ptr<BacklogPage> Page;
```
KKhanhH, line 33: Instead of having a backlogpage pointer per backlog entry (which adds a lot of indirection and fragmentation), could we instead maybe store a vector of backlogpages inside backlog menu (since it's not polymorphic), and then access that in backlogentry with an index and maybe a pointer to the backlogmenu?
If backlog max count should be a bounded value, we should store it in a static array as well.

src/text/textparser.h
```diff
@@ -0,0 +1,112 @@
...
+
+  std::set<size_t> ParallelStartGlyphs;
+
+  DialoguePage::AdvanceMethodType AdvanceMethod =
+      DialoguePage::AdvanceMethodType::Skip;
+
+  DialoguePageMode PageMode = DPM_ADV;
+
+ protected:
+  void ParseLineBreak(const StringToken& token);
```
KKhanhH: I would suggest a template function for this templated on the StringToken, then you can specialize it in the textparser.cpp since it's only being used there anyway. To handle runtime to compiletime lookup, I would recommend doing a constexpr static lookup table like what we had for backgrounds2d table rather than a heap allocated map + std function, just do nullptr checks if you wanna leave some blank.

```cpp
template <StringTokenType T>
void ParseStringToken(const Stringtoken& t);

using TextParserProc = auto (TextParser::*)(const StringToken&) -> void;
auto constexpr static textParserLUT = std::to_array<TextParserProc>({
 &TextParser::ParseStringToken<STT_LineBreak>,
 &TextParser::ParseStringToken<STT_CharacterNameStart>,
 nullptr, 
});

```

PringlesGang: Is this about what you meant? https://github.com/CommitteeOfZero/impacto/pull/435/changes/a35519b8d5a9afacf27222cd605227042e8a08ba

KKhanhH: Yeah pretty much, i think you can use magic enum's enum containers too

src/profile/dialogue.cpp
```diff
@@ -54,15 +56,58 @@ static void ConfigureNametag() {
...
+
+    tryOverwrite("LineSpacing", info.LineSpacing);
+    tryOverwrite("RubyLineSpacing", info.RubyLineSpacing);
+    tryOverwrite("AlwaysAddRubySpacing", info.AlwaysAddRubySpacing);
+    tryOverwrite("LinefeedSpacing", info.LinefeedSpacing);
+
+    tryOverwrite("NameAlignment", info.NameAlignment);
+
+    tryOverwrite("UseNameLengthL", info.UseNameLengthL);
+    tryOverwrite("NameLengthL", info.NameLengthL);
```
KKhanhH:
```suggestion
    info.DisplayMode = TryGetMember<decltype(*info.DisplayMode)>("DisplayMode");
    info.WindowId = TryGetMember<decltype(*info.WindowId)>("WindowId");

    info.WindowPos = TryGetMember<decltype(*info.WindowPos)>("WindowPos");

    info.NameDispMode =
        TryGetMember<decltype(*info.NameDispMode)>("NameDispMode");

    info.MaxNameWidth =
        TryGetMember<decltype(*info.MaxNameWidth)>("MaxNameWidth");
    info.NamePos = TryGetMember<decltype(*info.NamePos)>("NamePos");
    info.NameGlyphSize =
        TryGetMember<decltype(*info.NameGlyphSize)>("NameGlyphSize");

    info.MaxLineWidth =
        TryGetMember<decltype(*info.MaxLineWidth)>("MaxLineWidth");

    info.CurrentPageId =
        TryGetMember<decltype(*info.CurrentPageId)>("CurrentPageId");
    info.WaitIconPos = TryGetMember<decltype(*info.WaitIconPos)>("WaitIconPos");

    info.TextGlyphSize =
        TryGetMember<decltype(*info.TextGlyphSize)>("TextGlyphSize");
    info.RubyGlyphSize =
        TryGetMember<decltype(*info.RubyGlyphSize)>("RubyGlyphSize");

    info.LineSpacing = TryGetMember<decltype(*info.LineSpacing)>("LineSpacing");
    info.RubyLineSpacing =
        TryGetMember<decltype(*info.RubyLineSpacing)>("RubyLineSpacing");
    info.AlwaysAddRubySpacing =
        TryGetMember<decltype(*info.AlwaysAddRubySpacing)>(
            "AlwaysAddRubySpacing");
    info.LinefeedSpacing =
        TryGetMember<decltype(*info.LinefeedSpacing)>("LinefeedSpacing");

    info.NameAlignment =
        TryGetMember<decltype(info.NameAlignment)>("NameAlignment");

    info.UseNameLengthL =
        TryGetMember<decltype(*info.UseNameLengthL)>("UseNameLengthL");
    info.NameLengthL = TryGetMember<decltype(*info.NameLengthL)>("NameLengthL");
```

KKhanhH: It's that or you can just add a helper function in profile internal to take in an optional as well as the second arg

src/profile/dialogue.h
```diff
@@ -117,6 +117,39 @@ inline Sprite MenuButtonSprite;
...
+  std::optional<float> RubyLineSpacing;
+  std::optional<bool> AlwaysAddRubySpacing;
+  std::optional<float> LinefeedSpacing;
+
+  std::optional<TextAlignment> NameAlignment;
+
+  // Idk what these are either
+  std::optional<bool> UseNameLengthL;
+  std::optional<uint16_t> NameLengthL;
+};
```
KKhanhH: If you want to shave off padding maybe, but it would require you to manually do a has check before getting elements rather than being enforced at the type level, unless you just wanna add a getter as well that needs a switch case on everything (no reflection yet :( ).
```
struct ProfileTextModeInfo {
  enum class Field : uint32_t {
    None = 0,
    DisplayMode = 1 << 0,
    WindowId = 1 << 1,
    WindowPos = 1 << 2,
    NameDispMode = 1 << 3,
    MaxNameWidth = 1 << 4,
    NamePos = 1 << 5,
    NameGlyphSize = 1 << 6,
    MaxLineWidth = 1 << 7,
    CurrentPageId = 1 << 8,
    WaitIconPos = 1 << 9,
    TextGlyphSize = 1 << 10,
    RubyGlyphSize = 1 << 11,
    LineSpacing = 1 << 12,
    RubyLineSpacing = 1 << 13,
    AlwaysAddRubySpacing = 1 << 14,
    LinefeedSpacing = 1 << 15,
    NameAlignment = 1 << 16,
    UseNameLengthL = 1 << 17,
    NameLengthL = 1 << 18,
    All = 0xFFFF,
  };

  Field FieldFlags = Field::None;

  uint16_t DisplayMode{};
  size_t WindowId{};

  glm::vec2 WindowPos{};

  TextModeInfo::NameDispModeType NameDispMode{};

  float MaxNameWidth{};
  glm::vec2 NamePos{};
  glm::vec2 NameGlyphSize{};

  float MaxLineWidth{};

  size_t CurrentPageId{};
  glm::vec2 WaitIconPos{};

  glm::vec2 TextGlyphSize{};
  glm::vec2 RubyGlyphSize{};

  float LineSpacing{};
  float RubyLineSpacing{};
  bool AlwaysAddRubySpacing{};
  float LinefeedSpacing{};

  TextAlignment NameAlignment{};

  // Idk what these are either
  bool UseNameLengthL{};
  uint16_t NameLengthL{};

  bool HasField(Field field) const {
    return (+FieldFlags & +field) != +Field::None;
  }
};
```

src/vm/inst_dialogue.cpp
```diff
@@ -389,6 +395,36 @@ VmInstruction(InstSetMesModeFormat) {
...
+      ProfileOrProcessed(TextGlyphSize),
+      ProfileOrProcessed(RubyGlyphSize),
+      ProfileOrProcessed(LineSpacing),
+      ProfileOrProcessed(RubyLineSpacing),
+      ProfileOrProcessed(AlwaysAddRubySpacing),
+      ProfileOrProcessed(LinefeedSpacing),
+      ProfileOrProcessed(NameAlignment),
+      ProfileOrProcessed(UseNameLengthL),
+      ProfileOrProcessed(NameLengthL),
+  };
```
KKhanhH: Something like this maybe? and call if to merge the two

```
  void Merge(const ProfileTextModeInfo& other) {
    auto checkAndAssign = [&](auto ProfileTextModeInfo::* member) {
      if ((other.*member).has_value()) {
        this->*member = other.*member;
      }
    };

    checkAndAssign(&ProfileTextModeInfo::DisplayMode);
    checkAndAssign(&ProfileTextModeInfo::WindowId);
    checkAndAssign(&ProfileTextModeInfo::WindowPos);
    checkAndAssign(&ProfileTextModeInfo::NameDispMode);
    checkAndAssign(&ProfileTextModeInfo::MaxNameWidth);
    checkAndAssign(&ProfileTextModeInfo::NamePos);
    checkAndAssign(&ProfileTextModeInfo::NameGlyphSize);
    checkAndAssign(&ProfileTextModeInfo::MaxLineWidth);
    checkAndAssign(&ProfileTextModeInfo::CurrentPageId);
    checkAndAssign(&ProfileTextModeInfo::WaitIconPos);
    checkAndAssign(&ProfileTextModeInfo::TextGlyphSize);
    checkAndAssign(&ProfileTextModeInfo::RubyGlyphSize);
    checkAndAssign(&ProfileTextModeInfo::LineSpacing);
    checkAndAssign(&ProfileTextModeInfo::RubyLineSpacing);
    checkAndAssign(&ProfileTextModeInfo::AlwaysAddRubySpacing);
    checkAndAssign(&ProfileTextModeInfo::LinefeedSpacing);
    checkAndAssign(&ProfileTextModeInfo::NameAlignment);
    checkAndAssign(&ProfileTextModeInfo::UseNameLengthL);
    checkAndAssign(&ProfileTextModeInfo::NameLengthL);
  }
};
```

src/vm/inst_dialogue.cpp
```diff
@@ -356,14 +358,18 @@ VmInstruction(InstSetMesModeFormat) {
   Sc3VmThread dummy;
   dummy.IpOffset = modeDataAdr;
   dummy.ScriptBufferId = thread->ScriptBufferId;
-  RawMesModeInfo* info = std::bit_cast<RawMesModeInfo*>(dummy.GetIp());
+  const RawMesModeInfo* const info =
+      std::bit_cast<RawMesModeInfo*>(dummy.GetIp());
+
+  const ProfileTextModeInfo* const profileInfo =
+      ProfileTextModesInfo.contains(id) ? &ProfileTextModesInfo[id] : nullptr;
```
KKhanhH: Don't use contains and [], just use find. it's ugly but it's standard c++

src/ui/widgets/backlogentry.h
PringlesGang: Yeah that's definitely a good idea. I'm already working on a commit to fix this, which would also simultaneously add a cap on the number of backlog entries (because right now it'll just keep appending indefinitely), but it's growing kinda large and out of scope for this PR. Mind if I fix this in a later PR?

Enorovan (changes requested):
Review feedbacks, some may be addressed elsewhere:
- [x] <del>You can still open empty backlog in C;HLCC.</del>
- [x] Double linebreaks aren't parsed in Tips Menus.
- [x] Weird case when skipping on "looped" lines in Map Trigger.
- [x] Weird Cursor Pointer in C;CLCC Tips Menu.
- [x] Erin Box is supposed to fade-in/fade-out.

PringlesGang:
> Review feedbacks, some may be addressed elsewhere :
> 
> * [ ]  You can still open empty backlog in C;HLCC.
> * [ ]  Double linebreaks aren't parsed in Tips Menus.
> * [ ]  Weird case when skipping on "looped" lines in Map Trigger.
> * [ ]  Weird Cursor Pointer in C;CLCC Tips Menu.
> * [ ]  Erin Box is supposed to fade-in/fade-out.

All should be fixed now, except for being able to open an empty backlog in C;HLCC. On PS3 this is also possible though, so I don't think that was an issue in the first place

src/hud/dialoguebox.cpp
```diff
@@ -29,23 +29,30 @@ std::unique_ptr<DialogueBox> DialogueBox::Create() {
...
-      return std::make_unique<PlainDialogueBox>();
+      return std::make_unique<PlainDialogueBox>(page);
   }
 }
 
-PlainDialogueBox::PlainDialogueBox() {
+glm::vec2 DialogueBox::GetScrWorkPos() const {
+  using namespace Impacto::Profile::ScriptVars;
+  const int id = ParentPage.get().Id;
+  return {ScrWork[SW_MESWIN0POSX + 10 * id], ScrWork[SW_MESWIN0POSY + 10 * id]};
```
KKhanhH: Should 10 be hardcoded here

src/profile/dialogue.cpp
```diff
@@ -54,15 +56,61 @@ static void ConfigureNametag() {
...
 
+static void ConfigureTextModesInfo() {
+  if (!TryPushMember("TextModesInfo")) return;
+  AssertIs(LUA_TTABLE);
+  PushInitialIndex();
+  while (PushNextTableElement() != 0) {
+    const size_t index = EnsureGetKey<size_t>();
+    if (index >= ProfileTextModesInfoFields.size()) {
+      throw std::runtime_error(
+          fmt::format("TextModeInfo index %u out of range [0, %u)", index,
```
KKhanhH: shouldn't these be {}

src/profile/dialogue.h
```diff
@@ -8,17 +10,28 @@
...
+}
+}  // namespace Profile
+}  // namespace Impacto
+
+namespace magic_enum::customize {
+template <>
+struct enum_range<Impacto::Profile::Dialogue::TextModeInfoFieldFlags> {
+  constexpr static bool is_flags = true;
+};
+}  // namespace magic_enum::customize
```
KKhanhH, line 28: You can put this below so you don't have to declare two namespaces

src/profile/dialogue.cpp
```diff
@@ -54,15 +56,61 @@ static void ConfigureNametag() {
   }
 }
 
+static void ConfigureTextModesInfo() {
+  if (!TryPushMember("TextModesInfo")) return;
+  AssertIs(LUA_TTABLE);
+  PushInitialIndex();
+  while (PushNextTableElement() != 0) {
```
KKhanhH: You can also use ForEachProfileArray

src/text/backlogpage.cpp
```diff
@@ -0,0 +1,49 @@
...
+                         const RendererOutlineMode outlineMode) {
+  TextPage::Render(alpha, outlineMode);
+  Renderer->DrawProcessedText(Name, DialogueFont, alpha, outlineMode);
+}
+
+RectF BacklogPage::SetBounds() {
+  if (Glyphs.empty()) return Bounds = RectF{};
+  TextPage::SetBounds();
+
+  Bounds =
```
KKhanhH, line 40: is bounds supposed to be names or glyphs?

src/vm/inst_dialogue.cpp
```diff
@@ -328,13 +321,92 @@ VmInstruction(InstMesMain) {
...
-  (void)modeDataAdr;
-  ImpLogSlow(LogLevel::Warning, LogChannel::VMStub,
-             "STUB instruction SetMesModeFormat(id: {:d})\n", id);
-}
+
+  Sc3VmThread dummy;
+  dummy.IpOffset = modeDataAdr;
+  dummy.ScriptBufferId = thread->ScriptBufferId;
+  const RawMesModeInfo* const info =
+      std::bit_cast<RawMesModeInfo*>(dummy.GetIp());
```
KKhanhH: I think this is still technically UB. you probably need to do `std::bit_cast<RawMesModeInfo>(*reinterpret_cast<const std::array<uint8_t, sizeof(RawMesModeInfo)>*>(raw_bytes));`

src/vm/inst_dialogue.cpp
```diff
@@ -328,13 +321,92 @@ VmInstruction(InstMesMain) {
...
+
+  const auto profileFields = ProfileTextModesInfoFields[id];
+  TextModeInfo& dest = TextModesInfo[id];
+  const auto setVal = [profileFields]<typename T, typename U>(
```
KKhanhH: You could also use auto params instead of declaring the template, tho you'd need decltype value instead of U

KKhanhH (changes requested):

Enorovan (approved):

src/profile/dialogue.h
PringlesGang: No, I can't. The `TextModeInfoFieldFlags` needs to be forward-declared for the `magic_enum::customize`, and the `magic_enum::customize` is needed for the `magic_enum::container::bitset<TextModeInfoFieldFlags>` at the end of the file

src/text/backlogpage.cpp
PringlesGang: `Bounds` is the bounding rectangle of all text on the page—glyphs, name and ruby

src/vm/inst_dialogue.cpp
PringlesGang: Could you write this out for me? I'm not sure how auto params would work here, and get a bit confused by the reference-stacking that would happen :p
Would this also maintain perfect forwarding?

src/vm/inst_dialogue.cpp
PringlesGang: Dang... Does this have to do with alignment or something?

src/text/textparser.cpp
```diff
@@ -0,0 +1,677 @@
...
+            break;
+          case Right:
+            pos.x += ModeInfo.NamePos.x + ModeInfo.MaxNameWidth - nameWidth;
+            break;
+        }
+      }
+      break;
+
+    case Invisible:
+      assert(false);
```
Gunoshozo, line 466: Impossible state?

src/hud/dialoguebox.cpp
Gunoshozo: Seen same `+ 10 *` in couple othe places
And also `SW_MESWIN0POSX_OFS + 2 *`

IIRC those numbers are struct sizes inside ScrWork?

src/vm/inst_dialogue.cpp
KKhanhH: Alignment, object lifetime, strict aliasing rules apply here

src/vm/inst_dialogue.cpp
KKhanhH: Use auto inplace of template params `const auto setVal = [profileFields](auto& member, TextModeInfoFieldFlags flag, auto&& value)`
And `std::forward<decltype(value)>(value);` to forward

src/text/textparser.cpp
PringlesGang: Yes, a couple lines above we're early-returning on `ModeInfo.NameDispMode == Invisible`

src/hud/dialoguebox.cpp
PringlesGang: I've added a profile var for the `10 *` cases. For `SW_MESWIN0POSX_OFS + 2 *` I think it's a bit overkill, since that's just alternating X and Y, which I don't think they'd ever change

src/vm/inst_dialogue.cpp
```diff
@@ -361,8 +361,9 @@ VmInstruction(InstSetMesModeFormat) {
   Sc3VmThread dummy;
   dummy.IpOffset = modeDataAdr;
   dummy.ScriptBufferId = thread->ScriptBufferId;
-  const RawMesModeInfo* const info =
-      std::bit_cast<RawMesModeInfo*>(dummy.GetIp());
+  const RawMesModeInfo info = std::bit_cast<RawMesModeInfo>(
+      *reinterpret_cast<const std::array<uint8_t, sizeof(RawMesModeInfo)>*>(
```
KKhanhH: Forgot about object lifetime rules with std array, use `        *reinterpret_cast<const uint8_t(*)[sizeof(RawMesModeInfo)]>`

KKhanhH (approved):

src/vm/inst_dialogue.cpp
KKhanhH: ~~`std::bit_cast<RawMesModeInfo, const uint8_t(*)[sizeof(RawMesModeInfo)]>`~~

~~You might be able to use the second template argument of bitcast actually~~

KKhanhH: Eh maybe not, stick with the reinterpret_cast to c-array, itll probably decay to a pointer otherwise

Gunoshozo (approved):
👍💗

PR #436, opened 2026-03-24
EFFWave Effect

Gunoshozo:
EFFWave Effect
- Extend DrawPrimitives method to swap current frameBuffer
- Reorder some stuff in Wave methods, to make less unnecessary calculations
- Add default log to the InstBGeffectWave, to catch newer versions of instruction

Rendering Hot Path Optimizations
- Character2D::Render Moved out a loop invariant
- Renderer::InsertVertices unite 2 passes over the same vector + fix possible side effect of using restart indices
- Renderer::DrawPrimitives preallocate TransformedVertices and use spans of it + branch on sheet.IsScreenCap to std::transform or std::copy

### DIFF
--- a/src/renderer/opengl/renderer.cpp
+++ b/src/renderer/opengl/renderer.cpp
@@ -269,14 +269,22 @@ void Renderer::InsertVertices(
 
   if (IndexBuffer.size() + indices.size() > MaxIndexCount) Flush();
 
-  const uint16_t maxIndex = *std::max_element(indices.begin(), indices.end());
-
   VertexBuffer.insert(VertexBuffer.end(), vertices.begin(), vertices.end());
 
-  IndexBuffer.reserve(IndexBuffer.size() + indices.size());
-  std::transform(
-      indices.begin(), indices.end(), std::back_inserter(IndexBuffer),
-      [this](uint16_t index) { return index + this->NextFreeIndex; });
+  const size_t offset = IndexBuffer.size();
+  const size_t indicesCount = indices.size();
+  IndexBuffer.resize(offset + indicesCount);
+
+  constexpr uint16_t restartIndex = 0xFFFF;
+  uint16_t maxIndex = 0;
+  for (size_t i = 0; i < indicesCount; ++i) {
+    if (indices[i] == restartIndex) {
+      IndexBuffer[i + offset] = restartIndex;
+    } else {
+      IndexBuffer[i + offset] = indices[i] + NextFreeIndex;
+      if (indices[i] > maxIndex) maxIndex = indices[i];
+    }
+  }
 
   NextFreeIndex += maxIndex + 1;
 }
@@ -776,15 +784,22 @@ void Renderer::DrawPrimitives(
     });
   }
 
-  std::vector<VertexBufferSprites> transformedVertices;
-  transformedVertices.resize(vertices.size());
+  std::span<const VertexBufferSprites> transformedVertices;
+  if (sheet.IsScreenCap) {
+    static std::vector<VertexBufferSprites> transformedVerticesBuffer(
+        MaxIndexCount);
 
-  const auto transformVertex = [sheet](VertexBufferSprites info) {
-    if (sheet.IsScreenCap) info.UV.y = 1.0f - info.UV.y;
-    return info;
-  };
-  std::transform(vertices.begin(), vertices.end(), transformedVertices.begin(),
-                 transformVertex);
+    const auto transformVertex = [](VertexBufferSprites info) {
+      info.UV.y = 1.0f - info.UV.y;
+      return info;
+    };
+    std::transform(vertices.begin(), vertices.end(),
+                   transformedVerticesBuffer.begin(), transformVertex);
+    transformedVertices =
+        std::span(transformedVerticesBuffer).subspan(0, vertices.size());
+  } else {
+    transformedVertices = vertices;
+  }
 
   InsertVertices(transformedVertices, indices);
 }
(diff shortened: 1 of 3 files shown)


src/renderer/opengl/renderer.cpp
```diff
@@ -758,17 +781,22 @@ void Renderer::DrawPrimitives(
...
+  if (sheet.IsScreenCap) {
+    const auto transformVertex = [](VertexBufferSprites info) {
+      info.UV.y = 1.0f - info.UV.y;
+      return info;
+    };
+    std::transform(vertices.begin(), vertices.end(),
+                   TransformedVertices.begin(), transformVertex);
+  } else {
+    std::copy(vertices.begin(), vertices.end(), TransformedVertices.begin());
+  }
```
PringlesGang, line 802: You could even
```cpp
std::vector<VertexBufferSprites> flippedVertices;
std::span<const VertexBufferSprites> transformedVertices;
if (sheet.IsScreenCap) {
  transformedVertices.reserve(vertices.size());

  const auto transformVertex = [](VertexBufferSprites info) {
    info.UV.y = 1.0f - info.UV.y;
    return info;
  };
  std::transform(vertices.begin(), vertices.end(),
                 std::back_inserter(flippedVertices), transformVertex);

  transformedVertices = flippedVertices;
} else {
  transformedVertices = vertices;
}
```

profiles/common/scriptvars.lua
```diff
@@ -303,6 +303,8 @@ root.ScriptVars = {
     SW_EFF_CAP_BUF=6313,
     SW_EFF_CAP_PRI2=6314,
     SW_EFF_CAP_BUF2=6316,
+    SW_EFF_WAVE_PRI=6317,
+    EFFECT_WAVE_ALPHA=6324,
```
KKhanhH: SW_EFF_WAVE_ALPHA?

src/effects/wave.cpp
```diff
@@ -16,9 +16,25 @@ void Init() {
...
-  WavePos.resize(BGWaveGridSize.x * BGWaveGridSize.y);
-  Vertices.resize(BGWaveGridSize.x * BGWaveGridSize.y);
-  Indices.resize((BGWaveGridSize.x * 2 + 1) * (BGWaveGridSize.y - 1) - 1);
+  WavePos.resize((size_t)BGWaveGridSize.x * (size_t)BGWaveGridSize.y);
+  Vertices.resize((size_t)BGWaveGridSize.x * (size_t)BGWaveGridSize.y);
+  Indices.resize(
+      ((size_t)BGWaveGridSize.x * 2 + 1) * ((size_t)BGWaveGridSize.y - 1) - 1);
+
+  // triangle strips with primitive restart index
+  constexpr uint16_t PRIMITIVE_RESTART = 0xFFFF;
```
KKhanhH: just regular naming convention imo for constexpr locals. Full caps snake should only be for macros.

src/renderer/opengl/renderer.cpp
```diff
@@ -260,20 +260,28 @@ YUVFrame* Renderer::CreateYUVFrame(float width, float height) {
...
-      indices.begin(), indices.end(), std::back_inserter(IndexBuffer),
-      [this](uint16_t index) { return index + this->NextFreeIndex; });
+  const size_t offset = IndexBuffer.size();
+  const size_t indicesCount = indices.size();
+  IndexBuffer.resize(offset + indicesCount);
+
+  constexpr uint16_t RESTART_INDEX = 0xFFFF;
+  uint16_t maxIndex = 0;
+  for (size_t i = 0; i < indicesCount; ++i) {
+    if (usingRestartIndices && indices[i] == RESTART_INDEX) {
```
KKhanhH: Is usingRestartIndices bool needed? Does indices[i] == RESTART_INDEX alone take care of that check? 

If usingRestartIndices is false, and indices[i] == 0xFFFF, IndexBuffer[i + offset] would be set to 0xFFFF + NextFreeIndex which seems kinda weird.

src/game.cpp
```diff
@@ -446,6 +454,25 @@ static void RenderMain() {
       Effects::Blur.Render(ScrWork[SW_FEATHERING2]);
     }
 
+    if (Profile::UseWaveEffects) {
+      const bool isCC =
+          +Profile::Vm::GameInstructionSet == +Vm::InstructionSet::CC;
+      if (GetFlag(SF_BGEFF1DISP) &&
+          ((isCC && ScrWork[EFFECT_WAVE_ALPHA]) || !isCC)) {
```
KKhanhH: `GetFlag(SF_BGEFF1DISP) && (!isCC || ScrWork[EFFECT_WAVE_ALPHA])`

src/renderer/opengl/renderer.cpp
Gunoshozo: The idea here was to remove allocations and preallocate TransformedVertices once (cause we know max possible vertex index - uint16 numeric limit) , since DrawPrimitives is commonly used method and effectively it's a hot path for rendering. Bringing `reserve` (`span` doesn't have `reserve`, so it doesn't count, idk) and `back_inserter` defeats the purpose

PringlesGang: Yes but with your change, the source is always copied to a new vector, regardless of whether or not it's a screencap. The change I requested was just to pass the `vertices` vector directly as a span in that case, instead of copying it verbatim to a new vector first

If you want the `transformedVertices` vector to only be allocated once, I'd recommend making it static instead of as member variable

Gunoshozo: I totaly missed the point, yeah

PR #437, opened 2026-03-30
Hardware accelerated video decoding

KKhanhH:
Adds hardware accelerated video decoding on the FFmpeg Video Player


KKhanhH:
This still requires copying hardware frames from gpu memory into cpu memory and then resubmitting as a texture through the renderer. 

Might be able to pass hardware buffer directly but would require platform specific code and probably more effort than its worth. 
https://stackoverflow.com/questions/57242800/how-to-convert-an-ffmpeg-texture-to-open-gl-texture-without-copying-to-cpu-memor

KKhanhH:
bgfx might make it easier but put a pin on that until 0.10 or post release

CMakeLists.txt
```diff
@@ -1248,7 +1257,7 @@ if (NOT DEFINED IMPACTO_DISABLE_VULKAN)
         src/renderer/vulkan/renderer.h
         src/renderer/vulkan/pipeline.h
         src/renderer/vulkan/utils.h
-        src/renderer/vulkan/yuvframe.h
+        src/renderer/vulkan/nv12frame.h
```
PringlesGang, line 1261: I don't think you intended to overwrite this?

src/renderer/opengl/renderer.cpp
```diff
@@ -1014,6 +1023,35 @@ void Renderer::DrawVideoTexture(const YUVFrame& frame, const RectF& dest,
...
 }
 
+void Renderer::DrawVideoTexture(const NV12Frame& frame, const RectF& dest,
+                                const glm::vec4 tint, const bool alphaVideo) {
+  if (!Drawing) {
+    ImpLog(LogLevel::Error, LogChannel::Render,
+           "Renderer->DrawVideoTexture() called before BeginFrame()\n");
+    return;
+  }
+
```
Gunoshozo, line 1033: Missing `EnsureFBO(0)` ?

UPD:
Messed up a little, didn't expect that the comment will look that, way.
 I meant here for the line 1033/1034

src/video/ffmpegplayer.cpp
```diff
@@ -608,9 +744,15 @@ void FFmpegPlayer::Update(float dt) {
...
+    if (frame->Frame.pixelFormat() == AV_PIX_FMT_NV12) {
+      auto& nv12Frame = std::get<NV12Frame*>(VideoTexture);
+      nv12Frame->Submit(frame->Frame.data(0), frame->Frame.raw()->linesize[0],
+                        frame->Frame.data(1), frame->Frame.raw()->linesize[1]);
+    } else if (frame->Frame.pixelFormat() == AV_PIX_FMT_YUV420P) {
+      auto& yuvFrame = std::get<YUVFrame*>(VideoTexture);
+      yuvFrame->Submit(frame->Frame.data(0), frame->Frame.data(1),
+                       frame->Frame.data(2));
+    }
     VideoClock.Set(frame->Timestamp.toDuration<Clock::Microseconds>(),
```
Gunoshozo, line 759: maybe add else with log in case of non supported pixelFormat, though not sure how likely is that, since not super familiar with our ffmpegplayer implementation

Gunoshozo (approved):
Nice, works on my machine™️

src/video/ffmpegplayer.cpp
KKhanhH: Logging is probably unnecessary here since you just won't see video (and it should be pretty rare unless its a pretty niche device/decoder/video decoding)

PringlesGang (approved):

PR #440, opened 2026-04-05
delusion-mouse-support

### LINKED ISSUE #150: Add Mouse Support to delusion triiger rotating circles

Enorovan:
C;HLCC: Add "Heart" Buttons that are clickable like in C;HN (better design though).
C;CLCC: Drag trigger "circles" around.


Enorovan:
C;HLCC: Add "Heart" Buttons that are clickable like in C;HN (better design though).
C;CLCC: Drag trigger "circles" around.

Tested both PC and Android and it responds as intended.

Ready for CR.

src/game.cpp
```diff
@@ -227,6 +228,12 @@ void UpdateGameState(float dt) {
       SetFlag(SF_UIHIDDEN, !GetFlag(SF_UIHIDDEN));
     }
   }
+  if (Profile::Vm::GameInstructionSet == Vm::InstructionSet::CHLCC) {
+    UI::CHLCC::DelusionTrigger::GetInstance().UpdateHeartButtons();
+  }
+  if (Profile::Vm::GameInstructionSet == Vm::InstructionSet::CC) {
```
KKhanhH: Maybe use Profile::GameSpecific::GameSpecificType instead for this line cuz it's technically both cc and cclcc if you use inst set

src/games/cclcc/delusiontrigger.cpp
```diff
@@ -65,37 +69,210 @@ bool DelusionTrigger::CheckStartTransitionComplete() {
...
+      ((hasPositiveHitbox &&
+        positiveHitbox.ContainsPoint(Input::CurMousePos)) ||
+       (hasNegativeHitbox &&
+        negativeHitbox.ContainsPoint(Input::CurMousePos)))) {
+    RequestCursor(CursorType::Pointer);
+  }
+
+  if (!pointerDown) {
+    if (DragPressPending) {
+      const float timeDiff =
```
KKhanhH: use dt

src/games/cclcc/delusiontrigger.cpp
```diff
@@ -65,37 +69,210 @@ bool DelusionTrigger::CheckStartTransitionComplete() {
...
+      DragMouseDownTime = SDL_GetPerformanceCounter() /
+                          static_cast<float>(SDL_GetPerformanceFrequency());
+      DragPressPending = true;
+      DragHoldActive = false;
+    }
+  }
+
+  if (ActiveDragHitbox == DragHitbox::None) return;
+
+  const float timeDiff = SDL_GetPerformanceCounter() /
```
KKhanhH: dt

src/games/cclcc/delusiontrigger.cpp
```diff
@@ -65,37 +69,210 @@ bool DelusionTrigger::CheckStartTransitionComplete() {
...
+    PADinputMouseWentDown &= ~PAD1A;
+  }
+
+  if (!DragHoldActive) return;
+
+  bool dragLeftTrigger = false;
+  bool dragRightTrigger = false;
+  if (std::abs(dragDelta.x) >= dragThreshold) {
+    if (ActiveDragHitbox == DragHitbox::Positive) {
+      if (dragDelta.x > 0.0f && DelusionState == DS_Neutral) {
```
KKhanhH: Maybe some travel distance?

src/games/cclcc/delusiontrigger.cpp
```diff
@@ -65,37 +69,210 @@ bool DelusionTrigger::CheckStartTransitionComplete() {
...
+  }
+
+  leftTrigger |= dragLeftTrigger;
+  rightTrigger |= dragRightTrigger;
+}
+
+void DelusionTrigger::UpdateDragging() {
+  bool leftTrigger = false;
+  bool rightTrigger = false;
+  UpdateDraggingTriggers(leftTrigger, rightTrigger, true);
```
KKhanhH: leftTrigger and rightTrigger should just be returned from UpdateDraggingTriggers

KKhanhH: could use std::pair\<bool>

src/games/chlcc/delusiontrigger.cpp
```diff
@@ -214,6 +218,87 @@ void DelusionTrigger::PlayClickSound() {
...
+  RectF rightBounds =
+      RightDelusionHeartSprite.ScaledBounds().Translate(RightDelusionHeartPos);
+  const bool isTouch = Input::CurrentInputDevice == Input::Device::Touch;
+  const glm::vec2 pointerPos =
+      isTouch ? Input::CurTouchPos : Input::CurMousePos;
+  const bool pointerWentDown = PADinputMouseWentDown & PAD1A;
+
+  if (Input::CurrentInputDevice == Input::Device::Mouse) {
+    LeftHeartHovered = ScrWork[SW_DELUSION_STATE] != DS_Positive &&
+                       leftBounds.ContainsPoint(Input::CurMousePos);
```
KKhanhH: You can just use leftpressed and rightpressed instead of duping this check and set the hovered variables and request cursor if they're true and mouse is active input.

KKhanhH (changes requested):

src/games/chlcc/delusiontrigger.cpp
Enorovan:
```c++
  const bool leftPressed = ScrWork[SW_DELUSION_STATE] != DS_Positive &&
                           leftBounds.ContainsPoint(pointerPos);

  const bool rightPressed = ScrWork[SW_DELUSION_STATE] != DS_Negative &&
                            rightBounds.ContainsPoint(pointerPos);

  if (Input::CurrentInputDevice == Input::Device::Mouse) {
    LeftHeartHovered = leftPressed;
    RightHeartHovered = rightPressed;

    if (LeftHeartHovered || RightHeartHovered) {
      RequestCursor(CursorType::Pointer);
    }
  }
  ```
  like this?

File: src/games/cclcc/delusiontrigger.cpp (hunk 4 above)
Comment (Enorovan) [reply src/games/cclcc/delusiontrigger.cpp]: 10.0f?

File: src/games/cclcc/delusiontrigger.cpp [hunk 7]
```diff
@@ -228,40 +226,37 @@ void DelusionTrigger::UpdateDraggingTriggers(bool& leftTrigger,
...
-  leftTrigger |= dragLeftTrigger;
-  rightTrigger |= dragRightTrigger;
+  return std::make_pair(dragLeftTrigger, dragRightTrigger);
 }
 
 void DelusionTrigger::UpdateDragging(float dt) {
-  bool leftTrigger = false;
-  bool rightTrigger = false;
-  UpdateDraggingTriggers(leftTrigger, rightTrigger, true, dt);
+  std::pair<bool, bool> triggers = UpdateDraggingTriggers(true, dt);
```
KKhanhH: deconstruct it with auto [leftTrigger, rightTrigger] = UpdateDraggingTriggers(true, dt);

src/game.cpp
```diff
@@ -227,6 +228,12 @@ void UpdateGameState(float dt) {
       SetFlag(SF_UIHIDDEN, !GetFlag(SF_UIHIDDEN));
     }
   }
+  if (Profile::GameSpecific::GameSpecificType == UI::GameSpecificType::CHLCC) {
+    UI::CHLCC::DelusionTrigger::GetInstance().UpdateHeartButtons();
+  }
```
PringlesGang, line 232: Why are we updating both here and in `DelusionTrigger::UpdateShown()`?

src/games/chlcc/delusiontrigger.cpp
```diff
@@ -214,6 +218,84 @@ void DelusionTrigger::PlayClickSound() {
...
+  }
+  if (pointerWentDown && (leftPressed || rightPressed)) {
+    PADinputMouseWentDown &= ~PAD1A;
+    if (leftPressed) {
+      TriggerLeft();
+    } else {
+      TriggerRight();
+    }
+  }
+}
```
PringlesGang, line 309: We could use two `ClickArea`s for most of this

src/games/chlcc/delusiontrigger.cpp
```diff
@@ -226,35 +308,26 @@ void DelusionTrigger::UpdateShown(float dt) {
...
+  const float rightTarget =
+      (ScrWork[SW_DELUSION_STATE] == DS_Negative) ? 0.25f : 0.75f;
+
+  const float fadeSpeed = 10.0f;
+
+  LeftHeartAlpha +=
+      (leftTarget - LeftHeartAlpha) * glm::min(dt * fadeSpeed, 1.0f);
+
+  RightHeartAlpha +=
+      (rightTarget - RightHeartAlpha) * glm::min(dt * fadeSpeed, 1.0f);
```
PringlesGang: With this, the shape of the curve (and the length of the fade) changes with the size of `dt`. Can we not just use two `Animation`s and a `glm::smoothstep` or smth?

src/games/chlcc/delusiontrigger.cpp
```diff
@@ -321,6 +394,31 @@ void DelusionTrigger::UpdateShown(float dt) {
 }
 
 void DelusionTrigger::Update(float dt) {
+  if (Profile::HasDelusionMouseSupport && State != Hidden) {
+    HeartPulseTime += dt;
+    if (DelusionHeartPulseDuration > 0.0f) {
+      HeartPulseTime = std::fmod(HeartPulseTime, DelusionHeartPulseDuration);
+    }
```
PringlesGang: Why not use an `Animation`?

src/games/chlcc/delusiontrigger.cpp
```diff
@@ -321,6 +394,31 @@ void DelusionTrigger::UpdateShown(float dt) {
...
+    targetHeartAlpha = 1.0f;
+  }
+
+  if (State == Hiding || State == Hidden) {
+    targetHeartAlpha = 0.0f;
+  }
+
+  if (State == Showing) {
+    targetHeartAlpha = (AnimationState >= 9) ? 1.0f : 0.0f;
+  }
```
PringlesGang: This should be in a switch

src/games/chlcc/delusiontrigger.cpp
```diff
@@ -321,6 +394,31 @@ void DelusionTrigger::UpdateShown(float dt) {
...
+
+  const float fadeSpeed = 6.0f;
+
+  HeartButtonsAlpha +=
+      (targetHeartAlpha - HeartButtonsAlpha) * glm::min(dt * fadeSpeed, 1.0f);
```
PringlesGang: Same here. Animation and smoothstep

src/games/cclcc/delusiontrigger.cpp
```diff
@@ -65,37 +69,200 @@ bool DelusionTrigger::CheckStartTransitionComplete() {
...
+    const float minY = std::min(std::min(q.TopLeft.y, q.TopRight.y),
+                                std::min(q.BottomLeft.y, q.BottomRight.y));
+    const float maxY = std::max(std::max(q.TopLeft.y, q.TopRight.y),
+                                std::max(q.BottomLeft.y, q.BottomRight.y));
+    return RectF(minX, minY, maxX - minX, maxY - minY);
   };
 
+  constexpr float dragThreshold = 10.0f;
+  constexpr float mouseAdvanceTime = 0.5f;
+  const float spinAngle = MtrgAng / 65535.0f * 2.0f * std::numbers::pi_v<float>;
```
PringlesGang: `ScrWorkAngleToRad(MtrgAng)`

src/games/cclcc/delusiontrigger.cpp
```diff
@@ -65,37 +69,200 @@ bool DelusionTrigger::CheckStartTransitionComplete() {
...
+  RectF positiveHitbox;
+  RectF negativeHitbox;
+  bool hasPositiveHitbox = false;
+  bool hasNegativeHitbox = false;
```
PringlesGang: `std::optional<RectF>`

src/games/cclcc/delusiontrigger.cpp
```diff
@@ -65,37 +69,200 @@ bool DelusionTrigger::CheckStartTransitionComplete() {
...
+
+  if (DelusionState == DS_Neutral) {
+    if ((ScrWork[SW_DELUSION_LIMIT] == Delusion_Both) ||
+        (ScrWork[SW_DELUSION_LIMIT] == Delusion_PosOnly)) {
+      positiveHitbox = quadBounds(PositiveDelusionSprite.ScaledBounds()
+                                      .Rotate(spinAngle, {600, 557})
+                                      .Translate({-800, -109}));
```
PringlesGang: It's a circle, so rotating won't change its hitbox. Now its hitbox will just grow and shrink depending on the angle of the square bounding box

src/games/cclcc/delusiontrigger.cpp
```diff
@@ -65,37 +69,200 @@ bool DelusionTrigger::CheckStartTransitionComplete() {
...
+    const float minX = std::min(std::min(q.TopLeft.x, q.TopRight.x),
+                                std::min(q.BottomLeft.x, q.BottomRight.x));
+    const float maxX = std::max(std::max(q.TopLeft.x, q.TopRight.x),
+                                std::max(q.BottomLeft.x, q.BottomRight.x));
+    const float minY = std::min(std::min(q.TopLeft.y, q.TopRight.y),
+                                std::min(q.BottomLeft.y, q.BottomRight.y));
+    const float maxY = std::max(std::max(q.TopLeft.y, q.TopRight.y),
+                                std::max(q.BottomLeft.y, q.BottomRight.y));
+    return RectF(minX, minY, maxX - minX, maxY - minY);
   };
```
PringlesGang: This should just be a method on `CornersQuad`

src/games/cclcc/delusiontrigger.cpp
```diff
@@ -65,37 +69,200 @@ bool DelusionTrigger::CheckStartTransitionComplete() {
...
+
+  if (ActiveDragHitbox == DragHitbox::None &&
+      (pointerWentDown || DragHoldActive)) {
+    if (hasPositiveHitbox && positiveHitbox.ContainsPoint(curPos)) {
+      ActiveDragHitbox = DragHitbox::Positive;
+    } else if (hasNegativeHitbox && negativeHitbox.ContainsPoint(curPos)) {
+      ActiveDragHitbox = DragHitbox::Negative;
+    }
+    if (ActiveDragHitbox != DragHitbox::None && pointerWentDown) {
+      DragMouseDownTime = dt;
```
PringlesGang: `dt` is not a timestamp what

src/games/cclcc/delusiontrigger.cpp
```diff
@@ -65,37 +69,200 @@ bool DelusionTrigger::CheckStartTransitionComplete() {
...
+      DragPressPending = true;
+      DragHoldActive = false;
+    }
+  }
+
+  if (ActiveDragHitbox == DragHitbox::None) return std::make_pair(false, false);
+
+  const float timeDiff = dt - DragMouseDownTime;
+  if (timeDiff >= mouseAdvanceTime ||
+      ((Input::CurMousePos - Input::PrevMousePos) != glm::vec2(0))) {
```
PringlesGang: I'd put a small buffer on here of a few pixels (preferably dependent on the window size) so it doesn't activate at the smallest tremor

src/games/cclcc/delusiontrigger.cpp
```diff
@@ -65,37 +69,200 @@ bool DelusionTrigger::CheckStartTransitionComplete() {
...
+    PADinputMouseWentDown &= ~PAD1A;
+  }
+
+  if (!DragHoldActive) return std::make_pair(false, false);
+
+  bool dragLeftTrigger = false;
+  bool dragRightTrigger = false;
+  if (std::abs(dragDelta.x) >= dragThreshold) {
+    if (ActiveDragHitbox == DragHitbox::Positive) {
+      if (dragDelta.x > 10.0f && DelusionState == DS_Neutral) {
```
PringlesGang: The `10.0f` should be `dragThreshold`, but I think you just mean to do a simple sign check

src/games/cclcc/delusiontrigger.cpp
```diff
@@ -163,18 +330,21 @@ void DelusionTrigger::Update(float dt) {
...
-  auto leftTrigger = GetControlState(CT_DelusionTriggerL);
-  auto rightTrigger = GetControlState(CT_DelusionTriggerR);
 
   if (LastDelusionState == DelusionState) {
-    if (!leftTrigger && !rightTrigger) return;
-    if (!leftTrigger && rightTrigger) {
-      onRightTrigger();
+    std::pair<bool, bool> triggers = UpdateDraggingTriggers(false, dt);
+
+    if (!triggers.first && !triggers.second) return;
```
PringlesGang: `if (!(triggers.first ^ triggers.second)) return;`
and then just `if (triggers.first) { [...] } else { [...] }`

src/game.cpp
```diff
@@ -227,6 +228,12 @@ void UpdateGameState(float dt) {
...
+  if (Profile::GameSpecific::GameSpecificType == UI::GameSpecificType::CCLCC) {
+    UI::CCLCC::DelusionTrigger::GetInstance().UpdateDragging(dt);
+  }
```
PringlesGang, line 235: Why are we updating both here and in `DelusionTrigger::Update()`?

PringlesGang (changes requested):

src/game.cpp
Enorovan: Because we need to update before ChkMesSkip so that the 1A block is taken into account. Same for C;CLCC

PringlesGang: Ok but then why are we also updating in `DelusionTrigger::UpdateShown`?

src/games/cclcc/delusiontrigger.cpp
```diff
@@ -65,37 +70,185 @@ bool DelusionTrigger::CheckStartTransitionComplete() {
...
+    }
+    return std::make_pair(false, false);
+  }
+
+  constexpr float dragThreshold = 10.0f;
+  constexpr float movementBufferWindowFraction = 0.002f;
+  constexpr float mouseAdvanceTime = 0.5f;
+  const float movementBuffer =
+      std::max(2.0f, std::min(Window->WindowWidth, Window->WindowHeight) *
+                         movementBufferWindowFraction);
```
PringlesGang, line 134: `movementBuffer` is now based in window size, whereas `Input::CurMousePos` is based on design dimensions

src/games/cclcc/delusiontrigger.cpp
```diff
@@ -65,37 +70,185 @@ bool DelusionTrigger::CheckStartTransitionComplete() {
...
+            .Transform(TransformationMatrix({600, 557}, {1.8f, 1.8f},
+                                            {600, 557}, spinAngle, {448, -109}))
+            .BoundingBox();
+  } else if (DelusionState == DS_Negative) {
+    negativeHitbox =
+        NegativeDelusionSprite.ScaledBounds()
+            .Transform(TransformationMatrix({424, 557}, {1.8f, 1.8f},
+                                            {424, 557}, spinAngle, {448, -109}))
+            .BoundingBox();
+  }
```
PringlesGang, line 157: These are still rotated. I'd also recommend using .Scale and .Translate here, because those don't do a full matrix multiplication

src/games/cclcc/delusiontrigger.cpp
```diff
@@ -65,37 +70,185 @@ bool DelusionTrigger::CheckStartTransitionComplete() {
...
+    if (ActiveDragHitbox != DragHitbox::None && pointerWentDown) {
+      DragHoldTime = 0.0f;
+      DragPressPending = true;
+      DragHoldActive = false;
+    }
+  }
+
+  if (ActiveDragHitbox == DragHitbox::None) return std::make_pair(false, false);
+
+  if (DragPressPending && !updateInputBlocking) DragHoldTime += dt;
```
PringlesGang: I don't see the use of `updateInputBlocking` here? This should happen regardless of that, right?

src/games/cclcc/delusiontrigger.cpp
```diff
@@ -163,18 +316,27 @@ void DelusionTrigger::Update(float dt) {
...
+
+    if (!(leftTrigger ^ rightTrigger)) return;
+
+    if (leftTrigger) {
+      TriggerLeft();
+      ActiveDragHitbox = DragHitbox::None;
+    } else {
+      TriggerRight();
+      ActiveDragHitbox = DragHitbox::None;
     }
```
PringlesGang: The dragging part of this is also already done in `UpdateDragging`

src/util.cpp
```diff
@@ -7,6 +7,18 @@
...
 
+RectF CornersQuad::BoundingBox() const {
+  const float minX = std::min(std::min(TopLeft.x, TopRight.x),
+                              std::min(BottomLeft.x, BottomRight.x));
+  const float maxX = std::max(std::max(TopLeft.x, TopRight.x),
+                              std::max(BottomLeft.x, BottomRight.x));
+  const float minY = std::min(std::min(TopLeft.y, TopRight.y),
+                              std::min(BottomLeft.y, BottomRight.y));
+  const float maxY = std::max(std::max(TopLeft.y, TopRight.y),
+                              std::max(BottomLeft.y, BottomRight.y));
```
PringlesGang: `const auto [minX, maxX] = std::minmax({TopLeft.x, TopRight.x, BottomLeft.x, BottomRight.x});`

src/games/chlcc/delusiontrigger.cpp
```diff
@@ -214,6 +230,91 @@ void DelusionTrigger::PlayClickSound() {
...
+    LeftHeartClickArea.UpdateInput(0.0f);
+  } else {
+    LeftHeartClickArea.Hide();
+  }
+  if (RightHeartClickArea.Enabled) {
+    RightHeartClickArea.Show();
+    RightHeartClickArea.UpdateInput(0.0f);
+  } else {
+    RightHeartClickArea.Hide();
+  }
```
PringlesGang:
```cpp
if (ScrWork[SW_DELUSIONSTATE] != DS_Positive) {
  LeftHeartClickArea.Show();
else {
  LeftHeartClickArea.Hide();
}

LeftHeartClickArea.Update(0.0f);
```

src/games/chlcc/delusiontrigger.cpp
```diff
@@ -214,6 +230,91 @@ void DelusionTrigger::PlayClickSound() {
...
+
+  if (Input::CurrentInputDevice == Input::Device::Mouse) {
+    LeftHeartHovered = LeftHeartClickArea.Hovered;
+    RightHeartHovered = RightHeartClickArea.Hovered;
```
PringlesGang: No need for these members if you just use `LeftHeartClickArea.Hovered`

src/games/chlcc/delusiontrigger.cpp
```diff
@@ -226,35 +327,27 @@ void DelusionTrigger::UpdateShown(float dt) {
...
+  const AnimationDirection rightHeartDirection =
+      (ScrWork[SW_DELUSION_STATE] == DS_Negative) ? AnimationDirection::Out
+                                                  : AnimationDirection::In;
+
+  if (LeftHeartFade.Direction != leftHeartDirection) {
+    LeftHeartFade.Start(leftHeartDirection);
+  }
+  if (RightHeartFade.Direction != rightHeartDirection) {
+    RightHeartFade.Start(rightHeartDirection);
   }
```
PringlesGang: No conditional needed for this. That is already handled in the `Start` method itself

src/games/chlcc/delusiontrigger.cpp
```diff
@@ -321,6 +414,36 @@ void DelusionTrigger::UpdateShown(float dt) {
 }
 
 void DelusionTrigger::Update(float dt) {
+  if (Profile::HasDelusionMouseSupport && State != Hidden) {
+    if (DelusionHeartPulseDuration > 0.0f) {
+      const float pulseHalfDuration = DelusionHeartPulseDuration * 0.5f;
+      if (HeartPulseAnimation.DurationIn != pulseHalfDuration) {
+        HeartPulseAnimation.SetDuration(pulseHalfDuration);
+      }
```
PringlesGang: This cannot change at runtime

src/games/chlcc/delusiontrigger.cpp
```diff
@@ -321,6 +414,36 @@ void DelusionTrigger::UpdateShown(float dt) {
...
+      if (!HeartPulseAnimation.IsPlaying()) HeartPulseAnimation.StartIn();
```
PringlesGang: So long as you don't set the `reset` bool, you can call this without the conditional

src/games/chlcc/delusiontrigger.cpp
```diff
@@ -321,6 +414,36 @@ void DelusionTrigger::UpdateShown(float dt) {
...
+      break;
+    case Hiding:
+    case Hidden:
+      heartButtonDirection = AnimationDirection::Out;
+      break;
+  }
+
+  if (HeartButtonFade.Direction != heartButtonDirection) {
+    HeartButtonFade.Start(heartButtonDirection);
+  }
```
PringlesGang: Again, no need for the conditional

PringlesGang (changes requested):

src/games/chlcc/delusiontrigger.cpp
Enorovan: I see the issue of SetDuration in Update but where should I move it? Constructor?

src/games/cclcc/delusiontrigger.cpp
Enorovan: You mean I remove the `ActiveDragHitbox = DragHitbox::None;`s?

src/games/chlcc/delusiontrigger.cpp
PringlesGang: It's already done in the constructor. It just shouldn't do it again

src/games/cclcc/delusiontrigger.cpp
Enorovan: just .Scale().Translate() instead of Transform(....).BoundingBox with all the values? I get why multiplying for nothing in a circle is annoying lol

src/games/cclcc/delusiontrigger.cpp
PringlesGang: No I mean you don't do the `UpdateDraggingTriggers` in `Update`; it's already done in game.cpp

src/games/cclcc/delusiontrigger.cpp
PringlesGang: Yup, that's it

src/util.cpp
```diff
@@ -8,14 +8,10 @@
...
-  const float maxX = std::max(std::max(TopLeft.x, TopRight.x),
-                              std::max(BottomLeft.x, BottomRight.x));
-  const float minY = std::min(std::min(TopLeft.y, TopRight.y),
-                              std::min(BottomLeft.y, BottomRight.y));
-  const float maxY = std::max(std::max(TopLeft.y, TopRight.y),
-                              std::max(BottomLeft.y, BottomRight.y));
+  const auto [minX, maxX] =
+      std::minmax({TopLeft.x, TopRight.x, BottomLeft.x, BottomRight.x});
+  const auto [minY, maxY] =
+      std::minmax({TopLeft.x, TopRight.y, BottomLeft.y, BottomRight.y});
```
PringlesGang: First one here should `TopLeft.x` -> `TopLeft.y`

KKhanhH (approved):

src/games/chlcc/delusiontrigger.cpp
PringlesGang: You didn't actually do this one

PringlesGang (commented):
The hearts don't seem to be pulsating anymore?

For cclcc the dragging seems to be based on how the distance you move your cursor over a small period of time. This means if you hold and move your cursor slowly enough, you can make it across the whole screen without triggering the swipe. Maybe it would be more intuitive if the moment you click, it registers the x position of the cursor, and the trigger activates when you move a certain distance away from that x value while dragging (in the appropriate direction)?

Also, maybe for chlcc you could adjust the "blocking" conditions to extend into the transition when you clicked on one of the buttons, because now if you are, say, in the left delusion and spam-click to go through neutral into right, you still advance a line by clicking during the transition, when the buttons are disabled (very small nitpick, but still thought I'd mention)

Other than that, looks good. Feels very intuitive

PringlesGang (approved):

PR #441, opened 2026-04-05
constant-text-fade-duration

### LINKED ISSUE #408: Text fade duration should be constant per character
Currently, the speed at which an individual character fades in is based on the duration the entire text takes to fade.
Each character should actually fade in at a constant speed, with the duration between two consecutive characters fading in differing based on the duration of the entire text fading.

Enorovan:
Makes it so glyphs start fading in based on the progress of the typewriter, but the fading in of each individual glyph takes a constant amount of time

The fading should be smooth (based on `dt`), and should be resilient to changing the `ConfigSystem::TextSpeed` and `ConfigSystem::SyncVoice` mid-typewriting

Also fixes an issue where if you press 'advance' whilst the typewriter is cancelling (that is, you press 'advance' while it is typewriting so all glyphs start fading in immediately, and then press 'advance' again as these glyphs are still fading in), the cancellation would carry over to the next line as well, immediately cancelling it

---

This is a deceptively complex change, as the glyph fading progress used to be a direct calculation from the animation progress—now only the start time is; the opacity of each glyph is updated accordingly by incrementing and decrementing the opacity dynamically

The `DialoguePage`'s state is dependent on the typewriter's progress, so when reviewing please take careful consideration and verify whether this state machine can not break with this new implementation

\- Pringles


src/text/typewritereffect.cpp
```diff
@@ -10,6 +10,16 @@ namespace Impacto {
...
+namespace {
+float CalcLinearProgress(float progress, float startProgress,
+                         float endProgress) {
+  if (endProgress <= startProgress)
+    return progress >= endProgress ? 1.0f : 0.0f;
+  return std::clamp((progress - startProgress) / (endProgress - startProgress),
+                    0.0f, 1.0f);
+}
+}  // namespace
+
```
PringlesGang: Remove the unnamed namespace

src/text/typewritereffect.cpp
```diff
@@ -90,41 +100,62 @@ TypewriterEffect::ParallelBlock TypewriterEffect::GetParallelBlock(
...
 
-  const float startTime = glyphPropagateDuration * parallelBlockGlyphNo;
-  const float endTime = startTime + singleGlyphDuration;
+float TypewriterEffect::GetSingleGlyphFadeDurationSeconds(
+    float totalDurationSeconds) const {
+  if (totalDurationSeconds <= 0.0f || TextFadeInDuration <= 0.0f) return 0.0f;
 
-  return {startTime / totalDuration, endTime / totalDuration};
+  return std::min(TextFadeInDuration, totalDurationSeconds);
+}
```
PringlesGang: I think this function is a bit overkill for its single use. Should just be a `std::min` at the call-site

src/text/typewritereffect.cpp
```diff
@@ -90,41 +100,62 @@ TypewriterEffect::ParallelBlock TypewriterEffect::GetParallelBlock(
...
 
-  constexpr float singleGlyphDuration = 1.0f;
-  constexpr float glyphPropagateProgress = 0.25f;
-  constexpr float glyphPropagateDuration =
-      singleGlyphDuration * glyphPropagateProgress;
-  const float totalDuration =
-      block.Size * glyphPropagateDuration +
-      (1.0f - glyphPropagateProgress) * singleGlyphDuration;
+  return static_cast<float>(GlyphCount) / Profile::ConfigSystem::TextSpeed;
+}
```
PringlesGang: I don't think the typewriter duration should be relied on here. The user can change the text speed while the line is still appearing, which would change the total duration in seconds of the typewriting effect. The text speed should _only_ affect the speed at which the normalized typewriter progress advances

We should only use the normalized progress here

src/text/typewritereffect.h
```diff
@@ -41,7 +41,12 @@ struct TypewriterEffect : public Animation {
   };
   ParallelBlock GetParallelBlock(size_t glyph);
 
+  float GetTypewriterDurationSeconds() const;
+  float GetSingleGlyphFadeDurationSeconds(float totalDurationSeconds) const;
+  // {startProgress, endProgress} in [0, 1]
+  std::pair<float, float> GetWritingProgresses(
+      size_t glyphNo, size_t glyphCount, float totalDurationSeconds) const;
```
PringlesGang: These two can be standalone functions in the cpp file. They do not need to be members

PringlesGang (changes requested):
Overall structure is quite good

src/text/typewritereffect.cpp
KKhanhH: Make it a static function, thats what unnamed namespace is

PringlesGang: It's already in a cpp file

KKhanhH: That's not what it's for. All functions are technically in a cpp file, the static implies internal linkage which means it won't make it possible to access that function from another translation unit (so better optimizations) since functions are extern by default.

KKhanhH: Like uh if you had a function for CalcLinearProgress(float progress, float startProgress,
                         float endProgress); defined in any other file youd get a linker error

PringlesGang: Oh okay I see, thanks

src/text/typewritereffect.cpp
```diff
@@ -90,41 +98,36 @@ TypewriterEffect::ParallelBlock TypewriterEffect::GetParallelBlock(
...
-std::pair<float, float> TypewriterEffect::GetGlyphWritingProgresses(
-    const size_t glyph) {
-  const ParallelBlock block = GetParallelBlock(glyph);
+std::pair<float, float> GetWritingProgresses(const size_t glyphNo,
+                                             const size_t glyphCount,
+                                             const float totalProgress) {
+  if (glyphCount == 0 || totalProgress <= 0.0f) return {0.0f, 1.0f};
 
-  const size_t parallelBlockGlyphNo = glyph - block.Start;
+  const float glyphFadeProgress = std::min(TextFadeInDuration, totalProgress);
```
PringlesGang: `TextFadeInDuration` is in seconds and `totalProgress` is an arbitrary basis? `glyphFadeProgress` is completely meaningless like this

src/text/typewritereffect.cpp
```diff
@@ -90,41 +98,36 @@ TypewriterEffect::ParallelBlock TypewriterEffect::GetParallelBlock(
   }
 }
 
-std::pair<float, float> TypewriterEffect::GetGlyphWritingProgresses(
-    const size_t glyph) {
-  const ParallelBlock block = GetParallelBlock(glyph);
+std::pair<float, float> GetWritingProgresses(const size_t glyphNo,
```
PringlesGang: Should be static

src/text/typewritereffect.cpp
```diff
@@ -142,13 +145,12 @@ float TypewriterEffect::CalcOpacity(size_t glyph) {
...
-        (ProgressOnCancel - startProgress) / (endProgress - startProgress);
+        CalcLinearProgress(ProgressOnCancel, startProgress, endProgress);
     return glyphProgressBeforeCancellation +
            cancelProgress * (1.0f - glyphProgressBeforeCancellation);
   }
 
-  return std::clamp((Progress - startProgress) / (endProgress - startProgress),
-                    0.0f, 1.0f);
+  return CalcLinearProgress(Progress, startProgress, endProgress);
 }
```
PringlesGang: This function would be a little easier to read if it early-exits on `!IsCancelled` instead

PringlesGang (changes requested):
The glyph start progress should be dependent on `TextSpeed`, and their end progress should be an offset away from that starting progress depending on `TextFadeInDuration` and the currently predicted remaining duration of the line, such that it ends `TextFadeInDuration` seconds after the beginning progress

PringlesGang (approved):
Image

KKhanhH:
For voiced lines, it feels like it's not really fading in, but just kinda blinking into existence (or the fade is super sudden)

PringlesGang:
> For voiced lines, it feels like it's not really fading in, but just kinda blinking into existence (or the fade is super sudden)

What lines specifically? For me voiced lines just kinda fade in as they should. Do you maybe have Sync Voice disabled?

KKhanhH (approved):

Gunoshozo (approved):

PR #443, opened 2026-04-06
C;CLCC: TitleMenu and SystemMenu Fog Effect

### LINKED ISSUE #411: CCLCC system/title menu fog effect
The moving fog sprites should be added to the system menu.
Both there and on the title menu, it should use a different shader (probably additive)

Enorovan:
- [x] Add Fog Animation to System Menu
- [ ] Add  Hard Light shader


src/profile/games/cclcc/systemmenu.cpp
```diff
@@ -45,6 +60,12 @@ void Configure() {
 
   BGTranslationOffset = EnsureGetMember<glm::vec2>("BGTranslationOffset");
 
+  UI::CCLCC::SystemMenu* menu = new UI::CCLCC::SystemMenu();
+
+  menu->SmokeAnimation.LoopMode = AnimationLoopMode::Loop;
+  menu->SmokeAnimation.DurationIn = SmokeAnimationDurationIn;
+  menu->SmokeAnimation.DurationOut = SmokeAnimationDurationOut;
```
PringlesGang: The menu pointer is already created and pushed a couple lines below? The durations are also already set in the constructor

src/profile/games/cclcc/systemmenu.h
```diff
@@ -14,13 +14,26 @@ inline Sprite SystemMenuBG;
...
 inline glm::vec2 MenuEntriesPositions[MenuEntriesNumMax];
 inline RectF MenuEntriesButtonBounds[MenuEntriesNumMax];
 
+inline float SmokeOpacityNormal;
+inline float SmokeX;
+inline float SmokeY;
+inline float SmokeBoundsX;
+inline float SmokeBoundsY;
+inline float SmokeBoundsWidth;
+inline float SmokeBoundsHeight;
```
PringlesGang: Just use a `glm::vec2` and a `RectF`

src/games/cclcc/systemmenu.cpp
```diff
@@ -292,12 +298,39 @@ void SystemMenu::Render() {
...
+  SmokeSprite.Bounds = RectF(
+      SmokeBoundsX, SmokeBoundsY,
+      SmokeBoundsWidth - (SmokeAnimationBoundsXMax * SmokeAnimation.Progress),
+      SmokeBoundsHeight);
+  Renderer->DrawSprite(
+      SmokeSprite,
+      glm::vec2(SmokeBoundsWidth - (SmokeAnimationBoundsXMax *
+                                    (1.0f - SmokeAnimation.Progress)),
+                SmokeY),
+      col);
```
PringlesGang, line 486: I don't really like the idea of modifying this profile variable's bounds...
I'd make a local copy of the sprite and modify that instead

PringlesGang (approved):
Still doesn't use the same shader/blend mode, but that can also come in a different PR I suppose

PR #447, opened 2026-04-08
Exit button nullptr fixes

### LINKED ISSUE #442: Title menu deadlock
Swapping between keyboard and mouse for selecting items causes a game to deadlock

here's a call stack
```cpp
impacto.exe!Impacto::UI::Widget::GetFocus(Impacto::UI::FocusDirection dir) Line 39
impacto.exe!Impacto::UI::Menu::AdvanceFocus(Impacto::UI::FocusDirection dir) Line 75
impacto.exe!Impacto::UI::Menu::UpdateInput(float dt) Line 52 
impacto.exe!Impacto::UI::CCLCC::TitleMenu::UpdateInput(float dt) Line 308
impacto.exe!Impacto::UI::CCLCC::TitleMenu::Update(float dt) Line 336 
```

stuck in this loop
```cpp
 while (nextFocus && !nextFocus->Enabled) {
   nextFocus = nextFocus->FocusElements[dir];
 }
```

PringlesGang:
Fixes the title menu's Exit buttons throwing nullptr exceptions when `HasScriptedExitLogic` is false（ガッ！）

Also fixes an infinite loop in `Widget::GetFocus()` when advancing focus whilst the highlight animation is playing in the cclcc title menu, because _all_ buttons are disabled for this duration

### DIFF
--- a/src/games/chlcc/titlemenu.h
+++ b/src/games/chlcc/titlemenu.h
@@ -54,7 +54,7 @@ class TitleMenu : public Menu {
   Widgets::CHLCC::TitleButton* Load;
   Widgets::CHLCC::TitleButton* Extra;
   Widgets::CHLCC::TitleButton* System;
-  Widgets::CHLCC::TitleButton* Exit;
+  std::optional<std::reference_wrapper<Widgets::CHLCC::TitleButton>> Exit;
 
   Widgets::Group* LoadItems;
   Widgets::CHLCC::TitleButton* SubLoad;
--- a/src/games/cclcc/titlemenu.cpp
+++ b/src/games/cclcc/titlemenu.cpp
@@ -144,11 +144,12 @@ TitleMenu::TitleMenu() {
 
   if (HasScriptedExitLogic) {
     // Exit menu button (Configuration/Patch driven)
-    Exit = new TitleButton(
+    auto* const exitPtr = new TitleButton(
         5, ExitSprite, ExitSprite, ItemHighlightSprite,
         glm::vec2(ItemHighlightOffsetX, (ItemYBase + (5 * ItemPadding))));
+    Exit.emplace(*exitPtr);
     setupBtn(
-        Exit,
+        exitPtr,
         [this](Widgets::Button* target) { return ExitButtonOnClick(target); },
         MainItems, FDIR_DOWN);
   }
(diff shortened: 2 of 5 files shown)


Gunoshozo (approved):
Maybe we can have a debug assert for Widget::GetFocus if we looped back to the start
```cpp
assert(nextFocus != FocusElements[dir])
```
Not that it hard to immideately debug when you're stuck in an infinite loop, idk, just a thought

src/games/chlcc/titlemenu.h
```diff
@@ -54,7 +54,7 @@ class TitleMenu : public Menu {
   Widgets::CHLCC::TitleButton* Load;
   Widgets::CHLCC::TitleButton* Extra;
   Widgets::CHLCC::TitleButton* System;
-  Widgets::CHLCC::TitleButton* Exit;
+  std::optional<Widgets::CHLCC::TitleButton*> Exit;
```
KKhanhH: I think we can just keep it as a raw pointer and check for null

PringlesGang: Yeah I considered that, but because it's then so similar to all other buttons, for which you _don't_ have to check for null each time, I thought it clearer and less error-prone to store it in an `optional` anyway. So you don't have to actively remember, but that the compiler reminds you, that this variable should be treated like a special snowflake

I mean, that's how it slipped past all of us in the initial PR `¯\_(ツ)_/¯`

KKhanhH: yeah but an optional pointer also has double nullability semantics. at the very least it could be in a optional ref wrapper

PringlesGang: True, but you would still then be able to access it in exactly the same way you would a regular pointer, so that wouldn't really warn you when using it the same way you would the others...

KKhanhH: i think you need .get if you want to use the members

PringlesGang: Or actually no, doesn't ref_wrapper require you to `.get()` first?

KKhanhH (approved):

Enorovan (approved):
Good for me for the Exit Part, don't know how to test the other thing but I trust you.

PR #448, opened 2026-04-09
Audio subtitles

KKhanhH:


src/audio/audiosystem.h
```diff
@@ -13,12 +15,19 @@ void AudioInit();
...
+void AudioSubtitlesRender();
+
 inline AudioBackend* Backend = nullptr;
 
 inline float MasterVolume = 1.0f;
 inline std::array<float, ACG_Count> GroupVolumes;
 inline std::array<std::unique_ptr<AudioChannel>, AC_Count> Channels;
 
+// 3 subtitle players for 3 bgm channels
+inline std::array<std::optional<Subtitle::SubtitlePlayer>, 3> SubtitlePlayers;
```
PringlesGang, line 29: Then just replace the 3 with `Channels.size()`

PringlesGang (approved):

src/audio/audiosystem.h
KKhanhH: Channels.size is every channel,not just bgm

PringlesGang: Ah, alright then

Enorovan (approved):

PR #455, opened 2026-04-22
Scroll system bg with backlog scroll in CCLCC

### LINKED ISSUE #406: CCLCC backlog system background should scroll with the menu
The system menu background should scroll up and down as the player is scrolling the entries up and down

https://github.com/user-attachments/assets/5a43545a-8b1c-43cd-9b67-3baebd7adfcf

KKhanhH:


Enorovan (commented):
Two things are wrong:

- It doesn't scroll only on Y axis, it also slightly moves on the X axis
- It looks like the background actuallly curves because of the maths from the frame from system menu.
See on discord for what I mean

EDIT: ✅

src/renderer/opengl/renderer.cpp
```diff
@@ -600,6 +600,13 @@ void Renderer::DrawPrimitives(
   EnsureFBO(fboId.value_or(0));
   EnsureTopologyMode(topologyMode);
 
+  if (textureWrapRepeat) {
+    Flush();
+    glBindSampler(0, Samplers[0]);
+    glSamplerParameteri(Samplers[0], GL_TEXTURE_WRAP_S, GL_REPEAT);
+    glSamplerParameteri(Samplers[0], GL_TEXTURE_WRAP_T, GL_REPEAT);
+  }
```
PringlesGang, line 603: This should maybe also be an `EnsureTextureWrapMode` method or something

src/games/cclcc/systemmenu.cpp
```diff
@@ -235,57 +314,78 @@ void SystemMenu::Render() {
...
+              // Second triangle
+              for (auto v : {v3, v1, v2}) {
+                result[index++] = static_cast<uint16_t>(v);
+              }
+            }
+          }
+          return result;
+        }();
+    Renderer->DrawPrimitives(SystemMenuBG.Sheet, ShaderProgramType::Sprite,
+                             Vertices, indices, glm::vec2(0.0f), false, true);
```
Gunoshozo: Uses regular triangles for indices, can use ~4 times less indices with trinagle strips & degenerate triangles (though current master still uses restart index)

can be something like this:
```cpp
    // triangle strips with degenerate triangles
    static constexpr auto indices = []() {
      constexpr uint16_t width = GridColCount + 1;
      constexpr int totalSize =
          GridRowCount * width * 2 + (GridRowCount - 1) * 2;

      std::array<uint16_t, totalSize> result{};
      size_t index = 0;

      for (uint16_t row = 0; row < GridRowCount; row++) {
        // degenerate triangle here
        if (row > 0) {
          result[index] = result[index - 1];
          index++;
          result[index++] = row * width;
        }
        for (uint16_t col = 0; col < width; col++) {
          result[index++] = row * width + col;
          result[index++] = (row + 1) * width + col;
        }
      }
      return result;
    }();

    Renderer->DrawPrimitives(SystemMenuBG.Sheet, nullptr,
                             ShaderProgramType::Sprite, Vertices, indices,
                             glm::mat4(1.0f), glm::mat4(1.0f), false,
                             TopologyMode::TriangleStrips, 0, true);
```

Gunoshozo (changes requested):

src/games/cclcc/systemmenu.cpp
```diff
@@ -225,6 +225,85 @@ void SystemMenu::Update(float dt) {
...
+
+  for (size_t i = 0; i < vertices.size(); i++) {
+    vertices[i] = VertexBufferSprites{
+        .Position =
+            TransformImageVertex(displayVertices[i], rotation, scale, origin),
+        .UV = spriteVertices[i] / sheetBounds,
+        .Tint = glm::vec4(1.0f),
+    };
+  }
+}
```
PringlesGang, line 302: Maybe we should merge these two functions a bit, so the transformation matrix isn't recomputed for every vertex?

PringlesGang (changes requested):

src/renderer/opengl/renderer.cpp
KKhanhH: It's being unset though, I didn't wanna add even more state into the renderer

KKhanhH: also we are supposed to do a bgfx refactor at some point

Gunoshozo (approved):

Enorovan (approved):

src/renderer/opengl/renderer.cpp
PringlesGang: We could still salvage a lot of the OpenGL renderer for the VitaGL backend tho. Also, at this point I don't really know how far away that bgfx refactor is, so I don't think we should be using that as an excuse to hack things into our current renderer anymore...

KKhanhH: My point is that it's an extra state complexity and an unneeded abstraction for a couple of oneoffs in code that's planned to be refactored. We can add things to the renderer but shouldn't be doing refactors for it unnecessarily.

PringlesGang: I suppose this is something that would only really be used by `DrawPrimitives`, I just don't really like the idea of doing things differently from how every other state change is handled. Also there'd be a slight performance decrease if multiple texture-wrap calls are done in sequence :blini:

PringlesGang (approved):

src/renderer/opengl/renderer.cpp
KKhanhH: It's because it's not being done that I'm not adding it in. If you were to add it in, you now have to add the ensure call to every render function since we assume clamped by default.

PringlesGang: We should really add an `EnsureState` method or something that just takes a big-ass struct

KKhanhH: Bgfx does let you do it either during texture creation or during its bind texture stage

PringlesGang: Specify whether it should be wrapped on a texture-basis, instead of a render call basis, you mean?

KKhanhH: Either or and its a big flags variable for everything

PR #460, opened 2026-04-28
Fix CHLCC Erin Screen Smash Anim

### LINKED ISSUE #361: Erin Smashing Animation

Gunoshozo:


src/character2d.cpp
```diff
@@ -434,8 +434,8 @@ void Character2D::Render(const int layer) {
     }
 
   } else {
-    const glm::mat4 transformation =
-        TransformationMatrix({0.0f, 0.0f}, Scale, glm::vec3(0.0f), Rotation);
+    const glm::mat4 transformation = TransformationMatrix(
+        Position, Scale, glm::vec3(0.0f), Rotation);
```
PringlesGang: In that case the rotation origin is probs also Position. Can't think of many scenes that rotate a character tho, to test

PR #463, opened 2026-05-01
Fix Mask Transformation

### LINKED ISSUE #353: Fix Horizontal Rectangle masking (it goes all the way up like a screen split whereas its upper bound should not be touching the top) (even more broken when upscaling)
Script: AA05 when Takumi, Rimi and Yua search for Seira

What it's supposed to look like:
Image

Enorovan:
Test cases:
- [x] 720
- [x] <720
- [x] 1080
- [x] >16:9
- [x] Changing resolution on the fly
- [x] Non-regression on C;CLCC Backlog/Tips masks

### DIFF
--- a/src/renderer/opengl/renderer.cpp
+++ b/src/renderer/opengl/renderer.cpp
@@ -421,12 +421,14 @@ void Renderer::DrawSprite(const Sprite& sprite, const CornersQuad& dest,
   }
 }
 
-void Renderer::DrawMaskedSprite(
-    const Sprite& sprite, const Sprite& mask, const CornersQuad& spriteDest,
-    const CornersQuad& maskDest, int alpha, const int fadeRange,
-    glm::mat4 spriteTransformation, glm::mat4 maskTransformation,
-    const std::span<const glm::vec4, 4> tints, const bool isInverted,
-    const bool isSameTexture) {
+void Renderer::DrawMaskedSprite(const Sprite& sprite, const Sprite& mask,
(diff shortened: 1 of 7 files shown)


Enorovan:
All tests pass (I can't approve because I created the PR)

src/renderer/opengl/renderer.cpp
```diff
@@ -461,8 +466,7 @@ void Renderer::DrawMaskedSprite(
   // OK, all good, make quad
 
   CornersQuad uvDest = sprite.NormalizedBounds();
-  CornersQuad maskUVDest = CornersQuad(maskDest).Scale(
-      {1.0f / sprite.Bounds.Width, 1.0f / sprite.Bounds.Height}, {0.0f, 0.0f});
+  CornersQuad maskUVDest = RectF(0.0, 0.0, 1.0f, 1.0f);
```
PringlesGang, line 469: This breaks the cclcc backlog outline mask (pretty sure this is what I remembered from some masks still being broken by this PR)

PringlesGang (changes requested):

src/renderer/opengl/renderer.cpp
Gunoshozo: addressed here https://github.com/CommitteeOfZero/impacto/pull/463/changes/43ce54856b1e9b0c8870ec7d6a65b0b8a9b8c15d

PringlesGang (approved):

PR #468, opened 2026-05-10
Fix frame decode not displaying final frames on hw accel desktop

KKhanhH:


KKhanhH:
image

Because github's diff is terrible

Gunoshozo (approved):
Seems to be fixed 👍

PR #471, opened 2026-05-11
Detect default audio device change on windows

egyzol:
Use IMMNotificationClient interface from win32api to get notified of the change and reintialize openal

I tested it on windows 10/11 worked in both cases

If anything is wrong I'll try to fix it

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -955,12 +955,14 @@ if (NOT DEFINED IMPACTO_DISABLE_OPENAL)
         src/audio/openal/audiobackend.cpp
         src/audio/openal/openalaudiochannel.cpp
         src/audio/openal/ffmpegaudioplayer.cpp
+	    src/audio/openal/notificationclient.cpp
     )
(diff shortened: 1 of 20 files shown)


KKhanhH:
Hi @egyzol, thanks for the PR. For this issue I was hoping for a more crossplatform solution than just Windows. I believe OpenAL soft already does have detection built in according to https://github.com/kcat/openal-soft/issues/555, would you be able to use that for the notification implementation instead?

egyzol:
I'll look into that. I made this for windows only because on Ubuntu(PulseAudio) and Linux Mint(PipeWire) it worked without needing to change anything. (Though on mint I got another problem but it was hopefully because it was running in a vm with bad performance, gonna check that as well)

KKhanhH:
Also try to keep to our naming conventions. You might also need to recreate the PR to be from another source branch, the linter doesnt play very well with branches named master.

egyzol:
Closing this in favor of #472

PR #473, opened 2026-05-13
Mask changes to fix multi-image rendering

Gunoshozo:
Move Newo's code to the MaskRender and revert state change for mode 0

### DIFF
--- a/src/background2d.cpp
+++ b/src/background2d.cpp
@@ -378,10 +378,7 @@ void Background2D::SetTransformState(int dispMode, BgTransformState state) {
 
   switch (dispMode) {
     case 0: {
-      const glm::vec2 scaledPos = state.Position / RenderSprite.BaseScale;
-      const glm::vec2 newDimensions = RenderSprite.Bounds.GetSize() - scaledPos;
-      RenderSprite.Bounds.SetPos(scaledPos);
-      RenderSprite.Bounds.SetSize(newDimensions);
+      TransformState.Position = -state.Position;
     } break;
 
     case 1: {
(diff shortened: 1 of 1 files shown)


Enorovan (approved):
Masks still work and bgSetLink work again.

KKhanhH (approved):

src/background2d.cpp
```diff
@@ -807,9 +804,18 @@ void Background2D::RenderRegular() {
 }
 
 void Background2D::RenderMasked() {
+  const glm::vec2 scaledPos = -TransformState.Position / RenderSprite.BaseScale;
+  const glm::vec2 newDimensions = RenderSprite.Bounds.GetSize() - scaledPos;
+  RenderSprite.Bounds.SetPos(scaledPos);
+  RenderSprite.Bounds.SetSize(newDimensions);
```
PringlesGang: This is assuming every call to RenderMasked is with a background under dispMode 0, which I don't think we can

PringlesGang (commented):
Still don't think this would actually work... The whole reason we have to adjust the mask position is because in `case 0` we're setting it to render the sprite off-screen, whereas the mask should render only over the visible part.

Now if the sprite is rendered with the top-left on, say, (100, 100), it would move the mask's UVs farther out as well, whereas in that case it _should_ just render the mask from the top-left of where the sprite is rendered (so a translation of 0)

I think maybe only doing the whole song and dance with the position in `RenderMasked` if the position is negative would work in the general case?
```cpp
maskTransformation.Position =
  glm::min(TransformState.Position, {0.0f, 0.0f}) / designSize;
```

PringlesGang (approved):
Maybe a quick comment explaining why we're doing this so we don't forget, but apart from that all good 👍

PR #474, opened 2026-05-15
Add dt to instruction

KKhanhH:


KKhanhH:
Wondering if it makes more sense to have dt arg be 0 when running thread after createthread or if it should just be dt.

Gunoshozo (approved):
👍

PringlesGang (approved):

PR #480, opened 2026-05-29
Sysmenu Fade for direct Menus

### LINKED ISSUE #273: Fix C;C LCC menu cross-fade out animation
When exiting from a menu (backlog, tips) opened directly from in-game, it should cross-fade between the game visuals and the menu rendered on separate frame buffers

Gunoshozo:


src/games/cclcc/optionsmenu.cpp
```diff
@@ -316,22 +332,28 @@ void OptionsMenu::UpdateInput(float dt) {
 }
 
 void OptionsMenu::Render() {
-  if (State != Hidden && ScrWork[SW_SYSSUBMENUCT] > 0 &&
-      ScrWork[SW_SYSSUBMENUNO] == 5) {
-    const glm::vec4 col(1.0f, 1.0f, 1.0f, FadeAnimation.Progress);
+  if (State != Hidden &&
+      ((ScrWork[SW_SYSSUBMENUCT] > 0 && ScrWork[SW_SYSSUBMENUNO] == 5) ||
```
KKhanhH: How come this is the only one with a different render if

Gunoshozo: I think I can just replace it with `if (State == Hidden) return;` as in other menus

KKhanhH (approved):

src/games/cclcc/commonmenu.h
```diff
@@ -0,0 +1,50 @@
...
+namespace Impacto {
+namespace UI {
+namespace CCLCC {
+
+class CommonMenu {
+ public:
+  using GridVertices =
+      std::array<VertexBufferSprites,
+                 (Profile::CCLCC::SystemMenu::GridRowCount + 1) *
+                     (Profile::CCLCC::SystemMenu::GridColCount + 1)>;
```
PringlesGang: These constexpr variables should then probably be moved to this file as well

src/games/cclcc/optionsmenu.cpp
```diff
@@ -215,14 +216,26 @@ OptionsMenu::OptionsMenu() : UI::OptionsMenu() {
...
+  }
+
   UI::OptionsMenu::Show();
 
-  if (State != Shown) PoleAnimation.StartIn();
+  if (State != Shown) {
+    if (OpenedAsDirect) {
+      PoleAnimation.Finish(AnimationDirection::In);
+    } else {
+      PoleAnimation.StartIn(true);
```
PringlesGang: I don't think this should have the reset bool set. Because the conditional is on `State != Shown`, if we call `Show()` whilst it is already in the showing animation, it would reset the pole animation where it should just continue the animation

src/games/cclcc/commonmenu.h
PringlesGang, line 1: Most of these members could be protected or even private

PringlesGang (approved):

src/games/cclcc/optionsmenu.cpp
Gunoshozo: Yeah, I'd rather need to do this when menu is completely hidden
```cpp
    if (OpenedAsDirect) PoleAnimation.Finish(AnimationDirection::Out);
```

PR #486, opened 2026-06-14
Make the delusion trigger animation skippable

### LINKED ISSUE #484: C;H LCC delusion trigger skipping logic
C;H LCC's delusion trigger still needs logic to have its animations be skipped when skip mode is on

Note that when the "stop skip at delusion trigger" option is selected, skip-read should disable _before_ the show animation is skipped, meaning it will show in full unless interrupted by the user re-enabling skip-read during the animation

egyzol:
Make the delusion trigger animation skippable in C;H LCC #484 

On a side note I noticed that the triggers in impacto make it spin the opposite way compared to the PS3 and switch version is this deliberate?

### DIFF
--- a/src/games/chlcc/delusiontrigger.cpp
+++ b/src/games/chlcc/delusiontrigger.cpp
@@ -79,6 +79,11 @@ void DelusionTrigger::UpdateHiding(float dt) {
     State = Hidden;
     Reset();
   };
+  if (GetFlag(SF_MESSKIP)) {
+    SetFlag(SF_DELUSION_UI_ANIM_WAIT, 0);
+    Reset();
+    return;
+  }
   if (AnimationState < 13) {
     AnimCounter++;
     switch (AnimationState) {
@@ -153,6 +158,10 @@ void DelusionTrigger::UpdateHiding(float dt) {
 }
 
 void DelusionTrigger::UpdateShowing(float dt) {
+  if (GetFlag(SF_MESSKIP)) {
+    Load();
+    return;
+  }
   switch (AnimationState) {
     case 0: {
       SpinRate -= 24;
@@ -539,7 +548,7 @@ void DelusionTrigger::Load() {
   UnderlayerXOffset = 20000;
   AnimCounter = 0;
   SpinAngle = 0;
-  SpinRate = 5;
+  SpinRate = -5;
   BackgroundAlpha = 256;
   TextSystem.Clear();
   switch (DelusionState) {
--- a/src/games/chlcc/delusiontrigger.cpp
+++ b/src/games/chlcc/delusiontrigger.cpp
@@ -72,13 +72,6 @@ void DelusionTrigger::Hide() {
 }
 
 void DelusionTrigger::UpdateHiding(float dt) {
-  auto setHidden = [this] {
-    AnimCounter = 0;
-    AnimationState = 0;
-    SetFlag(SF_DELUSION_UI_ANIM_WAIT, 0);
-    State = Hidden;
-    Reset();
-  };
   if (AnimationState < 13) {
     AnimCounter++;
     switch (AnimationState) {
@@ -125,7 +118,7 @@ void DelusionTrigger::UpdateHiding(float dt) {
           BackgroundAlpha -= 16;
         }
         if (AnimCounter == 50) {
-          setHidden();
+          Reset();
         }
       } break;
       case 11: {
@@ -135,7 +128,7 @@ void DelusionTrigger::UpdateHiding(float dt) {
           BackgroundAlpha -= 8;
         }
         if (AnimCounter == 100) {
-          setHidden();
+          Reset();
         }
       } break;
       case 12: {
@@ -145,7 +138,7 @@ void DelusionTrigger::UpdateHiding(float dt) {
           BackgroundAlpha -= 8;
         }
         if (AnimCounter == 100) {
-          setHidden();
+          Reset();
         }
       } break;
     }
@@ -532,6 +525,7 @@ void DelusionTrigger::Render() {
 }
 
 void DelusionTrigger::Load() {
+  SetFlag(SF_DELUSION_UI_ANIM_WAIT, 0);
   State = Shown;
   ShakeState = 0;
   MaskOffsetX = 0;
@@ -576,6 +570,7 @@ void DelusionTrigger::Load() {
 }
 
 void DelusionTrigger::Reset() {
+  SetFlag(SF_DELUSION_UI_ANIM_WAIT, 0);
   State = Hidden;
   SpinAngle = 62431;
   SpinRate = 0;
@@ -593,21 +588,6 @@ void DelusionTrigger::Reset() {
   LeftHeartFade.Finish(AnimationDirection::In);
   RightHeartFade.Finish(AnimationDirection::In);
   TextSystem.Clear();
-}
-
-void DelusionTrigger::SetHidden() {
-  if (GetFlag(SF_MESALLSKIP) && State == Hiding) {
-    Reset();
-    SetFlag(SF_DELUSION_UI_ANIM_WAIT, 0);
-  }
-}
-
-void DelusionTrigger::SetShown() {
-  if ((SkipModeEnabled ||
-       GetControlState(CT_ForceSkip, InputDownType::IsDown)) &&
-      GetFlag(SF_MESALLSKIP) && State == Showing) {
-    Load();
-  }
 }
 
 DelusionTextSystem::DelusionTextSystem() {
--- a/src/games/chlcc/delusiontrigger.cpp
+++ b/src/games/chlcc/delusiontrigger.cpp
@@ -45,8 +45,8 @@ void DelusionTrigger::Show() {
     State = Showing;
     DelusionState = DS_Neutral;
     MaskScaleFactor = 131072;
-    SpinAngle = 0;
-    SpinRate = 3072;
+    SpinAngle = 62431;
+    SpinRate = -3072;
     UnderlayerAlpha = 0;
     BackgroundAlpha = 0;
     AnimationState = 0;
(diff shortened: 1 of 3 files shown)


Enorovan:
Hello, thanks for the PR

Two things : 

- Please squash your commits
- You're right about the startup portion being inversed, it'd be kind if you could fix it (it's probably just a sign error?

Enorovan:
Oh also, you correctly skipped the startup section, but the "locking in/zooming in" section should also be skipped (at the end of the trigger)

egyzol:
Happy to help where I can

I didn't mean just that the startup portion of the trigger is inverted I meant that on the ps3 and switch the pressing the right trigger makes it spin left and on impacto pressing the right trigger makes it spin right

Enorovan:
> Happy to help where I can
> 
> I didn't mean just that the startup portion of the trigger is inverted I meant that on the ps3 and switch the pressing the right trigger makes it spin left and on impacto pressing the right trigger makes it spin right

Hmm I'm not remembering this but in any case, no there's no reasoning behind our implem being different so it should be fixed

Enorovan:
I think the startup section is also inversed, i should load up clockwise, then left-right-left, then continue clockwise

Right now it does counterclockwise, right-left-right, continues clockwise, so only step 5 is correct

egyzol:
I found the fix for it but now I also found that it's still not quite alligned with the other versions so I am going to tinker with it a bit more
I'll commit it when it's fully done

egyzol:
It should be really close to the original now
Also added a small delay to the end of the dt when skipping

Enorovan (approved):

src/games/chlcc/delusiontrigger.cpp
```diff
@@ -79,6 +79,18 @@ void DelusionTrigger::UpdateHiding(float dt) {
     State = Hidden;
     Reset();
   };
+  if (GetFlag(SF_MESSKIP)) {
```
PringlesGang: I think this and the other one should be `SF_MESALLSKIP`, as you should not be able to skip the intro with the advance button

PringlesGang (changes requested):
Thanks for the PR, looks great for the most part!

In addition to the other comment, the skip-exiting of the trigger also does not seem to be as smooth yet as it should be. There's a noticeable delay between the animation being skipped and the new dialogue actually showing up, and if you let go of force-skip halfway through that delay it does not automatically load the next line

If you could fix those remaining issues up, that would be great. Thanks again

egyzol:
I added the delay because it felt like it had some on the other platforms altough it could just be my eyes playing tricks on me.
I could remove the delay alltogether or make it shorter. Which one would you prefer?

PringlesGang:
> I added the delay because it felt like it had some on the other platforms altough it could just be my eyes playing tricks on me. I could remove the delay alltogether or make it shorter. Which one would you prefer?

I did some digging for myself and I've figured out how the actual game does it. Bear with me, this may get a little technical

When hiding the delusion trigger, in the script (`_system.scs`) it enters a tight loop that keeps calling `InstDelusionTriggetCHLCC` type 7 until flag 2511, `SF_DELUSION_UI_ANIM_WAIT`, is reset to false:
```
46:
	Mwait 1, 0
	DelusionTrigger_07
	FlagOnJump 2511, 46  // Jumps back to label 46 if flag 2511 is true
	Jump 42
```
In the original game, `DelusionTrigger_07` keeps track of a counter which is incremented on each invocation, and handles the hiding animation. In Impacto this is done in `DelusionTrigger::UpdateHiding()` instead. These functions are pretty one-to-one implementation-wise, so those should probably not be updated for this

The skipping is handled by the `Mwait` instruction, which in this case waits for a single frame, except if `SF_MESALLSKIP` is true, in which case it just keeps going. This means that if the game is skipping, it will rapidly call `DelusionTrigger_07`, near-instantly setting its counter to 100 which will have it trip the `SF_DELUSION_UI_ANIM_WAIT` flag, and thus exit the loop, skipping the animation

Because we handle the hiding animation in `DelusionTrigger::UpdateHiding()`, which is called through `DelusionTrigger::Update()` instead of `InstDelusionTriggerCHLCC`, the repeated calls to this instruction will not do anything, and thus Impacto does not skip

I think the cleanest solution would be to extract `setHidden()` into a public method on the `CHLCC::DelusionTrigger` class and adding
```cpp
case 7:
  if (GetFlag(SF_MESALLSKIP) {
    inst.SetHidden();
  }
  break;
```
to `InstDelusionTriggerCHLCC`

Something similar would likely need to be done for skipping the showing animation and case 2

Let me know if anything's unclear, I understand this may be a lot

egyzol:
It's fairly straight forward thank you for the explanation.
Now I have some other questions what's the difference between SF_MESALLSKIP and SF_MESSKIP I'm guessing that MESALLSKIP is skip mode and MESSKIP is the player pressing advance but then is there a way to differentiate between skip mode and force skip?

PringlesGang:
> It's fairly straight forward thank you for the explanation. Now I have some other questions what's the difference between SF_MESALLSKIP and SF_MESSKIP I'm guessing that MESALLSKIP is skip mode and MESSKIP is the player pressing advance but then is there a way to differentiate between skip mode and force skip?

That's right, `SF_MESALLSKIP` is for force skip, skip all, and an _active_ skip read (i.e., skip mode is on, the config is set to skip read, and the line is actually read).
`SF_MESSKIP` is for any reason the line would be skipped, whether that be through a skip like `SF_MESALLSKIP` or through the advance button.
This logic is all handled in [`ChkMesSkip`](https://github.com/CommitteeOfZero/impacto/blob/68d7dacfea45d3836453966d45174027e410877f/src/vm/inst_dialogue.cpp#L866) if you wanna take a closer look

There isn't really a ScrFlag for differentiating skip mode from force skip, but neither does the engine ever make a difference through the script. Within the engine code that difference would be `GetControlState(CT_SkipMode)` vs. `GetControlState(CT_ForceSkip)`

egyzol:
Is this how you imagined to work?

src/vm/inst_gamespecific.cpp
```diff
@@ -1036,8 +1036,12 @@ VmInstruction(InstDelusionTriggerCHLCC) {
       inst.State = UI::Shown;
       break;
     case 2:
+      inst.SetShown();
+      BlockThread;
+      break;
```
PringlesGang, line 1042: I don't think this should actually block the thread

src/games/chlcc/delusiontrigger.cpp
```diff
@@ -594,6 +595,21 @@ void DelusionTrigger::Reset() {
   TextSystem.Clear();
 }
 
+void DelusionTrigger::SetHidden() {
+  if (GetFlag(SF_MESALLSKIP) && State == Hiding) {
+    Reset();
+    SetFlag(SF_DELUSION_UI_ANIM_WAIT, 0);
+  }
+}
```
PringlesGang: I more so meant you extract the `setHidden` lambda in `UpdateHiding` into this method and call that instead (also in `UpdateHiding`)

I also think it's cleaner to do the `GetFlag(SF_MESALLSKIP)` check in the VM instruction instead, for both `SetHidden` and `SetShown`

src/games/chlcc/delusiontrigger.cpp
```diff
@@ -594,6 +595,21 @@ void DelusionTrigger::Reset() {
...
+
+void DelusionTrigger::SetShown() {
+  if ((SkipModeEnabled ||
+       GetControlState(CT_ForceSkip, InputDownType::IsDown)) &&
+      GetFlag(SF_MESALLSKIP) && State == Showing) {
+    Load();
+  }
+}
```
PringlesGang: This should probably also `SetFlag(SF_DELUSION_UI_ANIM_WAIT, false)` and reset the `AnimationState`

`GetFlag(SF_MESALLSKIP)` already covers `(SkipModeEnabled || GetControlState(CT_ForceSkip, InputDownType::IsDown)` btw

PringlesGang (changes requested):
Yeah you got the idea I was going for, but I've still got a small handful of pointers

src/games/chlcc/delusiontrigger.cpp
egyzol: The problem here is that SF_MESALLSKIP gets set to false after SetShown has been called thus skipping the animation and skipping only stops on the first line of the dt

src/vm/inst_gamespecific.cpp
egyzol: When I first tried it without the game hang on a black screen when the animation supposed to be playing and when I put it back it worked, now I tried it again and hang on black screen then I tried it again, it now works so idk

KKhanhH: Blockthread is correct, thats how it did it before (with a fallthrough

src/games/chlcc/delusiontrigger.cpp
egyzol: Putting a BlockThread after inst.Show() seems to fix it

src/vm/inst_gamespecific.cpp
PringlesGang: That's weird, I don't see it setting `SCRflag` anywhere in this instruction in the decomp

src/games/chlcc/delusiontrigger.cpp
```diff
@@ -594,6 +588,16 @@ void DelusionTrigger::Reset() {
...
 
+void DelusionTrigger::SetHidden() {
+  SetFlag(SF_DELUSION_UI_ANIM_WAIT, 0);
+  Reset();
+}
+
+void DelusionTrigger::SetShown() {
+  SetFlag(SF_DELUSION_UI_ANIM_WAIT, 0);
+  Load();
+}
```
PringlesGang: Looking at these now, I think you can safely move the `SetFlag`s into `Reset()` and `Load()` and just use those everywhere instead, removing `SetShown()` and `SetHidden()` altogether

src/vm/inst_gamespecific.cpp
KKhanhH: Its not gonna be 1 for 1 since we update at main level rather than in the instruction level. We could try moving the update here, now that dt is part of the thread context and see how that changes things.

PringlesGang: Ah right that makes sense, yeah

PringlesGang (approved):

KKhanhH (approved):

PR #490, opened 2026-06-21
Allow sprite sheets to be loaded by the script

Gunoshozo:
- Allow sprite sheets to be loaded by the script, which offloads preloading them at the start of the game
- Speeds up Switch's start up time & eases RAM strain during start up ( sprite sheets need to be marked as `ScriptHandled = true` in the lua)

### DIFF
--- a/src/renderer/opengl/renderer.cpp
+++ b/src/renderer/opengl/renderer.cpp
@@ -9,6 +9,7 @@
 #include "3d/scene.h"
 #include "yuvframe.h"
 #include "nv12frame.h"
+#include <ankerl/unordered_dense.h>
 
 #ifndef IMPACTO_DISABLE_IMGUI
 #include <imgui_custom/backends/imgui_impl_opengl3.h>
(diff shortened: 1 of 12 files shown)


src/renderer/opengl/renderer.cpp
```diff
@@ -178,6 +179,74 @@ void Renderer::EndFrame() {
...
+bool Renderer::LoadSurf(int surfId, int archiveId, int fileId) {
+  auto path = Io::AssetPathKey{.MountId = static_cast<uint32_t>(archiveId),
+                               .Id = static_cast<uint32_t>(fileId)};
+  auto innerTextureIdIter = SheetPathToId.find(path);
+  if (innerTextureIdIter == SheetPathToId.end()) {
+    return false;
+  }
+  SurfToId.try_emplace(surfId, path);
+
+  auto innerTextureId = innerTextureIdIter->second;
```
KKhanhH: maybe name these lookupTextureId

src/renderer/opengl/renderer.cpp
```diff
@@ -1216,5 +1311,27 @@ void Renderer::EnsureTopologyMode(TopologyMode newMode) {
   }
 }
 
+uint32_t Renderer::UnwrapTextureId(Sprite const& sprite) {
```
KKhanhH, line 1307: this could probably just call the overload for sheet with its sheet member.

KKhanhH (approved):

src/renderer/opengl/renderer.cpp
```diff
@@ -178,6 +179,74 @@ void Renderer::EndFrame() {
   glBindSampler(0, 0);
 }
 
+static ankerl::unordered_dense::map<Io::AssetPath, uint32_t, Io::AssetPathHash,
+                                    Io::AssetPathEqual>
+    SheetPathToId;
+
+static ankerl::unordered_dense::map<uint32_t, uint32_t> IdToTexture;
+
+static ankerl::unordered_dense::map<int, Io::AssetPathKey> SurfToId;
```
PringlesGang: You also have these as static members of the base renderer

src/renderer/opengl/renderer.cpp
```diff
@@ -178,6 +179,74 @@ void Renderer::EndFrame() {
...
+
+uint32_t Renderer::MapSpriteSheet(SpriteSheet const& sheet) {
+  if (!sheet.ScriptHandled) {
+    assert(false);
+  }
```
PringlesGang, line 186: `assert(sheet.ScriptHandled);` or
```cpp
if (!sheet.ScriptHandled) {
  assert(false);
  return 0;
}
```

PringlesGang (approved):

PR #491, opened 2026-06-21
DT BgWave Fix

### LINKED ISSUE #489: Add WaveEffect below DelusionTrigger
Everything below the Delusion Trigger composition should WaveEffect during delusion triggers (sprites, background, etc.)

Gunoshozo:
- Update `FadeCt` based on the `BackgroundAlpha` field from DT
- Override ScriptVar from common and fix extra `-1`


Enorovan (approved):

profiles/chlcc/scriptvars.lua
```diff
@@ -95,6 +95,7 @@ sv.SW_BG1NO = 2407;
 sv.SW_BG1PRI = 2408;
 sv.SW_BG1PRI2 = 2408; -- Effectively disable
 sv.SW_BG1DISPMODE = 2409;
+sv.SW_CAP1FADECT = 2410;
 sv.SW_BG1FADECT = 2410;
```
PringlesGang, line 99: The same variable?

PringlesGang (commented):
It also looks like it sorta cross-fades between the waved image and the still one when ending the delusion trigger, whereas for us it just kinda blips out

Enorovan (changes requested):
Fix da crossfade

profiles/chlcc/scriptvars.lua
Gunoshozo: - DT code uses `SW_CAP1FADECT`
- `SW_CAP1FADECT` was inherited from common which is `5010`
- in the binary decomp `2410` is used as a base offset

PringlesGang (approved):

Enorovan (approved):

KKhanhH (approved):

PR #492, opened 2026-06-24
Add cap to backlog entries

PringlesGang:
Stores the backlog entries in a circular buffer and adds a maximum number of entries to the profile. When full, the oldest entry will be overwritten by the new entry.

The maximum is profile-configurable and defaulted to 400, as per the MAGES. engine.

- [x] Add cap to backlog entries
- [x] Add Boost::circular_buffer dependency
- [x] Turn backlog class's scrollbar from a `Scrollbar*` to simply a `Scrollbar`
- [x] Have scrolling down to close the backlog when at the bottom be based on a short timer, so scrolling to the bottom of the backlog does not immediately close it on high FPS


KKhanhH (changes requested):
Instead of circular buffer of unique_ptrs of backlog entries (of which there is only one type tied to the game, so the polymorphism support isn't needed), could we have a circular buffer of concrete classes instead? for 400 entries I could imagine this might add some performance hit to it. If we wanna avoid code duplication we could maybe do a CRTP type class where backlog entry type is one of the template args

Gunoshozo (approved):

PringlesGang:
> Instead of circular buffer of unique_ptrs of backlog entries (of which there is only one type tied to the game, so the polymorphism support isn't needed), could we have a circular buffer of concrete classes instead? for 400 entries I could imagine this might add some performance hit to it. If we wanna avoid code duplication we could maybe do a CRTP type class where backlog entry type is one of the template args

Yeah I was also not happy having the circular buffer store pointers, so I'd love to fix this. I've never heard of CRTP before and I don't fully understand how this would work here yet.

Your main idea is to make the backlog entry type a template parameter, right? But for the generic `inline BacklogMenu* BacklogMenuPtr` we cannot give generic template parameters, so I don't see how CRTP would work then either...

KKhanhH:
> Your main idea is to make the backlog entry type a template parameter, right? But for the generic `inline BacklogMenu* BacklogMenuPtr` we cannot give generic template parameters, so I don't see how CRTP would work then either...

You would probably need an intermediate class that derives from BacklogMenu that contains the queue and has all the methods to act on the queue. Then game classes would derive from that.

```cpp

template <typename Derived, typename BacklogEntry>
class BacklogMenuBase : public BacklogMenu {
 protected:
   // show entries methods, hide entries, render entries, etc,
   boost::circular_buffer<BacklogEntry> Entries;
};

class CC::BacklogMenu : public BacklogMenuBase<CC::BacklogMenu, CC::BacklogMenuEntry> {
 //...
};

```

KKhanhH:
Well the derived part might not be needed if you don't need to call anything on the derived classes, it's mainly useful if you do in the form of `static_cast<Derived*>(this)->method`

KKhanhH (approved):

PR #493, opened 2026-06-26
[CCLCC] Tips menu bugfixes.

dendi2008:
This PR will fix (Screenshots were made before changes to show bugs in PR):

- Tip can be hovered even if can not see it.
{BCE66071-8E5D-484E-B130-6AFB5AFD82E0}

- Tip highlight not reaching last visible Tip when holding/tapping (not fast) down with keyboard/gamepad.
{8B9B44C8-4DC9-43F0-A213-BBF82532B181}

### DIFF
--- a/src/ui/widgets/cclcc/tipsentrybutton.cpp
+++ b/src/ui/widgets/cclcc/tipsentrybutton.cpp
@@ -75,7 +75,7 @@ void TipsEntryButton::Update(float dt) {
 }
 
 void TipsEntryButton::UpdateInput(float dt) {
-  if (TipsTabBounds.Intersects(Bounds) || TipsTabBounds.Contains(Bounds)) {
+  if (TipsTabBounds.Contains(Bounds)) {
     Button::UpdateInput(dt);
   } else {
     Hovered = false;
--- a/src/ui/widgets/cclcc/tipstabgroup.cpp
+++ b/src/ui/widgets/cclcc/tipstabgroup.cpp
@@ -69,8 +69,7 @@ TipsTabGroup::TipsTabGroup(
 
   TipsScrollTrackBounds = {TipsScrollThumbSprite.Bounds.Width,
                            TipsScrollYEnd - TipsScrollYStart};
-  EntriesPerPage =
-      (int)std::ceil(TipsTabBounds.Height / TipsEntryBounds.Height);
+  EntriesPerPage = (int)std::ceil(TipsTabBounds.Height / GetEntryStride());
 }
 
 void TipsTabGroup::UpdatePageInput(float dt) {
@@ -106,7 +105,7 @@ void TipsTabGroup::UpdatePageInput(float dt) {
           if (CurrentlyFocusedElement == TipsEntriesGroup.Children.front()) {
             ScrollPosY = 0;
           } else {
-            ScrollPosY += TipsEntryBounds.Height;
+            ScrollPosY += GetEntryStride();
           }
         }
       } else {
@@ -115,7 +114,7 @@ void TipsTabGroup::UpdatePageInput(float dt) {
           if (CurrentlyFocusedElement == TipsEntriesGroup.Children.back()) {
             ScrollPosY = TipsEntriesScrollbar->EndValue;
           } else {
-            ScrollPosY -= TipsEntryBounds.Height;
+            ScrollPosY -= GetEntryStride();
           }
         }
       }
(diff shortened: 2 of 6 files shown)


src/ui/widgets/cclcc/tipstabgroup.cpp
```diff
@@ -84,7 +84,9 @@ void TipsTabGroup::UpdatePageInput(float dt) {
     }
 
     auto checkScrollBounds = [&]() {
-      return !TipsTabBounds.Contains(CurrentlyFocusedElement->Bounds);
+      const float entryTop = CurrentlyFocusedElement->Bounds.Top();
+      return entryTop < TipsTabBounds.Top() ||
+             entryTop >= TipsTabBounds.Bottom();
     };
```
PringlesGang, line 87: Thanks for another PR Dendi! This does work and fixes the issues, but I think the real underlying issue is that the height of `TipsEntryBounds` should be 40 but is now 50, as it seemingly includes the 10-pixel vertical padding between entries. Some uses of the entries' bounds probably expect the padding to be there, whereas these two uses clearly don't

I think a more idiomatic solution would thus be to reduce `TipsEntryBounds` to 40 and add a new profile variable `TipsEntryYPadding = 10` and add those for the individual instances that _do_ make use of the vertical padding. For the hover issue you'd then probably also need to change the first line of `TipsEntryButton::UpdateInput` to only check against `Contains` (not `Intersects`), as it should only allow to be hovered if the entire entry is in the visible bounds

If you could see if that would also fix these issues, that'd be awesome

dendi2008: > I think a more idiomatic solution would thus be to reduce `TipsEntryBounds` to 40 and add a new profile variable `TipsEntryYPadding = 10` and add those for the individual instances that _do_ make use of the vertical padding. For the hover issue you'd then probably also need to change the first line of `TipsEntryButton::UpdateInput` to only check against `Contains` (not `Intersects`), as it should only allow to be hovered if the entire entry is in the visible bounds

Done.

src/ui/widgets/cclcc/tipsentrybutton.cpp
```diff
@@ -75,10 +75,20 @@ void TipsEntryButton::Update(float dt) {
...
+    } else if (Input::CurrentInputDevice == Input::Device::Touch &&
+               Input::TouchIsDown[0]) {
+      Hovered = inBounds && Bounds.ContainsPoint(Input::CurTouchPos);
+    }
+    if (OnClickHandler && HasFocus &&
+        ((Hovered &&
+          Vm::Interface::PADinputMouseWentDown & Vm::Interface::PAD1A) ||
+         (Vm::Interface::PADinputButtonWentDown & Vm::Interface::PAD1A))) {
+      OnClickHandler(this);
+    }
```
PringlesGang: I actually meant you can just leave it as
```cpp
void TipsEntryButton::UpdateInput(float dt) {
  if (TipsTabBounds.Contains(Bounds)) {
    Button::UpdateInput(dt);
  } else {
    Hovered = false;
  }
}
```
De-duplicates the code a bit

src/ui/widgets/cclcc/tipstabgroup.h
```diff
@@ -50,6 +50,7 @@ class TipsTabGroup : public Menu {
   glm::vec2 TipsScrollStartPos;
   glm::vec2 TipsScrollTrackBounds;
   int EntriesPerPage;
+  float EntryStride;
```
PringlesGang: Not the biggest fan of a member for a constant value that is not dependent on any internal state... Maybe you could replace the member with a static method?
```cpp
static float GetEntryStride() {
  using namespace Profile::CCLCC::TipsMenu;
  return TipsEntryBounds.Height + TipsEntryYPadding;
}
```
If you define it in the header, the compiler should then just inline it at all call sites

PringlesGang (commented):
Looks great! I got two more pointers for code cleanliness, but apart from that I'm happy

src/ui/widgets/cclcc/tipsentrybutton.cpp
dendi2008: Done.

src/ui/widgets/cclcc/tipstabgroup.h
dendi2008: Done.

PringlesGang (approved):

PR #494, opened 2026-07-10
Patchwork profile

### LINKED ISSUE #415: File Redirection for patch assets
patch.lua

KKhanhH:
Improve config fallback, profile handling, and patch integration.

- Platform-aware config directory resolution
- Centralizing startup configuration to be in three primary components: base paths, game definition, user configuration
- Changes folder directories to be more configurable, decoupling gamedata and savedata (this is a breaking change, install tree will need an update)
- Updating Lua path handling to rely on the base paths variables
- VFS improvements for patch extensibility and base game config
- Moving patch-specific content out of included profiles


src/io/filemeta.cpp
```diff
@@ -64,5 +64,40 @@ IoError GetFilePermissions(std::string const& path,
...
+      exit(1);
+    }
+    return configPath;
+#elif defined(__unix__)  // prefer .config over .local/share
+    std::filesystem::path result;
+    const char* configPath = SDL_getenv("XDG_CONFIG_HOME");
+    if (!configPath || *configPath == '\0') {
+      configPath = SDL_getenv("HOME");
+      if (!configPath || *configPath == '\0') {
+        return std::string{};
```
PringlesGang, line 85: Should this not throw an error?

src/io/filemeta.cpp
```diff
@@ -64,5 +64,40 @@ IoError GetFilePermissions(std::string const& path,
...
+      result = std::filesystem::path(configPath) / ".config";
+    } else {
+      result = std::filesystem::path(configPath);
+    }
+    return (result / "impacto").string();
+#else
+    char* configPath = SDL_GetPrefPath("Committee of Zero", "Impacto");
+    std::string result = configPath;
+    SDL_free(configPath);
+    return result;
```
PringlesGang: You should probably check against `configPath == nullptr` here and throw if something went wrong

src/io/filemeta.cpp
KKhanhH: I made it a graceful fallback to just using the one inside install directory but it's such a rare and impossible scenario anyway.

PringlesGang (approved):
lgtm

KKhanhH:
@dendi2008 @egyzol merging this PR soon, it's gonna change the file tree layout and configuration for impacto a little bit so just letting you all know you'll need to change it. Updated docs should tell you how to set it up

PR #495, opened 2026-07-13
C;C LCC system and title menu graphics

### LINKED ISSUE #411: CCLCC system/title menu fog effect

PringlesGang:
- [x] Adds smoke effect to system and title menu and their submenus
- [x] Adds hard-light overlay to system and title menu and their submenus
- [x] Adds some smooth fade transitions for some sprites of the title menu
- [x] Cleans up some title menu profile stuff

Closes #411

I think this PR is easiest reviewed commit-by-commit


Enorovan (approved):
All good.

Gunoshozo (approved):

src/games/cclcc/titlemenu.cpp
KKhanhH, line 600: I'm wondering if it would make more sense if we just kept this lookup map inside title and system menu and then rendered and maybe update all the submenus in those classes. then it would be closer to what they do and we wouldn't have to put render overlay and smoke on every sub menu (just put them after the submenu render).

src/vm/inst_misc.cpp
```diff
@@ -528,20 +529,22 @@ VmInstruction(InstTitleMenuNew) {
           break;
         case InstructionSet::CC:
         case InstructionSet::CHN: {
-          if (ScrWork[SW_TITLEMODE] == 3) {
+          using enum UI::CCLCC::TitleMenuMode::Mode;
```
KKhanhH: Not entirely a fan of using CCLCC specific enums here if this will also apply for cc and chn (enum values will also be different in the LCC double pack too)

src/games/cclcc/titlemenu.cpp
PringlesGang: We couldn't just render the overlay and smoke overtop the submenus, because those effects have to render _behind_ the button guide sprite

Could still move the update and render calls over if you want tho

src/vm/inst_misc.cpp
PringlesGang: I think CHN is actually supposed to have its own implementation but is now leeching off of CC's to try to get it booted in the first place, so we can disregard that one

For CC I had a quick look and the values seem to be mostly identical, apart from that CC also has a value 10 but doesn't have a value 2. All values they _do_ have in common seem to share the same semantic meaning, so I think this shared enum is fine

KKhanhH: Maybe move it into CC then

KKhanhH (approved):

PR #498, opened 2026-07-20
[CCLCC] Icon not showing on Windows 11 taskbar fix.

dendi2008:
This PR will fix CCLCC icon not showing on Windows 11 taskbar.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -137,8 +137,6 @@ static void Init() {
     Effects::Blur.Init();
   }
 
-  SetWindowIcon(Window->SDLWindow);
-
   if (+Profile::GameFeatures & +GameFeature::ModelViewer) {
     ModelViewer::Init();
   }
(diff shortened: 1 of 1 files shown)


KKhanhH:
I'm not so sure this is really necessary and if it's mostly an issue during development builds or if it's something that actually occurs often with release builds.

dendi2008:
> I'm not so sure this is really necessary and if it's mostly an issue during development builds or if it's something that actually occurs often with release builds.

This bug appears both on dev. build and on release build. This bug also could be fixed by creating shortcut for impacto.exe with icon during patch/translation installation procces. It's not even implementation error. It's an Windows 11 only taskbar bug. Even without changes in code, icon can appear on Windows 11 taskbar, if you just reset taskbar (restarting explorer in task manager). And icon already appears in task manager, alt+tab window and Windows 11 sound settings. Maybe, it's caused by initialization time, because bug doesn't appear on CHLCC profile, because initialization proccess taking ~0.5 seconds, but CCLCC taking ~3 seconds.

Enorovan:
I haven't experienced it on release builds on my Windows 11 main, perhaps it depends on the performance AND the initialization.

dendi2008:
It's kinda weird bug, because I tested earlier build with my old fix (18.01.2026) and icon not showing even here, but before and when it was merged, and even some months later it was working. So, it could be even caused by different Windows 11 versions, or something else, not by impacto.

KKhanhH:
Yeah, Windows 11 taskbar is weird, it has had issues with icons just going blank sometimes too. I think this is probably an issue with icon caching or something on Windows. We're going to add icons and shortcut to the installer so it won't be that much of an issue IMO.

dendi2008:
> Yeah, Windows 11 taskbar is weird, it has had issues with icons just going blank sometimes too. I think this is probably an issue with icon caching or something on Windows. We're going to add icons and shortcut to the installer so it won't be that much of an issue IMO.

Yeah, that shoud work.

PR #499, opened 2026-07-22
[CCLCC] [CHLCC] Small Save/Quick Load/Load Menu changes.

dendi2008:
This PR will:

- Make Save/Quick Load/Load Menu after changing page select last selected entry before changing page (similar to System Menu #320).
- Make Save/Quick Load/Load Menu highlight the newest entry every time the menu is opened.
- Make selecting sound (sysse id 1) play when holding keyboard/gamepad button in CCLCC Save/Quick Load/Load Menu.

### DIFF
--- a/src/data/savesystem.cpp
+++ b/src/data/savesystem.cpp
@@ -5,8 +5,10 @@
 #include "../profile/vm.h"
 #include "../mem.h"
 
+#include <algorithm>
 #include <cstdint>
 #include <ctime>
+#include <ranges>
 
 namespace Impacto {
 namespace SaveSystem {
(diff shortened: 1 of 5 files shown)


PringlesGang:
I could be wrong but it seems like the original games actually always by default highlight the _newest_ entry every time the menu is opened. This could also be recomputed each time on `Show()`, so there would be no need for a separate member variable

dendi2008:
> I could be wrong but it seems like the original games actually always by default highlight the _newest_ entry every time the menu is opened. This could also be recomputed each time on `Show()`, so there would be no need for a separate member variable

Done. (I guess, because I might misunderstood you.)

PringlesGang (approved):
I actually misunderstood you; I thought you were talking about when you move in and out of the menu, but you were talking about between menu pages, my bad. Happy to see that implemented now as well though :p

It all looks good, thanks for the PR!

KKhanhH (approved):

KKhanhH:
Thanks for the PR!

PR #500, opened 2026-07-23
InstMesCls

PringlesGang:
Fully implements InstMesCls

`HideAllPages` appears in a handful of CCLCC spots in, but it doesn't seem to appear in any exotic cases

In CHLCC `HideAllPages` and `HidePage 1` appear only in `CHLCC_AY03` during the subway scene


src/vm/inst_dialogue.cpp
```diff
@@ -123,23 +123,172 @@ VmInstruction(InstMesSetID) {
...
+
+    } break;
+
+    case SetThreadPage: {
+      PopExpression(pageId);
+      SetFlag(SF_MESWINDOW0OPENFL + pageId, false);
+      SetFlag(SF_SHOWWAITICON + pageId, false);
+      DialoguePages[pageId].RenderName = false;
+      thread->DialoguePageId = pageId;
+    } break;
```
KKhanhH: this name kinda seems misleading

src/vm/inst_dialogue.cpp
```diff
@@ -263,7 +412,6 @@ VmInstruction(InstMesMain) {
 
     if (currentPage.AdvanceMethod == Skip && type != 1) {
       currentPage.PushBacklogEntry();
-      currentPage.Typewriter.Reset(AnimationDirection::Out);
```
KKhanhH, line 266: wasn't this the equivalent of MEStvramSetDispCt?

src/vm/inst_dialogue.cpp
PringlesGang: I mean it _does_ set `thread->DialoguePageId`, but yeah it also resets that page... Got any suggestions? `SetThreadPageAndReset`?

src/vm/inst_dialogue.cpp
PringlesGang: Yeah I made a mistake there. `MEStvramSetDispCt` sets the display counts (the point at which each glyph should start fading) all to zero, making it so all glyphs on the line should start fading in immediately. In our implementation the equivalent to this call is  actually `Typewriter.Finish(AnimationDirection::In)`, of which the resulting state is `Typewriter.IsIn() && Typewriter.Direction == AnimationDirection::In` which is equivalent to `pageState == Shown` which already has to be the case for these lines to be hit, so those calls wouldn't really do anything

KKhanhH (approved):

PR #501, opened 2026-07-25
C;HLCC C;CLCC original fonts support

### LINKED ISSUE #407: CHLCC text fade should fade the outline slightly before the main text

### LINKED ISSUE #466: Have the actual original fonts (with correct outlines) in Japanese
For C;HLCC, this is more or less done for the font part but the outline rendering isn't dealt with to look similar with the actual engine.
For C;CLCC, it uses SDF font and this might be another can of worm.

PringlesGang:
- [x] Refactor Font class
- [x] Add support for double-sheet fonts
- [x] Add support for edge-detected fonts
- [x] Add C;HLCC font
- [x] Add C;CLCC font
- [x] Add generic opacity curve profile configuration to font foreground and outline


Enorovan (approved):

profiles/cclcc/font-lb.lua
```diff
@@ -1,14 +1,17 @@
 root.Fonts = {
     ["Default"] = {
```
KKhanhH: Instead of putting everything under default, we should just make it so you can include both font files, under different keys, with another variable to select which key, and then under a patch you can just set which font is active.

src/font.h
```diff
@@ -2,98 +2,275 @@
...
+       OpacityCurve foregroundOpacityCurve, OpacityCurve outlineOpacityCurve)
+      : Type(type),
+        BitmapEmWidth(bitmapEmWidth),
+        BitmapEmHeight(bitmapEmHeight),
+        ForegroundOpacityCurve(foregroundOpacityCurve),
+        OutlineOpacityCurve(outlineOpacityCurve) {}
 };
 
-class BasicFont : public Font {
+class SingleSheetFont : public Font {
```
KKhanhH, line 73: Finalize derived classes

profiles/cclcc/game.lua
```diff
@@ -54,9 +54,11 @@ include('cclcc/charset.lua');
 include('cclcc/gamespecific.lua');
 include('common/animation.lua');
 include('common/charset.lua');
---include('cclcc/font.lua');
---include('cclcc/font-lb.lua');
-include('cclcc/font-lb-italic.lua');
+if (root.Language == "English") then
```
KKhanhH: include both

src/profile/fonts.cpp
```diff
@@ -11,95 +11,117 @@ void LoadFonts() {
...
+        const glm::ivec2 foregroundGridSize =
+            EnsureGetMember<glm::ivec2>("ForegroundGridSize");
+        const glm::ivec2 outlineGridSize =
+            TryGetMember<glm::ivec2>("OutlineGridSize")
+                .value_or(foregroundGridSize);
+
+        const std::optional<float> bitmapEmWidth =
+            TryGetMember<float>("BitmapEmWidth");
+        const std::optional<float> bitmapEmHeight =
+            TryGetMember<float>("BitmapEmHeight");
```
KKhanhH: you can factor these out to the top right?

KKhanhH (changes requested):

src/font.h
PringlesGang: It's not final though. `EdgeDetectedSingleSheet` derives from it

profiles/cc/dialogue.lua
```diff
@@ -51,7 +51,7 @@ root.Dialogue = {
     SkipIconCurrentType = SkipIconType.SpriteAnim,
     SkipIconSpriteAnim = "SkipIconSpriteAnim",
     SkipIconOffset = { X = 1712, Y = 702 },
-    DialogueFont = "Default",
+    DialogueFont = root.Language == "Japanese" and "Default" or "LanguageBarrier",
```
KKhanhH: leave it as "Default", "LanguageBarrier" can be added to impacto-patch repo

KKhanhH (approved):

Gunoshozo (approved):

PR #502, opened 2026-07-26
C;HLCC Seira Alarm Hide Text

### LINKED ISSUE #488: Seira Alarm Clock No MesWin
At the start of AA02 you have lines with the Seira alarm clock "talking".

Current behavior: we display it like a normal line but with the wait icon displayed below MesWin
Expected behavior: neither the line nor the meswin should display at all (for english we'll want to display but we'll do so via script patching)

Condition: SW_MESWIN0TYPE == 40

Enorovan:


src/text/dialoguepage.cpp
```diff
@@ -178,6 +178,8 @@ void DialoguePage::Render(const float alpha,
                           const RendererOutlineMode outlineMode) {
   // dialogue text
   if (GetFlag(SF_UIHIDDEN)) return;
+  if (ScrWork[SW_MESWIN0TYPE + Profile::Vm::ScrWorkMesStructSize * Id] == 40)
```
PringlesGang: Checking the binary (`MESdraw`) it looks to be a _little_ more nuanced than this. `SW_MESWIN0TYPE & (1 << 3)` means don't draw the textbox (including nametag but excluding any text), and `SW_MESWIN0TYPE & (1 << 5)` means don't draw the text (including name but excluding nametag)

PringlesGang (approved):
The wait icon fix is blowing up a lot larger than I anticipated (need to refactor the WaitIconDisplay and in doing so also the (Fixed)SpriteAnimations classes), so I think that would best fit in a separate PR. This one's okay to merge imo

PR #503, opened 2026-07-28
Wait icon refactor

PringlesGang:
- [x] Refactors `(Fixed)SpriteAnimation`
- [x] Refactors `WaitIconDisplay` to use classes instead
- [x] Gives each `DialoguePage` a `WaitIcon` instance
- [x] Adds a global `KeyWaitIcon` instance
- [x] Touches up wait icon display conditions
- [x] Touches up auto/skip/wait icon profile parameters
- [x] Make use of `WaitIconDispType` field in `MesModeFormat`
- [x] Add `SW_MESWINTYPE` enum

A very large part of this PR is mindless API changes dw

The `WaitIconDrawableWithoutTextbox` profile flag is required because the LCCs hardcode the wait icon display _inside_ the text window rendering function, whereas MOFDH does so _outside_


src/profile/animations.cpp
```diff
@@ -6,34 +6,40 @@ namespace Impacto {
 namespace Profile {
 
 void LoadAnimations() {
-  EnsurePushMemberOfType("Animations", LUA_TTABLE);
+  {
+    EnsurePushMemberOfType("Animations", LUA_TTABLE);
```
KKhanhH, line 10: you could declare lua getters for SpriteAnimationDef and then just do a EnsureGet for Animations, and do the same for FixedSpriteAnimationDef

src/hud/loadingdisplay.cpp
```diff
@@ -15,25 +15,26 @@ using namespace Impacto::Profile::ScriptVars;
...
 static bool IsResourceLoad = false;
 
-static SpriteAnimation ResourceLoadBg;
-static SpriteAnimation SaveLoadBg;
-static SpriteAnimation LoadingIcon;
-static SpriteAnimation LoadingText;
+static std::optional<SpriteAnimation> ResourceLoadBg;
+static std::optional<SpriteAnimation> SaveLoadBg;
+static std::optional<SpriteAnimation> LoadingIcon;
+static std::optional<SpriteAnimation> LoadingText;
```
KKhanhH, line 21: Why do these need to be optional if it's always being emplaced

src/hud/waiticondisplay.h
```diff
@@ -1,25 +1,147 @@
...
+
+  virtual void RenderImpl(glm::vec2 pos, glm::vec4 tint) = 0;
+
+  bool IsVisible() const;
+
+  bool IsKeyIcon() const { return ParentPage == nullptr; }
+  const DialoguePage* ParentPage = nullptr;
+};
+
+class VoidWaitIcon : public WaitIcon {
```
KKhanhH: we should get in the habit of finalizing derived classes

src/profile/dialogue.cpp
```diff
@@ -56,6 +56,47 @@ static void ConfigureNametag() {
...
+
+    case WaitIconType::Rotating:
+    case WaitIconType::Rotating3D: {
+      WaitIconSprite = EnsureGetMember<Sprite>("WaitIconSprite");
+      WaitIconAnimationDuration =
+          EnsureGetMember<float>("WaitIconAnimationDuration");
+    } break;
+
+    case WaitIconType::None: {
+      assert(false);
```
KKhanhH, line 93: should this be a false assertion? Maybe one of these games might not have a wait icon

src/hud/chlcc/dialoguebox.cpp
```diff
@@ -34,10 +34,13 @@ void DialogueBox::Render(const DialoguePageMode mode, const NameInfo& nameInfo,
...
 
     case DPM_REV: {
-      const glm::vec2 pos = (ScrWork[SW_MESWIN0TYPE] != 1
-                                 ? ParentPage.get().GetTextModeInfo().WindowPos
-                                 : GetScrWorkPos()) +
-                            ErinBoxPos - posOfs;
+      const glm::vec2 pos =
+          (ScrWork[SW_MESWIN0TYPE +
+                   ParentPage.get().Id * Profile::Vm::ScrWorkMesStructSize] &
+                   static_cast<int>(MesWinTypeBit::UseScrWorkPos)
```
KKhanhH: +MesWinTypeBit::UseScrWorkPos

KKhanhH: and all the other static_cast ints too

KKhanhH (changes requested):

src/hud/loadingdisplay.cpp
PringlesGang: Delayed construction. SpriteAnimations hold a reference wrapper to their definition, so the default constructor is deleted

src/profile/dialogue.cpp
PringlesGang: It already early returns on `WaitIconCurrentType == None` at the top of the function, so this should be unreachable. That's what the assert is for

src/profile/animations.cpp
PringlesGang: You mean just forego the whole `(FixedSprite)Animations` hashmaps and directly fetch the animation definitions in the profile initialization? Because right now `TryGet<SpriteAnimationDef>()` is implemented as getting from that map

Enorovan (approved):

src/profile/animations.cpp
KKhanhH: No, I mean just use the APIs we have that allow for you to do a tryget<Map> if the definition for tryget to the values for the map are defined

PringlesGang: Yeah but the problem is that `TryGet<SpriteAnimationDef>()` is currently implemented as fetching from the `Animations` map as per `HasEntityType<T>`, so I couldn't override it a second time to fetch the original map

Now, I don't see a particular need to go through this map anyway, because we're now pretty much just storing them twice (both in the map and in the inline profile variables)...

KKhanhH: ah yeah, then just skip it if it's not needed

KKhanhH (approved):

Gunoshozo (approved):

PR #512, opened 2026-08-11
LineAdd support

### LINKED ISSUE #497: Add separate bitfield support for keeping track of read added lines
Added lines currently just add to the line IDs where they are inserted. Because the read-lines bitfield is densely packed, line-read bits start overlapping with ones from subsequent scripts. We need to properly support tracking the read status of these added lines without overlap.

Could be done here (and then probs LB as well) or MgsScriptTools (possibly both).

A possible solution on the MgsScriptTools side is to give added lines a very large line ID that escapes the regular assigned bitfield into further unused space in the save file.

A possible solution on the Impacto side would be checking against script sizes in `Is/SetLineRead()` and accessing a different bitfield if it would go out of bounds for the script.

PringlesGang:
- [x] Adds proper save file support for added lines (so that their read state doesn't overlap with lines from later scripts) (#497)
- [x] Adds a system for marking ranges of lines as equivalent. These effectively share their "is read" bit

Use https://github.com/CommitteeOfZero/impacto-patch/pull/4 for testing LineAdd


src/profile/data/savesystem.cpp
```diff
@@ -136,9 +213,97 @@ void Configure() {
...
+
+  const size_t offset = lineId - lineRange->StartLineId;
+  const std::span<const LineRange> equivalentRanges =
+      EquivalentLineRanges[lineRange->EquivalentLineRangesIdx];
+  std::vector<std::pair<size_t, size_t>> equivalentLines(
+      equivalentRanges.size());
+  for (size_t idx = 0; idx < equivalentLines.size(); idx++) {
+    const LineRange& range = equivalentRanges[idx];
+    equivalentLines[idx] = {range.ScriptId, range.StartLineId + offset};
+  }
```
KKhanhH: You can probably switch this to use std transform instead

profiles/chlcc/equivalentlines.lua
```diff
@@ -0,0 +1,2019 @@
...
+    YU08A = 82,
+    YU08B = 83,
+    YU09A = 84,
+    YU09B = 85,
+};
+
+root.SaveData.EquivalentLines = {
+    {
+        {
+            ScriptId = scriptIds.AA08,
```
KKhanhH: it might be an easier read if you just had the 3 key values on one line

KKhanhH (approved):
Code is fine but didn't test the game

Enorovan (approved):

PR #515, opened 2026-08-12
BgEff fix

### LINKED ISSUE #269: Fix sprite overlay shader
The sprite overlay shaders, that mimic the effect of actual lighting by rendering textures unique to each background over the character sprites with a specific shader, still look slightly wrong compared to how these look on the PS4

PringlesGang:
C;CLCC PS4 seemingly only renders the top-right quadrant of the mask over the entire screen. I hope it at least makes sense to MAGES.

Fixes #269.
Also fixes the AdvanceWidths initialization for the fonts.


Gunoshozo (approved):
👍❤️

Enorovan (approved):

KKhanhH (approved):

PR #516, opened 2026-08-13
switch vcpkg

therigatoni:


KKhanhH:
Mostly fixes it so that vcpkg is used for switch dependencies, with port overlays to use the system packages whenever available

KKhanhH (approved):

PR #517, opened 2026-08-13
C;C LCC system menu button transition

PringlesGang:
The system menu buttons move in from/to the left as they're fading in/out when opening the system menu, and move to/from the right as they're fading out/in when you open a submenu


Enorovan (changes requested):
We're missing the button guide in the menu, I don't know if that's always been the case or not.

Enorovan (approved):

KKhanhH (approved):

PR #519, opened 2026-08-14
TOML User Config

### LINKED ISSUE #510: Change user configuration to be stored inside a toml file
Makes for easier writes programmatically.

KKhanhH:
Switch to TOML for user configuration
Move some variables to user config
Add game specific user config
Auto create user config if one doesn't exists, and add write support
Add hidden flag to game definitions


Enorovan (approved):

src/renderer/window.cpp
```diff
@@ -202,4 +204,88 @@ void ApplyCursorForFrame() {
...
+      haveBounds = SDL_GetDisplayBounds(0, &result) == 0;
+    } else {
+      haveBounds = SDL_GetDisplayUsableBounds(0, &result) == 0;
+    }
+    if (!haveBounds) {
+      ImpLog(LogLevel::Error, LogChannel::Render,
+             "Failed to get display bounds.\n");
+    }
+    return result;
+  }();
```
PringlesGang, line 222: This returns an uninitialized `SDL_Rect` on error, but just kinda continues initializing with it regardless. Maybe terminate or pass a default?

src/renderer/window.cpp
```diff
@@ -202,4 +204,88 @@ void ApplyCursorForFrame() {
...
+  };
+
+  auto const& config = UserConfig::CommonSettings;
+  WindowWidth = config.ResolutionWidth;
+  WindowHeight = config.ResolutionHeight;
+
+  if (!UserConfig::GetActiveGame().empty()) {
+    auto const& gameConfig = UserConfig::ActiveGameSettings();
+
+    if (gameConfig.ResolutionWidth && gameConfig.ResolutionHeight) {
```
PringlesGang, line 252: If it only uses the game config resolution if both width and height are provided, then maybe we should log a warning if only one of the two is

PringlesGang (approved):

PR #520, opened 2026-08-14
Config Overlay

### LINKED ISSUE #505: Add Impacto Overlays to select game and interact with engine settings/view achievements.
Overlay 1 (Launch): 

Image

Overlay 2 (Typical Recomp Overlay):

- Tab 1: Settings
  - Resolution
  - Fullscreen
  - About section with Framerate and Versions
  - Audio
  - Graphics (Renderer Choice)
- Tab 2: Enhancements (patch.lua stuff)
  - HasScriptedExitLogic
  - CloseBacklogWhenReachedEnd
  - DateFormat
  - HasDelusionMouseSupport
- Tab 3: Achievements
  - List achievement state based on the incoming Impacto Achievement System

KKhanhH:
Startup game selection screen
Overlay for user settings

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


src/overlay.cpp
```diff
@@ -0,0 +1,495 @@
...
+                           (SDL_GetModState() & KMOD_SHIFT))) ||
+        (Input::ControllerButtonWentDown[SDL_CONTROLLER_BUTTON_GUIDE])) {
+      OverlayShown = true;
+    } else if (OverlayShown &&
+               ImGui::IsKeyChordPressed(ImGuiMod_Shift | ImGuiKey_Tab)) {
+      OverlayShown = false;
+    }
+  } else {
+    OverlayShown = true;
+  }
```
PringlesGang: Maybe we shouldn't use the same keyboard shortcut as the Steam overlay? People (read: me) are gonna add this to their Steam library and play this through that as well. Maybe `Shift+0` or something?

For controllers there's not a whole lot we can do, but we could consider something like `Guide+Start` or smth

Maybe this should also be something configurable through user config

PringlesGang (changes requested):
- Restoring settings to default for General Settings while in-game removes resolution options for some reason
- Borderless acts just like exclusive fullscreen
- Changing the selected game should load that game's current settings imo
- Delusion mouse support is inverted
- I don't think the overlay should serialize the ActiveGame to the user profile, or should at least make it possible to reset it. Hell, now that we have a launcher maybe we should just drop the ActiveGame field altogether

src/overlay.cpp
```diff
@@ -0,0 +1,495 @@
...
+      bool tlOn =
+          magic_enum::enum_flags_test(enhancements.SubtitleConfig, Translation);
+      if (ImGui::Checkbox("Translation", &tlOn)) {
+        if (tlOn) enhancements.SubtitleConfig |= Translation;
+        if (!tlOn) enhancements.SubtitleConfig &= ~Translation;
+      }
+
+      bool karaOn =
+          magic_enum::enum_flags_test(enhancements.SubtitleConfig, Karaoke);
+      if (ImGui::Checkbox("Karaoke", &tlOn)) {
```
Enorovan:
```suggestion
      if (ImGui::Checkbox("Karaoke", &karaOn)) {
```

Enorovan (changes requested):

Enorovan:
Move About to last 😭

Enorovan (approved):

Enorovan (approved):
Re-approving to signify I've reviewed Dext's part and am done with mine, with the remaining question of the keyboard action to open it because for now everything proposed is less than ideal depending on US qwerty, EU qwerty, and azerty.

src/userconfig.cpp
```diff
@@ -312,7 +312,7 @@ void WriteUserConfig() {
 
   ImpLog(LogLevel::Debug, LogChannel::Config, "Writing user config to file\n");
 
-  userConfigFile->Seek(0, RW_SEEK_SET);
```
KKhanhH, line 287: what was the issue here?

Enorovan: Fix userconfig corrupting/crashing the game when the byte count diminishes because of config edits

src/overlay.cpp
```diff
@@ -0,0 +1,664 @@
...
+}
+
+void Show() {
+  if (Profile::Game::HasInit) {
+    if ((!OverlayShown && (Input::KeyboardButtonWentDown[SDL_SCANCODE_0] &&
+                           (SDL_GetModState() & KMOD_SHIFT))) ||
+        (Input::ControllerButtonWentDown[SDL_CONTROLLER_BUTTON_LEFTSTICK])) {
+      OverlayShown = true;
+    } else if (OverlayShown &&
+               ImGui::IsKeyChordPressed(ImGuiMod_Shift | ImGuiKey_0)) {
```
PringlesGang: We should also allow for closing the menu with controller

src/overlay.h
```diff
@@ -0,0 +1,10 @@
+
+
+namespace Impacto::Overlay {
+
+inline bool OverlayShown = false;
+inline bool OpenToAchievementsTab = false;
```
PringlesGang: Maybe we should just change this to an enum and allow for setting any tab to be opened to? Could also be nice for remembering which tab it was on when it was last closed

PringlesGang (commented):
When in the overlay, keyboard input is blocked but controller input is not. It should block both

PringlesGang (changes requested):
Uhh controller input is still not blocked for me and "closing" the menu with controller just blinks it out for a frame after which it's back again

PringlesGang (approved):
Borderless still isn't fixed, but as I'm the only one who can even reproduce that issue I'll take a look at it myself sometime after SDL3 or smth

PR #522, opened 2026-08-14
Update and render bgeff only if enabled

Gunoshozo:


PringlesGang (changes requested):
You're mixing up the BgEff that deals with sprite/background shader colouring and the BgEff that deals with framebuffer effects (cross-fading, picture-in-picture, etc.)

Games like S;G0 make use of the latter (which this code handles) but do not make use of the former (which `BgCha/FrameEffectType` configures). C;C LCC PS4 uses both, I'd assume C;C LCC Switch would use the latter but not the former.

PR #528, opened 2026-08-17
Switch font

PringlesGang:
- [x] Adds support for the LCC Switch fonts
- [x] Add position member to `ProcessedTextGlyph`
- [x] Refactor text functions
- [x] Refactor `Label` class

A large part of this PR is refactoring text and label functions. The Switch font is the first font type implemented in Impacto that does __not__ have fixed cell dimensions for each glyph in the sprite sheet. A _lot_ of the internal code depended on the position of the glyph to be the top left of its `DestRect`, which is thus no longer the case, so all these instances had to be fixed.

When testing, make sure all of the labels in all of the menus still behave properly. I had to touch just about all of them.

This PR was not built on top of the Switch assets PR, so to test you'd need to manually edit both `fonts.lua` files to have the default fonts resemble this:
```lua
["Default"] = {
    Type = FontType.EdgeDetectedSingleVariableWidthSheet,
    Sheet = "Font",
    BinaryPath = <FontBinaryAssetPath>,
    BitmapEmWidth = 53,
    BitmapEmHeight = 53,
    DifferenceFactor = 0.8,
    IntensityShift = 0.48,
    AlphaShift = 0.36,
},
```
with the `Font` entry in `sprites.lua` altered accordingly.


src/text/fonts/edgedetectedsinglevariablewidthsheetfont.cpp
```diff
@@ -0,0 +1,160 @@
...
+            Sprite(sheet, static_cast<float>(posX), static_cast<float>(posY),
+                   static_cast<float>(width), static_cast<float>(height)),
+        .Offset = {static_cast<float>(xOffset), static_cast<float>(yOffset)},
+        .AdvanceSize = {static_cast<float>(advanceWidth),
+                        static_cast<float>(advanceHeight)},
+    });
+  }
+}
+
+static void FillGlyphIndices(std::vector<uint16_t>& indices,
```
KKhanhH: why not just return the vector

src/text/text.cpp
```diff
@@ -232,237 +195,178 @@ std::pair<int, float> TextLayoutPlainLineHelper(
...
-  auto [count, currentX] = TextLayoutPlainLineHelper(
-      thd, stringLength, outGlyphs.begin(), font, fontSize, colors, opacity,
-      pos, alignment, blockWidth);
-  assert(outGlyphs.size() >= static_cast<size_t>(count));
-  TextLayoutAlignment(alignment, blockWidth, currentX, pos, count, outGlyphs);
-  return count;
+  return characterCount;
 }
 
+template std::vector<ProcessedTextGlyph> TextLayoutPlainLine(
```
KKhanhH, line 215: probably don't need to explicitly instantiate it, the concept handles it i think

src/text/text.cpp
```diff
@@ -232,237 +195,178 @@ std::pair<int, float> TextLayoutPlainLineHelper(
...
-  TextLayoutAlignment(alignment, blockWidth, currentX, pos, count, outGlyphs);
-  if (blockWidth > 0.0f) {
-    float containerRight = pos.x + blockWidth;
-    FitGlyphsForPlainLine(outGlyphs, containerRight);
+    Sc3Type auto&& stream, std::optional<size_t> maxLength, const Font& font,
+    const float fontSize, const DialogueColorPair colors, const float opacity,
+    const glm::vec2 pos, const TextAlignment alignment,
+    const float blockWidth) {
+  if (!maxLength.has_value()) {
+    auto lengthStream = decltype(stream)(stream);
```
KKhanhH: don't think you need decltype(stream), auto is by value so i think it should make a copy regardless

src/text/text.cpp
```diff
@@ -232,237 +195,178 @@ std::pair<int, float> TextLayoutPlainLineHelper(
...
+    maxLength.emplace(TextGetStringLength(lengthStream));
   }
-  return outGlyphs;
-}
 
-void FitGlyphsForPlainLine(std::span<ProcessedTextGlyph> glyphs,
-                           float containerRight) {
-  if (glyphs.empty()) return;
+  std::vector<ProcessedTextGlyph> outGlyphs(maxLength.value());
```
KKhanhH: just use *maxLength since you're guaranteed emplacing it

src/ui/widgets/label.cpp
```diff
@@ -97,109 +43,64 @@ void Label::SetSprite(Sprite const& label) {
...
+                                   RectF::Coalesce(lhs.DestRect, rhs.DestRect);
+                               return lhs;
+                             })
+                     .DestRect;
 }
 
-void Label::SetText(Vm::Sc3Stream& stream, float fontSize,
-                    RendererOutlineMode outlineMode,
+template <typename T>
+void Label::SetText(T str, float fontSize, RendererOutlineMode outlineMode,
```
KKhanhH: use the concept or extend it if needed

src/text/text.cpp
PringlesGang: Concepts are just abstractions over templates. If the definition is in a different translation unit from the use, the compiler still doesn't know which variants it should actually compile

KKhanhH: Oh didnt realize it was also in another TU

src/ui/widgets/chlcc/saveentrybutton.cpp
```diff
@@ -87,9 +87,17 @@ void SaveEntryButton::AddSceneTitleText(Vm::BufferOffsetContext strAdr,
                                         RendererOutlineMode outlineMode,
                                         glm::vec2 relativeTitlePosition,
                                         int colorIndex) {
-  SceneTitle =
-      Label(strAdr, glm::vec2(Bounds.X, Bounds.Y) + relativeTitlePosition,
-            fontSize, outlineMode, colorIndex, MaxTitleWidth);
+  Vm::Sc3VmThread dummy{};
+  dummy.ScriptBufferId = strAdr.ScriptBufferId;
+  dummy.IpOffset = strAdr.IpOffset;
+  std::vector<ProcessedTextGlyph> text =
```
Gunoshozo, line 93: maybe `dummy.useMsbStrings` should be used here. Add this to new additions, the rest is handled in the Switch branch

Gunoshozo (approved):

src/ui/widgets/label.h
```diff
@@ -14,6 +14,11 @@ namespace Widgets {
 
 using namespace Impacto::Fonts;
 
+template <typename T>
+concept LabelStringType = is_any_of_v<std::remove_reference_t<T>,
```
KKhanhH: decay_t to handle const

KKhanhH (approved):

src/ui/widgets/chlcc/saveentrybutton.cpp
PringlesGang: Uhh `UseMsbBuffers` isn't a member of `Sc3VmThread` yet; only on your branch
Also, should it be so unilaterally applied here? Iirc CCLCC PS4 doesn't make use of MSB strings yet?

Gunoshozo: Oh, right, my bad.
I think I use it on save entry buttons for cclcc, and I'm setting it's value for everything that might use text from msb

PR #530, opened 2026-08-22
MacOS Icons Impacto and LCCs

Gunoshozo:


resources/common/icons/icon.icns
KKhanhH, line 1: Move this into an icons subfolder since we'll have to also put .ico and .png

KKhanhH (approved):

PR #534, opened 2026-08-24
Bgfx refactor

### LINKED ISSUE #222: [After BGFX Refacto] Add a bicubic filter shader when upscaling
For C;HLCC in particular since it's 720p

PringlesGang:
Deprecate the existing OpenGL, DirectX9, and Vulkan renderers, and replace them with a new BGFX backend on a fully refactored renderer API.

Implements #222


src/userconfig.cpp
```diff
@@ -198,7 +198,7 @@ void Configure() {
       ImpLog(LogLevel::Fatal, LogChannel::Config, "\t{}\n",
              toml::format_error(err));
     }
-    throw std::runtime_error("Failed to parse user configuration file.");
+    Panic(LogChannel::Config, "Failed to parse user configuration file.");
```
KKhanhH: Leave this one, I'm planning on making it non fatal

vcpkg.json
```diff
@@ -49,9 +49,16 @@
...
     "boost-circular-buffer",
+    "boost-pfr",
     "toml11",
     "freetype",
-    "harfbuzz"
+    "harfbuzz",
+    {
+      "name": "bgfx",
+      "features": [
+        "tools"
```
KKhanhH, line 62: make sure you mark tools as a host dependency otherwise it won't compile on android

vcpkg.json
```diff
@@ -49,9 +49,16 @@
...
+      ]
+    }
```
KKhanhH, line 64:
```suggestion
    "bgfx",
    {
      "name": "bgfx",
      "host": true,
      "default-features": false,
      "features": [
        "tools"
      ]
    }
```
this should do it

